PR #29, opened 2019-07-12
Fix a typo in RangeError exception

ahmadov:
The condition for toExponential checks the argument to be between 0 and
100, but if the argument does not fit to that condition then the runtime
will raise a RangeError expection. Thus, the error message in RangeError
expection should also be between 0 and 100.


avp:
Good catch, I missed this in d61e5b05129c436087736aa85fc57a00d93ebaa0.

PR #31, opened 2019-07-13
Test Additions for JS Array Functions

SeunAdelekan:
The Hermes VM currently has support for Javascript's `Array.prototype.values()` and `Array.prototype.keys()` functions but tests for them were yet to be included.

I worked on tests for both functions in this pull request.


SeunAdelekan:
Thanks @avp! I have added tests to the `iterator.js` file instead.  Looking forward to your review.

PR #49, opened 2019-07-17
REPL: Load and write history to file

renatoalencar:
Use libreadline to load and write command history to file. By default the history is written to `~/.hermes_history' and is created when does not exists.

I was not able to resolve the full path, so I had to check the home directory and append to `.hermes_history`.

### DIFF
--- a/tools/repl/repl.cpp
+++ b/tools/repl/repl.cpp
@@ -230,23 +230,13 @@ static bool needsAnotherLine(llvm::StringRef input) {
 
 #if HAVE_LIBREADLINE
 // Load history file or create it
-std::error_code loadOrCreateHistoryFile(llvm::SmallString<128>& historyFile) {
+std::error_code loadHistoryFile(llvm::SmallString<128>& historyFile) {
   if (!llvm::sys::path::home_directory(historyFile)) {
     // Use ENOENT here since it could not found a home directory
     return std::error_code(ENOENT, std::system_category());
   }
 
   llvm::sys::path::append(historyFile, historyFileBaseName);
-
-  if (!llvm::sys::fs::exists(historyFile)) {
-    int fd;
-    auto err = llvm::sys::fs::openFileForWrite(llvm::Twine(historyFile), fd);
-    if (err) {
-      return err;
-    }
-
-    llvm::sys::fs::closeFile(fd);
-  }
 
   auto err = ::read_history(historyFile.c_str());
   if (err != 0) {
@@ -326,8 +316,8 @@ int main(int argc, char **argv) {
 
 #if HAVE_LIBREADLINE
   llvm::SmallString<128> historyFile{};
-  auto historyErr = loadOrCreateHistoryFile(historyFile);
-  if (historyErr) {
+  auto historyErr = loadHistoryFile(historyFile);
+  if (historyErr && historyErr.value() != ENOENT) {
     llvm::errs() << "Could not load history file: " << historyErr.message() << '\n';
   }
 #endif
@@ -345,8 +335,10 @@ int main(int argc, char **argv) {
       // EOF or user exit on non-continuation line.
       llvm::outs() << '\n';
 #if HAVE_LIBREADLINE
-      ::stifle_history(historyMaxEntries);
-      ::write_history(historyFile.c_str());
+      if (history_length > 0) {
+	::stifle_history(historyMaxEntries);
+	::write_history(historyFile.c_str());
+      }
 #endif
       return 0;
     }
--- a/tools/repl/repl.cpp
+++ b/tools/repl/repl.cpp
@@ -325,7 +325,7 @@ int main(int argc, char **argv) {
   runtime->getHeap().runtimeWillExecute();
 
 #if HAVE_LIBREADLINE
-  llvm::SmallString<128> historyFile = llvm::SmallString<128>();
+  llvm::SmallString<128> historyFile{};
   auto historyErr = loadOrCreateHistoryFile(historyFile);
   if (historyErr) {
     llvm::errs() << "Could not load history file: " << historyErr.message() << '\n';
--- a/tools/repl/repl.cpp
+++ b/tools/repl/repl.cpp
@@ -46,6 +46,7 @@
 #define C_STRING(x) #x
 
 #define HISTORY_FILE ".hermes_history"
+#define HISTORY_MAX_ENTRIES 500
 
 using namespace hermes;
 
@@ -346,6 +347,7 @@ int main(int argc, char **argv) {
       // EOF or user exit on non-continuation line.
       llvm::outs() << '\n';
 #if HAVE_LIBREADLINE
+      ::stifle_history(HISTORY_MAX_ENTRIES);
       ::write_history(historyFile.c_str());
 #endif
       return 0;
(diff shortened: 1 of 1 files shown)


tools/repl/repl.cpp
```diff
@@ -222,6 +227,38 @@ static bool needsAnotherLine(llvm::StringRef input) {
   return !stack.empty();
 }
 
+#if HAVE_LIBREADLINE
+// Load history file or create it
+std::error_code loadOrCreateHistoryFile(llvm::SmallString<128>& historyFile) {
+  llvm::Twine baseHistoryFile = llvm::Twine(HISTORY_FILE);
+
+  if (llvm::sys::path::home_directory(historyFile)) {
```
mhorowitz: nit: can you invert this check, so logic fails fast if there's no homedir?  Thanks!

tools/repl/repl.cpp
```diff
@@ -300,6 +345,9 @@ int main(int argc, char **argv) {
         (readResult == ReadResult::INTERRUPT && code.empty())) {
       // EOF or user exit on non-continuation line.
       llvm::outs() << '\n';
+#if HAVE_LIBREADLINE
+      ::write_history(historyFile.c_str());
```
mhorowitz: Does readline cap the size of the file?  I don't think we want to let it grow without bound.

tools/repl/repl.cpp
renatoalencar: The purpose of line 235 is not to only to check if there is a home directory, but to get the location of the home directory, so I can append to the file name to the home directory location. Maybe I didn't understand your idea, but I didn't see how to do that.

tools/repl/repl.cpp
renatoalencar: Sure, thanks for noticing this. There is a `stifle_history` function which can be used to limit this, I will build it.

renatoalencar: How many entries could be the limit? 500?

tools/repl/repl.cpp
mhorowitz: WhatI had in mind was to restructure the function body like this:
```C++
...
if (!llvm::sys::path::home_directory(historyFile)) {
  // Use ENOENT here since it could not found a home directory
  return std::error_code(ENOENT, std::system_category());
}

llvm::sys::path::append(historyFile, baseHistoryFile);
...
return std::error_code();

File: tools/repl/repl.cpp (hunk 2 above)
Comment (mhorowitz) [reply tools/repl/repl.cpp]: That's the default for bash.  At 100 chars per expression that's 50k file size, which seems unlikely to fill anybody's homedir, so seems reasonable.

File: tools/repl/repl.cpp (hunk 1 above)
Comment (renatoalencar) [reply tools/repl/repl.cpp]: Oh, that's great! I was actually feeling uncomfortable with this, and your idea fulfill exactly what I wanted. Thank you so much!

File: tools/repl/repl.cpp (hunk 2 above)
Comment (renatoalencar) [reply tools/repl/repl.cpp]: Thanks

File: tools/repl/repl.cpp (hunk 1 above)
Comment (renatoalencar) [reply tools/repl/repl.cpp]: Changed

File: tools/repl/repl.cpp [hunk 3]
```diff
@@ -42,6 +45,9 @@
 
 #define C_STRING(x) #x
 
+#define HISTORY_FILE ".hermes_history"
+#define HISTORY_MAX_ENTRIES 500
```
avp: Can we use `static const` variables here instead of preprocessor macros? They have the advantage of having proper typing.

tools/repl/repl.cpp
```diff
@@ -222,6 +228,38 @@ static bool needsAnotherLine(llvm::StringRef input) {
   return !stack.empty();
 }
 
+#if HAVE_LIBREADLINE
+// Load history file or create it
+std::error_code loadOrCreateHistoryFile(llvm::SmallString<128>& historyFile) {
+  llvm::Twine baseHistoryFile = llvm::Twine(HISTORY_FILE);
```
avp: `Twine` [should never be stored as a local variable](http://llvm.org/docs/ProgrammersManual.html#the-twine-class), because it may keep a reference to external memory.

`Twine` is cheap to create and should only be constructed as the argument to a function call. It has implicit constructors from things like `const char *` and string literals so you can pass `HISTORY_FILE` directly to `home_directory` and all the other `llvm::sys` functions below.

tools/repl/repl.cpp
renatoalencar: Done

tools/repl/repl.cpp
renatoalencar: Done

tools/repl/repl.cpp
```diff
@@ -288,6 +324,14 @@ int main(int argc, char **argv) {
 
   runtime->getHeap().runtimeWillExecute();
 
+#if HAVE_LIBREADLINE
+  llvm::SmallString<128> historyFile = llvm::SmallString<128>();
```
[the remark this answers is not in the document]
renatoalencar: Done

tools/repl/repl.cpp
```diff
@@ -222,6 +228,36 @@ static bool needsAnotherLine(llvm::StringRef input) {
...
+  if (!llvm::sys::path::home_directory(historyFile)) {
+    // Use ENOENT here since it could not found a home directory
+    return std::error_code(ENOENT, std::system_category());
+  }
+
+  llvm::sys::path::append(historyFile, historyFileBaseName);
+
+  if (!llvm::sys::fs::exists(historyFile)) {
+    int fd;
+    auto err = llvm::sys::fs::openFileForWrite(llvm::Twine(historyFile), fd);
```
mhorowitz: I'm running into some test errors around this code.  This creates an empty file, which causes a confusing error from `::read_history()`:
```
$ rm ~/.hermes_history
$ ./bin/hermes-repl
Could not load history file: No such file or directory
>>
```
I did some experimentation, and this seems to be due to the file being empty, and not including the magic version header.  If I start with `cp /dev/null ~/.hermes_history` I get the same warning.

And if I C-d immediately, so that a file is written with no contents, then when I run the next time, I get a different confusing error:
```
$ cat ~/.hermes_history
_HiStOrY_V2_
$ ./bin/hermes-repl
Could not load history file: Operation not permitted
>>
```
What's the purpose of creating the empty file here?  Can we suppress the empty write, so we don't get a warning?

Finally, I noticed the history file is mode 600, even though my umask is 0022.  Shouldn't it be mode 644?  Although I see other history files with the same mode, so maybe this is a readline "feature", it's not universal: node's history file is mode 644.

Sorry for all this.  The libreadline history support seems unfortunately brittle.

renatoalencar: Sometimes it could happens when you don't have a `HOME` environment variable set. Besides that, I really don't know what could happens.

The empty file is created in order to `write_history` to work, it opens the history file only on "append" mode, so if there is no file it returns a `ENOENT` error.

The LLVM file system API should actually create a 666 mode, but 600 is actually a good choice because of security issues on accessing history files.

I'm reading the libreadline source code in order to understand it.

renatoalencar: So, I'm avoiding creating empty files and ignoring `ENOENT` errors, because it happens in this case and that's an expected behavior.

About the permission, actually there is nothing we could do about it (at least using libreadline), both version [7](http://git.savannah.gnu.org/cgit/readline.git/tree/histfile.c?id=6c32f81cd66bbe86218469063690c84205661a5e#n649) and [8](http://git.savannah.gnu.org/cgit/readline.git/tree/histfile.c?id=8e6ccd0373d77b86ed37a9a7d232ccfea3d6670c#n654) set the file permission hardcoded.

PR #71, opened 2019-07-29
Enable building Hermes to x86:windows

mganandraj:
Hermes build fails on x86-windows as there is a type definition conflict between LLVM and Hermes on ssize_t type.. Both <llvm>\include\llvm-c\DataTypes.h and <Hermes>\external\zip\src\zip.h tries to define it with conflicting types on non-win64 windows platforms. This change makes those two definitions in sync.

### DIFF
--- a/external/zip/src/zip.h
+++ b/external/zip/src/zip.h
@@ -22,13 +22,19 @@ extern "C" {
 #if !defined(_SSIZE_T_DEFINED) && !defined(_SSIZE_T_DEFINED_) &&               \
     !defined(_SSIZE_T) && !defined(_SSIZE_T_) && !defined(__ssize_t_defined)
 #define _SSIZE_T
+
+#ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


dulinriley:
Instead of modifying zip.h (which is a third-party library header), could we instead add the macro `_SSIZE_T_DEFINED`? Then it shouldn't try to define its own.

mganandraj:
> Instead of modifying zip.h (which is a third-party library header), could we instead add the macro _SSIZE_T_DEFINED? Then it shouldn't try to define its own.

Agreed. The only tricky thing is to make sure that it we define "_SSIZE_T_DEFINED" only on platform/os combinations where it is guaranteed to be defined .. I'll take a look.. Thanks!

dulinriley:
@mganandraj Are you still working on this?

willholen:
Please rebase and reopen if you are still working on this.

PR #72, opened 2019-07-29
Vcpkg

mganandraj:
These are some changes that I had to make while building a VcPkg port of Hermes. VcPkg will make Hermes consumtion a breeze on Windows. I've just floated a PR in the VcPkg repo with a portfile for Hermes. 
https://github.com/microsoft/vcpkg/pull/7463

The changes in this PR in Hermes can be classified as follows,
1. Adding install targets to cmake scripts
2. Exporting Hermes runtime factory method for building dlls
3. Fixing the ICU lookup in cmake script for WindowsStore platform
4. Disable running regression tests when cross compiling to ARM, as we don't build some of the dependent tools when cross compiling.
5. Add an option to build_llvm script to not build, but only clone code and cmake configure.


CMakeLists.txt
```diff
@@ -493,6 +493,9 @@ set(LLVM_LIT_ARGS "-sv")
 # LLVM_RUNTIME_OUTPUT_INTDIR to our own build dir.
 set(LLVM_LIT_OUTPUT_DIR ${LLVM_BUILD_DIR}/${CMAKE_CFG_INTDIR}/bin)
 
+# Currently, when cross compiling to ARM on Windows host, we don't build some of the tools required for running the tests.
```
willholen: Does the target actually need to be conditionally defined? Is there harm in having a test target that doesn't yet pass on a particular platform?

mganandraj:
> Looks fine. Is this all that's required on our end? Is the rest maintained in the vcpkg repo? How would we know if we accidentally break that build?

Thanks for taking a look..  I'm currently working on end-end integration into the react-native-windows repo, and there will be some small subtle changes required.. I'll send another update in a couple of days..
I've a PR going to get the Hermes port checked into the vcpkg repo. Once that also goes in, you should be able to clone vcpkg repo and run "vcpkg install x64-windows" to build and install hermes for local development..

tmikov:
I have a couple of stylistic comments on how the PR is organized. It seems to contain a large number of commits, some of them repeatedly modifying the same files - is it possible to collapse them into a smaller number? 

I also noticed that the commits don't have descriptive titles and lack summaries. Ideally, to make it easy to review and accept each commit, it should look like an e-mail. A descriptive subject, then the body of the e-mail explaining what the commit does in some meaningful level of detail. Customarily we also apply a tag at the beginning of the subject, which in this case could be "[VCPkg]".

mganandraj:
> I have a couple of stylistic comments on how the PR is organized. It seems to contain a large number of commits, some of them repeatedly modifying the same files - is it possible to collapse them into a smaller number?
> I also noticed that the commits don't have descriptive titles and lack summaries. Ideally, to make it easy to review and accept each commit, it should look like an e-mail. A descriptive subject, then the body of the e-mail explaining what the commit does in some meaningful level of detail. Customarily we also apply a tag at the beginning of the subject, which in this case could be "[VCPkg]".



> I have a couple of stylistic comments on how the PR is organized. It seems to contain a large number of commits, some of them repeatedly modifying the same files - is it possible to collapse them into a smaller number?
> I also noticed that the commits don't have descriptive titles and lack summaries. Ideally, to make it easy to review and accept each commit, it should look like an e-mail. A descriptive subject, then the body of the e-mail explaining what the commit does in some meaningful level of detail. Customarily we also apply a tag at the beginning of the subject, which in this case could be "[VCPkg]".

Agreed. I noticed the commit history only after publishing the PR, and had the same thought. There are a couple more updates in the last week too. I'll revoke this PR and publish another one based on your suggestions. Thanks a lot.

willholen:
You could just update this PR as well.

PR #74, opened 2019-07-31
Remove libc++_shared.so from AAR and add hermes-cppruntime.aar

Kudo:
Currently the Hermes AAR includes libc++_shared.so, so that React Native needs to add [pickFirst](https://github.com/facebook/react-native/blob/0738fe5738c21463da93e3551f8f78ddcb6545e3/template/android/app/build.gradle#L181-L184) to workaround gradle conflict issue.
However, pickFirst is not a reliable way to select expected libc++_shared.so because we cannot ensure which gradle dependency is the first one for gradle to pick.
If user have other 3rd party gradle dependencies, the situation is even worst and this kind of bug is hard to troubleshoot.
Even though NDK maintains good ABI quality between different versions, I would still suggest if we could remove the pickFirst from RN project template.

This PR removes libc++_shared.so from Hermes AAR and adds another hermes-cppruntime.aar depedency if user need it for other purposes.
The approach is pretty much like as what we did in [jsc-android](https://github.com/react-native-community/jsc-android-buildscripts#for-react-native-version-058-below).


willholen:
Looks good to me. Am I understanding it correctly that this does not prevent ABI compatibility bugs, but instead makes such bugs easier to reproduce and fix because the `libc++_shared.so` is now chosen deterministically?

Will we need to e.g. update the react-native app template?

Kudo:
Yes, to truly prevent ABI incompatibility, we should align the NDK version for Hermes and RN.
Right now RN uses NDK [r19c](https://github.com/react-native-community/docker-android/blob/master/Dockerfile#L10) and Hermes seems to use latest NDK, say [r20](https://github.com/facebook/hermes/blob/master/.circleci/config.yml#L27).
After removing the pickFirst, we could determine the libc++_shared.so is from RN.
If user add 3rd party dependencies and add pickFirst manually,  that is user's risk.
Defensing ourselves that the pickFirst is not we introduced from RN template.

> Will we need to e.g. update the react-native app template?

Yes, if the PR landed and new version released. I will have a PR to remove the pickFirst from RN template.

PR #94, opened 2019-08-28
Remove build mode for TimeLimit

yuanyan:
`HERMESVM_TIMELIMIT` is not used, remove it in `API/hermes/hermes.cpp`.


yuanyan:
> @yuanyan, thanks for removing this. The change looks good. Can you add some basic description to the PR?

Added

PR #106, opened 2019-09-11
Include android .aars in npm package

azundo:
Update package filepaths for android aars so they are included in the
tarball generated by yarn pack.

### DIFF
--- a/npm/package.json
+++ b/npm/package.json
@@ -10,8 +10,8 @@
   },
   "files": [
     "README.md",
-    "android/hermes-debug.aar",
-    "android/hermes-release.aar",
+    "android/aar/hermes-debug.aar",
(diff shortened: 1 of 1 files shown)


willholen:
Awesome, you caught this in time (unlike me)!

It'll probably be better to undo the move than to update the path here though, that way it'll be compatible with previous releases. I'm currently running that through CircleCI in #108. 

Thanks!

azundo:
Thanks! Yes, that definitely seems like the better option - I wasn't sure if there was another reason for the path change and noticed the missing aars when trying to use the npm package from the ci artifacts.

azundo:
@willholen npm package looks good but I get the following crash when using 2.1.0 in dev and production with react-native 0.60.5: https://github.com/facebook/hermes/issues/109

PR #125, opened 2019-10-09
Remove 'hermesc' step from Android CircleCI build

willholen:
We don't need a compiler to build Hermes by default, so remove the failing build step.


willholen:
Welp, turns out we do.

PR #126, opened 2019-10-09
hermes

cmadunezim:
Add config for Docusaurus deploy command


rickhanlonii:
Thanks @cmadunezim but this change is already on master.

Issue #136, opened 2019-10-17
Hermes Date constructor problem


chufengma:
in hermes(0.2.1):

``` 
// test.js
let d = new Date("2019/10/16").getTime();
print(d);
----
// run with hermes 0.2.1
>>> hermes-cli-darwin-v0.2.1/hermes test.js
>>> NAN
```

in Node:
```
let d = new Date("2019/10/16").getTime();
console.log(d);
----
// run with node
>>> node test.js
>>> 1571155200000
```

chufengma:
And 
```
new Date("2019-10-16 00:00:00+0800").getTime()
---
NaN
```

mhorowitz:
@ljharb is correct.  In order to keep the VM small, we have chosen to parse only the strings the spec requires.  If you require parsing other date formats, we suggest using an npm module which implements richer date parsing such as https://www.npmjs.com/package/moment, or you can write your own.

flyskywhy:
https://www.npmjs.com/package/dayjs

    let d = new Date(dayjs("2019/10/16")).getTime();

PrimulaX:
In RN project on Android (Hermes enabled) app keeps crashing on `Date.prototype.toDateString()`

mhorowitz:
@PrimulaX please open a new issue using one of the forms at https://github.com/facebook/hermes/issues/new/choose.  Include as much logging or repro information as you can.

PR #144, opened 2019-11-01
fixed dateParse for supporting complex formats parsing

whatleo:
fixed dateParse problem, linked issue [#136](https://github.com/facebook/hermes/issues/136)
for example:
2019-11-02 01:02:03+08:00
2019-11-02 01:02:03+0800
2019-11-02 01:02:03+08
2019-11-02 01:02:03
2019-11-02T01:02:03
2019/11/02 01:02:03+08:00
2019/11/02/01:02:03+08:00
Fri Oct 02 2019 01:02:03 +08:00
Fri Oct 02 2019 01:02:03 +0800
Fri Oct 02 2019 01:02:03
Fri Oct 02 2019
Fri Oct 02 2019 01:02:03 GMT+08

### DIFF
--- a/lib/VM/JSLib/DateUtil.cpp
+++ b/lib/VM/JSLib/DateUtil.cpp
@@ -742,8 +742,13 @@ static double parseISODate(StringView u16str) {
   // Initialize these fields to their defaults.
   int32_t y, m{1}, d{1}, h{0}, min{0}, s{0}, ms{0}, tzh{0}, tzm{0};
 
+  // Assign separator character value, default '-'
+  char16_t sepChr = u'-';
+
   // See if need to adjust timezone
   bool adjustTZ = false;
+
+  // See if only year is set
   bool onlyYear = false;
 
   auto consume = [&](char16_t ch) {
@@ -767,15 +772,14 @@ static double parseISODate(StringView u16str) {
   y *= sign;
 
   if (it == end) {
+    // Only year is set
     onlyYear = true;
-  } else if (consume(' ')) {
-    onlyYear = false;
-  }
-
-  // Assign separator character value, default '-'
-  char16_t sepChr = u'-';
-  if (*it == u'/') {
+  } else if (*it == u'/') {
+    // Assign separator character to '/'
     sepChr = u'/';
+  } else if (*it == u' ' && it + 1 == end) {
+    // Only year and one blank character
+    consume(u' ');
   }
 
   if (consume(sepChr)) {
@@ -791,8 +795,9 @@ static double parseISODate(StringView u16str) {
     }
   }
 
-  // See if there's a time.
-  if (consume(u'T') || consume(u' ') || (sepChr == '/' && consume(sepChr))) {
+  // See if there's a time
+  // or see if separated day and time by '/' like '2016/01/01/00:00:00'.
+  if (consume(u'T') || consume(u' ') || (sepChr == u'/' && consume(sepChr))) {
     // Hours and minutes must exist.
     if (!scanInt(it, end, h)) {
       return nan;
@@ -841,22 +846,36 @@ static double parseISODate(StringView u16str) {
         return nan;
       }
       // Hour and minute of timezone adjustment.
-      if (it > end - 2 || !scanInt(it, it + 2, tzh)) {
-        return nan;
+      // See if timezone seperated by ':'.
+      if ((it + 1 <= end && *(it + 1) == u':') || (it + 2 <= end && *(it + 2) == u':')) {
+        if (!scanInt(it, end, tzh)) {
+          return nan;
+        }
+        tzh *= sign;
+        if (!consume(u':')) {
+          return nan;
+        }
+        if (!scanInt(it, end, tzm)) {
+          return nan;
+        }
+        tzm *= sign;
+      } else {
+        // Support if timezone didn't seperate by ':'.
+        if (it > end - 2 || !scanInt(it, it + 2, tzh)) {
+          return nan;
+        }
+        tzh *= sign;
+        if (it != end && (it > end - 2 || !scanInt(it, it + 2, tzm))) {
+          return nan;
+        }
+        tzm *= sign;
       }
-      tzh *= sign;
-      if (it < end && *it == ':') {
-        consume(u':');
-      }
-      if (it != end && (it > end - 2 || !scanInt(it, it + 2, tzm))) {
-        return nan;
-      }
-      tzm *= sign;
     } else if (it == end) {
       adjustTZ = true;
     }
-  } else if (!onlyYear) {
-    adjustTZ = true;
+  } else {
+    // If only year is set, don't need to adjust timezone.
+    adjustTZ = !onlyYear;
   }
 
   if (it != end) {
@@ -997,12 +1016,12 @@ static double parseESDate(StringView str) {
       return nan;
     if (!tok.equals(llvm::arrayRefFromStringRef("GMT")))
       return nan;
-    // Still paring, use default value tzh=0 & tzm=0
+    // Still parsing, use default value tzh=0 & tzm=0
     if (it == end) {
       goto complete;
     }
   } else if (it == end) {
-    // Still paring, use local timezone
+    // Still parsing, use local timezone
     adjustTZ = true;
     goto complete;
   }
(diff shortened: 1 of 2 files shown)


whatleo:
> In order for this to be approved, we'll need a lot of tests added.
> 
> Please update `test/hermes/date-constructor.js` with tests that demonstrate the new functionality.

Fail to pass the checks due to my mistake that didn't run unittests in my local.
I'll add more tests on test/hermes/date-constructor.js.
Thank you for your remind.

whatleo:
> Thanks for adding the tests.
> 
> Can you please add some more comments about how the new code works? I want to ensure it'll be easy to maintain going forward.
> 
> See the line comments for more information.

Hi there,
I fixed some dataParsing errors and add comments for my commit, and also add more unittests for date-constructor.js. 
But can't pass CircleCI tests, can't checkout my commit code to be causing the error.
Can you help me out?

lib/VM/JSLib/DateUtil.cpp
```diff
@@ -775,7 +792,7 @@ static double parseISODate(StringView u16str) {
   }
 
   // See if there's a time.
-  if (consume(u'T') || consume(u' ')) {
+  if (consume(u'T') || consume(u' ') || (sepChr == '/' && consume(sepChr))) {
```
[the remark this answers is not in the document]
whatleo: In chrome at present, can't parse date like '2016-01-01-00:00:00', but can parse date like '2016/01/01/00:00:00'，that's why I added this condition.

lib/VM/JSLib/DateUtil.cpp
```diff
@@ -823,17 +840,23 @@ static double parseISODate(StringView u16str) {
...
         return nan;
       }
       tzh *= sign;
-      if (!consume(u':')) {
-        return nan;
+      if (it < end && *it == ':') {
+        consume(u':');
       }
-      if (!scanInt(it, end, tzm)) {
+      if (it != end && (it > end - 2 || !scanInt(it, it + 2, tzm))) {
```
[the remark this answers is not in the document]
whatleo: Already comment there. Thank you for your remind.

lib/VM/JSLib/DateUtil.cpp
```diff
@@ -761,12 +765,25 @@ static double parseISODate(StringView u16str) {
     return nan;
   }
   y *= sign;
-  if (consume(u'-')) {
+
+  if (it == end) {
```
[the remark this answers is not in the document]
whatleo: That's indeed a problem. I fixed it.

lib/VM/JSLib/DateUtil.cpp
```diff
@@ -742,6 +742,10 @@ static double parseISODate(StringView u16str) {
   // Initialize these fields to their defaults.
   int32_t y, m{1}, d{1}, h{0}, min{0}, s{0}, ms{0}, tzh{0}, tzm{0};
 
+  // See if need to adjust timezone
+  bool adjustTZ = false;
+  bool onlyYear = false;
```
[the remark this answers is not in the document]
whatleo: Comments have been added.

lib/VM/JSLib/DateUtil.cpp
```diff
@@ -823,17 +840,23 @@ static double parseISODate(StringView u16str) {
...
       }
-      if (!scanInt(it, end, tzh)) {
+      // Hour and minute of timezone adjustment.
+      if (it > end - 2 || !scanInt(it, it + 2, tzh)) {
         return nan;
       }
       tzh *= sign;
-      if (!consume(u':')) {
-        return nan;
+      if (it < end && *it == ':') {
```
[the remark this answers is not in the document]
whatleo: It's fixed.

lib/VM/JSLib/DateUtil.cpp
```diff
@@ -959,10 +997,15 @@ static double parseESDate(StringView str) {
       return nan;
     if (!tok.equals(llvm::arrayRefFromStringRef("GMT")))
       return nan;
-  }
-
-  if (it == end)
+    // Still paring, use default value tzh=0 & tzm=0
```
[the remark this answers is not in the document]
whatleo: Has been corrected.

test/hermes/date-constructor.js
```diff
@@ -204,14 +212,60 @@ print(Date.parse('2016T12:30:47.1234-07:00'));
...
+// CHECK-NEXT: 1
+print((new Date('2017/02/15 15:01:37.243-0700')).getSeconds());
+// CHECK-NEXT: 37
+print((new Date('2017/02/15/15/01:37.243-07:00')).getTime());
+// CHECK-NEXT: NaN
+print((new Date('2017/02/15 15:01:37.243-07')).getTime());
+// CHECK-NEXT: 1487196097243
+print((new Date('2017/02/15 15:01:37.243-7')).getTime());
+// CHECK-NEXT: 1487196097243
+print((new Date('2017/02/15 15:01:37.243-7:')).getTime());
```
avp: I'm concerned about these trailing `:` formats. Is this really an expected working case? It seems malformed and that we should return `NaN` here.

lib/VM/JSLib/DateUtil.cpp
```diff
@@ -972,15 +1045,42 @@ static double parseESDate(StringView str) {
...
+  if (it == end) {
     return nan;
-  tzm *= sign;
+  }
+  // Hour and minute of timezone adjustment.
+  if (consume(u':') && it == end) {
+    // See if timezone is '-:' or '+:'
+    adjustTZ = true;
+  } else if ((it + 1 < end && *(it + 1) == u':') || (it + 2 < end && *(it + 2) == u':')) {
+    // See if timezone seperated by ':', is like '-07:00' or '-7:0' or '-7:00' or '-07:' or '-7:'
```
avp:
```suggestion
    // See if timezone separated by ':', is like '-07:00' or '-7:0' or '-7:00' or '-07:' or '-7:'
```

lib/VM/JSLib/DateUtil.cpp
```diff
@@ -972,15 +1045,42 @@ static double parseESDate(StringView str) {
...
-  tzh *= sign;
-  if (!scanInt(it, it + 2, tzm))
+  // Timezone MUST be set.
+  if (it == end) {
     return nan;
-  tzm *= sign;
+  }
+  // Hour and minute of timezone adjustment.
+  if (consume(u':') && it == end) {
+    // See if timezone is '-:' or '+:'
```
avp: Why is `-:` a valid timezone adjustment? Is the implication that the adjustment is `-00:00`, but because the numbers were omitted, we need to add the `localTZA()` anyway?

lib/VM/JSLib/DateUtil.cpp
```diff
@@ -972,15 +1045,42 @@ static double parseESDate(StringView str) {
...
+    }
+    tzh *= sign;
+    if (!consume(u':')) {
+      return nan;
+    }
+    if (scanInt(it, end, tzm)) {
+      tzm *= sign;
+    }
+  } else {
+    // See if timezone didn't seperate by ':'
```
avp:
```suggestion
    // If timezone adjustment isn't separated by ':'.
```

lib/VM/JSLib/DateUtil.cpp
```diff
@@ -972,15 +1045,42 @@ static double parseESDate(StringView str) {
...
+    if (it < end - 2) {
+      // See if timezone is like '0700' or '070'
```
avp: Do we really want to accept 3-digit timezone adjustments here? User intent here seems non-obvious - is it a number of minutes, are the first two digits the hour, are the last two digits the minutes?

avp:
As to the CircleCI tests failing, we're working on fixing a code mismatch between React Native and Hermes which is resulting in compilation failures.

test/hermes/date-constructor.js
whatleo: > I'm concerned we're being too permissive to the point that the semantics of `Date.parse` become too vague. Returning an unintended time can be more harmful and harder to debug than failing to parse and returning `NaN` for ambiguous input, particularly if we don't have extremely thorough documentation on non-obvious behavior. In the current state of the code on master, the permissible input is documented in the ECMAScript spec, with the only divergence being that certain minor delimiters are considered optional.
> 
> Browsers already support all kinds of different formats, but they are forced to hold a certain level of legacy compatibility which Hermes, as a newer engine, is not really obligated to.
> 
> I don't disagree with the spirit of this change, in that multiple of the new formats have readily apparent ways of understanding how they should be parsed, but I want to make sure that we don't confuse users and increase bugs because Hermes suddenly became too permissive in its Date parsing methods.
> 
> There are third-party libraries for parsing Dates in a more structured and well-documented format, so if a user wants to use those, the behavior is more easily understood.
> 
> Specific concerns are in the line comments, but in general, I think we should cut back a bit on the amount of possible inputs we're parsing in this PR.

I understand your concern.
Some of my changes were really too aggressive. I'll revert my changes and resubmit new PR that 
 isn't too permissive and support well-documented format. It is necessary to correct Date.parse on 
specified case. Because our business was done and invoked on JSC before. Switch to Hermes and we don't want to make big change to use third-party libraries. 
As for modifying some unittest's result,  I just want to keep logic of dateParsing consistant with JSC in the same situation.

lib/VM/JSLib/DateUtil.cpp
whatleo: I'll revert this changes.

lib/VM/JSLib/DateUtil.cpp
whatleo: Not good at English. LOL.

lib/VM/JSLib/DateUtil.cpp
whatleo: I'll revert this changes.

whatleo:
> As to the CircleCI tests failing, we're working on fixing a code mismatch between React Native and Hermes which is resulting in compilation failures.

Give this CircleCI tests work a thumbs up. It's amazing.

PR #173, opened 2020-01-11
Website playground

exced:
Introduces a playground for the Hermes website.

This is highly based on https://tmikov.github.io/hermes-explorer-page/ from @tmikov 

### Improvements
- Editor suggestions do not show up properly
- Selection highlighting between the input and output editors (like  [Compiler explorer ](https://godbolt.org) or [AST explorer](https://astexplorer.net)) to help understand and debug the engine.


tmikov:
I am excited about this!

exced:
Can I do something about the linter / test ?

Also did you have specific features in mind when you started this project ?

tmikov:
The linter is complaining that hermes.js doesn't have a copyright header, but that file is generated... I am checking how to add an exclusion rule. The test failure is not really relevant (I just need to rebase internally). So, no need to worry much about those.

rickhanlonii:
Hey @exced, this is awesome! Thanks so much for pushing this up.

I have a few suggestions:
- Can we use dark mode for the editor so it matches the site (`theme="vs-dark"`)?
- Can we soft wrap the lines so there's no horizontal scrolling (`wordWrap: 'on'`)?
- Can we remove the minimap (`minimap: { enabled: false }`)?
- Can we make the cli options and run button dark mode?
- Can we remove the footer and make the screen full size so there's no double scrolling?

Let me know what you think!

rickhanlonii:
Oh one other thing - could we make the editor split 1/3 JS and 2/3 Hermes output? I think the Hermes output it the main thing people will want to read so we should give it a little more room so it's easier to read.

exced:
Hey @rickhanlonii, thanks for reviewing this, I added dark mode consistency for the editor and cli input, made the editors more responsive for small screens and updated the editor options.

I was a little bit struggling to hook to dark / light theme updates and came up with a hacky hook.
I was looking for an exported hook or a theme context in Docusaurus v2 but could not find any. Am I missing something here ?

rickhanlonii:
This update is awesome!!

Screen Shot 2020-01-30 at 9 54 31 AM


Screen Shot 2020-01-30 at 9 53 27 AM

website/src/pages/playground/useTheme.js
```diff
@@ -0,0 +1,41 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { useState, useEffect } from 'react';
+import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
+
+function useTheme() {
```
rickhanlonii: Nice hack, I don't think there's a way to listen to theme updates (cc @yangshun in case he knows)

yangshun: This seems to be taken from https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-theme-classic/src/theme/hooks/useTheme.js.

To obtain the theme via context, you can use [ThemeProvider](https://github.com/facebook/docusaurus/blob/master/packages/docusaurus-theme-classic/src/theme/ThemeProvider/index.js). But I guess the intention here is to use a custom implementation.

website/src/pages/playground/index.js
```diff
@@ -0,0 +1,177 @@
...
+  ellapsed: undefined,
+};
+
+function Playground() {
+  const [{ loading, ellapsed, output }, dispatch] = useReducer(
+    reducer,
+    initialState
+  );
+  const [input, setInput] = useState('const a = 1; \nprint(a);');
+  const [args, setArgs] = useState('-help');
```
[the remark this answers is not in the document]
exced: Yes sure, I was also thinking about these " cli shortcuts". Do we want other buttons for options we use a lot, like `[?] [Bytecode] [AST]` ?

Yes totally, I can also remove the "log" thing in this case.

On a side note, what I "saw" ultimately was 
- left: args input + help + shortcuts + editor
- right: different "viewers" (JSON tree, bytecode, plain text) + buttons to switch  + interactions between the editor and the viewer, like 
  - fold JSON node <-> fold code
  - highlight statement <-> highlight lexer loc (AST / byte code) etc but it requires lot more work :sweat_smile:

It can also be incrementally added in the future

exced:
@yangshun Thanks for the feedbacks, I'll update the code accordingly and respect the Docusaurus conventions this time 😅

yangshun:
Btw this is so cool, I love that you guys even try to support dark mode for the playground!

website/src/pages/playground/index.js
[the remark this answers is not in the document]
tmikov: The original explorer was warming Hermes up immediately on page load, so usually by the time the user started doing something it was already warmed up. Are we doing the same here?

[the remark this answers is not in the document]
rickhanlonii: @exced that plan sounds good to me as a followup!

website/src/pages/playground/useTheme.js
rickhanlonii: @exced does using ThemeProvider work for you?

rickhanlonii:
This looks great! Let's ship and iterate on it!

website/src/pages/playground/useTheme.js
exced: I wasn't able to import it 😞

website/src/pages/playground/index.js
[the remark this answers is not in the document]
exced: @tmikov We're still warming it up upfront

website/src/pages/playground/useTheme.js
exced: @rickhanlonii Does it have to be in [exports](https://github.com/facebook/docusaurus/tree/master/packages/docusaurus/src/client/exports) in order to import it ? I didn't manage to import it

yangshun: @exced do this:

```js
import ThemeProvider from '@theme/ThemeProvider';
```

Not `@docusaurus/ThemeProvider`.

exced: Hey @yangshun sorry for being late. I tried to use this import and it did not work. Correct me if I'm wrong, `ThemeProvider` is coming from `docusaurus-theme-classic` (which is imported through `docusuraus-preset-classic` in my case) and I can't see it in `node_modules`. I can probably override some stuff with a `theme` folder but I don't think it's the right approach, have you heard of an issue about the server not loading the presets properly ? Thanks

yangshun: Oh that's because I added `ThemeProvider` in a recent commit and it hasn't been published onto npm yet. Give me a while we'll release a new version containing that and you can upgrade to it.

tmikov:
Update on the status of landing this PR: the only thing blocking us is that it contains the Hermes Asm.js build artifact itself (hermes.js):
a) It makes it too easy to publish a web site that is out of sync with the current version of Hermes 
b) More importantly it is currently impossible to generate that artifact the the repository itself :-) . (I generated that Asm.js build from a private hack which isn't checked in).

As far as b) is concerned, we are in the process of landing the pre-requisite build script changes, so soon we will be able to actually produce that artifact in a public way :-)

About a), we are not entirely sure what the correct solution is, so I would like to discuss the alternatives:
1. Delete hermes.js from the source tree and simply require whoever is updating the website to build it first. That is likely to be problematic because it requires installing the Emscripten toolchain. 
2. Build and check-in the current version and do it every time we make a public release. This at least makes it likely that the website will be in sync with the latest official release (except the first time). 
3. Publish a NPM with the Asm.js/Wasm build and have the website depend on it. This doesn't automatically solve anything, since someone still has to publish the NPM, but at least it doesn't burden the person updating the website with having to build Hermes.
4. Somehow have the website publishing script build Hermes/Asm.js. Perhaps it can download it from CircleCI? (I don't know whether that is possible).

Solution 2) looks like it would require the least effort and we can do it right away. Any thoughts?

website/src/pages/playground/useTheme.js
yangshun: `docusaurus@2.0.0-alpha.43` has been released which contains the `ThemeProvider`. Should work now!

exced: @yangshun thanks, it works like a charm

rickhanlonii:
@yangshun I tried deploying this with `yarn build && yarn deploy` but Docusarus fails with 

Screen Shot 2020-04-01 at 12 53 56 PM

I think the issue is that `monaco-editor-webpack-plugin` requires webpack 4.5 but Docusarus uses webpack 4.4X. Do you have plans to upgrade to 4.5?

yangshun:
Oic. We can totally upgrade it. Will do it within the next few days.

rickhanlonii:
Actually, I tried with webpack 4.5X and got the same failure, not sure what the root cause is :/

rickhanlonii:
Submitted an issue here: https://github.com/facebook/docusaurus/issues/2494

tmikov:
Folks, the playground was deployed but it doesn't work. What can we do to fix it?

tmikov:
https://github.com/facebook/hermes/issues/217

exced:
Hi everyone, sorry I didn't respond earlier, is there anything I can do to help or is it something internal ? Thanks

tmikov:
@exced can we figure out what's wrong and how to fix it?

exced:
I ran my branch locally and it seems to work. Do you think it might come from the Hermes build ?
I'll build it locally and try to see if I can find something 🤞

tmikov:
I am certain it is not from the Hermes build, the same build is working at https://tmikov.github.io/hermes-explorer-page/ .

Can you try the website from the current master branch?

(BTW, let's move the discussion to #217)

PR #176, opened 2020-01-13
[Hermes] Remove unnecessary trailing semicolons

ridiculousfish:
gcc9 is salty about useless semicolons, particularly those that come after
brackets which close namespaces. Remove them.


ridiculousfish:
Note I updated the PR with a few more

PR #177, opened 2020-01-14
Remove the separate build_llvm.py step, just build LLVM along with Hermes

ridiculousfish:
This is a PR to make it easier to integrate Hermes into existing CMake projects.

Prior to this change, Hermes required downloading and separately building  LLVM. Hermes then builds using the LLVM build directory. This custom build  step is difficult to support within a larger CMake project.

With this change, Hermes no longer requires a separate LLVM build step.  Hermes simply includes the LLVM CMake files, and the LLVM dependencies get  built the normal way.

This allows LLVM to be sourced from someplace other than the download script, e.g. a git submodule.

The CMake build flow with this change is:

    fetch_llvm.py
    configure.py
    cd build && ninja

There's some caveats:

1. `ninja clean` will now remove all the LLVM artifacts too.
2. I have not yet tested cross-compilation, Windows builds, or CI.
3. There's an annoying issue where Hermes implicitly depends on certain LLVM generated headers. Specifically, `hermes/IR/Instrs.h` pulls in `llvm/IR/CFG.h` which eventually tries to include `llvm/IR/Attributes.inc`, which is generated.

    This means that the `Instrs.h` header has always had an implicit dependency on the LLVM target `llvm-headers`. My preference for fixing this would be to add a Hermes `include` target, which depends on `llvm-headers`, and then every Hermes library depends on `include`.

    To keep the diff small, I simply added `llvm-headers` to `LLVM_COMMON_DEPENDS`. This causes the LLVM headers to be "rebuilt" too often, but it is very fast, so it is livable.


ridiculousfish:
I'll take a crack at the android CI failure.

tmikov:
Will this add LLVM as a direct dependency in an IDE? Initially we were building Hermes as part of LLVM; CLion would slow down because it had to reindex the entire LLVM every time.

More importantly, we are working on including the only the parts of LLVM that we need as part of Hermes (essentially a fork of LLVM, with most of it deleted and with a changed namespace). That should remove all complexity associated with it. I think it would supersede this change. It should happen very soon (couple of weeks).

ridiculousfish:
@tmikov interesting points. Agreed that the LLVM factoring you describe would be better than this. I'll close (for now!)

PR #204, opened 2020-04-02
Update playground to be deployable

rickhanlonii:
## Overview

We need to switch this over to using Suspense to support server side rendering and unblock the deploy.

## Test

Tested by patching, building, and deploying the website 😏


Huxpro:
@rickhanlonii Heeey Ricky I was wondering when does this SSR being used? I am finding here from git blames.

rickhanlonii:
@Huxpro docusaurus 2 uses SSR, check out the issue here: https://github.com/facebook/docusaurus/issues/2494

PR #220, opened 2020-04-17
DeclareGlobalVar to not overwrite existing properties

ridiculousfish:
Ha ha, this is a doozy.

Prior to this change, DeclareGlobalVariable would unconditionally store
undefined as a property on global. With this change, only do so if there is
not already such an own-property.

The reason for the fix is that eval() can otherwise overwrite existing
global variables. For example:

var q = true;
eval('var q;');

it is wrong for 'q' to result in undefined (and in practice, v8 and jsc do not).

Test Plan: Added a new test

### DIFF
--- a/lib/VM/Interpreter.cpp
+++ b/lib/VM/Interpreter.cpp
@@ -2221,6 +2221,8 @@ tailCall:
         DefinePropertyFlags dpf =
             DefinePropertyFlags::getDefaultNewPropertyFlags();
         dpf.configurable = 0;
+        // Do not overwrite existing globals with undefined.
+        dpf.setValue = 0;
 
         CAPTURE_IP_ASSIGN(
             auto res,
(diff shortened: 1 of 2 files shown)


lib/VM/Interpreter.cpp
```diff
@@ -2218,6 +2218,24 @@ CallResult<HermesValue> Interpreter::interpretFunction(
...
+                ownDesc));
+        if (ownPropFound) {
+          ip = NEXTINST(DeclareGlobalVar);
+          DISPATCH;
+        }
+
+        // We know that there is no such own-property; however the set may still
+        // fail (for example, if global is frozen).
         DefinePropertyFlags dpf =
             DefinePropertyFlags::getDefaultNewPropertyFlags();
```
tmikov, line 2240: What if instead we don't set `dpf.setValue = 1`? Then the old value shouldn't be overwritten.

ridiculousfish: Huh! I falsely believed that `setProperty=0` would *not* store a new property. A definite improvement!

PR #223, opened 2020-04-20
Add website README.md with deploy instructions

rickhanlonii:
## Overview

This PR adds a README to the `website/` directory that briefly explains the website and provides deploy instructions.

### DIFF
--- a/website/README.md
+++ b/website/README.md
@@ -1,0 +1,36 @@
+# Hermes JS Engine Website
+
+This directory contains the source code for the Hermes JS Engine website available at https://hermesengine.dev/.
+
+## Developing
+
+This project uses [Docusaurus v2](https://v2.docusaurus.io/).
+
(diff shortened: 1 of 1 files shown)


rickhanlonii:
Yeah @yangshun I agree but unfortunately the files are too big to land internally so we can't check them into master and they only live on the github pages branch.

rickhanlonii:
Also, interestingly, when I tried running `yarn deploy` previously it didn't update or push to the gh-pages branch so I needed to so all these steps anyway?

yangshun:
> Also, interestingly, when I tried running yarn deploy previously it didn't update or push to the gh-pages branch so I needed to so all these steps anyway?

You could have been on a broken version of Docusaurus 2. Upgrading to alpha.50 should fix it. Also, I noticed that going to https://hermesengine.dev/playground/ gives a broken page because the asset paths are incorrect (they need to have a leading slash). That's really weird to me because the configuration looks correct. I have to dig deeper.

PR #224, opened 2020-04-20
Improve sanitizer support in CMake builds

ridiculousfish:
This makes the `--enable-asan` option to `configure.py` work again, and adds an analogous `--enable-ubsan`. It also disables misaligned reads in miniz when run under UBSan, which was the one test failure. With this change, the hermes tests pass with both ASan and UBSan (gcc-flavored).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,6 +116,14 @@ set(HERMESVM_API_TRACE OFF CACHE BOOL
 set(HERMESVM_SANITIZE_HANDLES OFF CACHE BOOL
   "Enable Handle sanitization")
 
+# Enable Address Sanitizer
+set(HERMES_ENABLE_ADDRESS_SANITIZER OFF CACHE BOOL
+  "Enable -fsanitize=address")
+
+# Enable Undefined Behavior Sanitizer
+set(HERMES_ENABLE_UNDEFINED_BEHAVIOR_SANITIZER OFF CACHE BOOL
+  "Enable -fsanitize=undefined")
+
 # Build with -DHERMES_SLOW_DEBUG for debug builds
 # This does not affect release builds
 set(HERMES_SLOW_DEBUG ON CACHE BOOL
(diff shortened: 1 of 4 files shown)


external/zip/src/miniz.h
```diff
@@ -281,7 +281,7 @@
 #define MINIZ_LITTLE_ENDIAN 1
 #endif
 
-#if MINIZ_X86_OR_X64_CPU
+#if MINIZ_X86_OR_X64_CPU && !HERMES_UBSAN
```
ridiculousfish, line 284: This is admittedly clumsy, but miniz does not allow overriding these defines from outside the miniz file, so I think a more principled change would require writing a lot more code.

dulinriley: Alternatively, you could add a ubsan suppression annotation
```
__attribute__((no_sanitize("undefined")))
```
to any functions in this header that use unaligned accesses.
I think doing that would be fine to upstream to the miniz.h owner, and help other users of this library.

(replace `"undefined"` with a more specific annotation type like `"alignment"` if possible)

CMakeLists.txt
```diff
@@ -281,6 +289,14 @@ endif()
 if(HERMESVM_SANITIZE_HANDLES)
     add_definitions(-DHERMESVM_SANITIZE_HANDLES)
 endif()
+if (HERMES_ENABLE_ADDRESS_SANITIZER)
+    append("-fsanitize=address" CMAKE_CXX_FLAGS CMAKE_C_FLAGS CMAKE_EXE_LINKER_FLAGS)
+endif()
+if (HERMES_ENABLE_UNDEFINED_BEHAVIOR_SANITIZER)
+    add_definitions(-DHERMES_UBSAN)
+    # Do not enable the vptr sanitizer, as it requires RTTI.
+    append("-fsanitize=undefined -fno-sanitize=vptr" CMAKE_CXX_FLAGS CMAKE_C_FLAGS CMAKE_EXE_LINKER_FLAGS)
```
dulinriley, line 298: As this line is currently with `-fsanitize=undefined`, UBSAN will only print *warnings* to stderr and will not halt the program.

It is probably better to have a harder condition, by using in addition to that flag `-fno-sanitize-recover=all`, which forces the program to halt whenever a UBSAN rule is violated.

utils/build/common.py
```diff
@@ -121,6 +122,8 @@ def build_dir_suffix(args):
     suffices = []
     if args.enable_asan:
         suffices += ["asan"]
+    if args.enable_asan:
```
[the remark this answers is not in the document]
ridiculousfish: Yikes, whoops

CMakeLists.txt
ridiculousfish: Done

external/zip/src/miniz.h
ridiculousfish: I agree this would be a better fix, but we are all busy people :)

utils/build/common.py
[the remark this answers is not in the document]
ridiculousfish: Fixed

external/zip/src/miniz.h
tmikov: Yeah, I don't need to be fixing miniz.

avp: Please do add this change to `external/zip/README.md` in case we update it, though.

tmikov: What's the purpose of the external/zip/sanitize-blacklist.txt? @avp @dulinriley

tmikov: Shouldn't we use `-fsanitize-blacklist=sanitize-blacklist`?

dulinriley: Oh good point, I forgot I had already made that file.
Yes `-fsanitize-blacklist=sanitize-blacklist.txt` should suppress the problems from miniz.

kuba--: Thanks, for using `kuba--/zip`. Would be cool, if you can contribute back ;-)

PR #225, opened 2020-04-20
Introduce DecoratedObject and rebuild HostObject on top of it

ridiculousfish:
This PR conceptually factors out the "holds a native pointer" aspect of HostObject, allowing it to be used independently from the "interposes properties." The new intermediate class is called DecoratedObject.

This fills an integration gap.  Say you wish to attach a native thing to a JS object, and manipulate the native bits from prototype functions. This is not possible with HostObject, because the prototype is not searched. DecoratedObject is the missing intermediate layer: it manages the native bits (memory management, etc) without affecting the object model.

The third commit rebuilds HostObject on top of DecoratedObject; now HostObject only does the property-interposing parts.


include/hermes/VM/HermesValueTraits.h
```diff
@@ -67,6 +67,8 @@ HERMES_VM_GCOBJECT(JSArrayIterator);
 HERMES_VM_GCOBJECT(JSStringIterator);
 HERMES_VM_GCOBJECT(JSProxy);
 HERMES_VM_GCOBJECT(JSCallableProxy);
+HERMES_VM_GCOBJECT(DecoratedObject);
+HERMES_VM_GCOBJECT(HostObject);
```
ridiculousfish, line 71: Apparently nobody ever tried to form `Handle<HostObject>`!

ridiculousfish:
Hitting pause while I fix the failing builds

PR #226, opened 2020-04-21
Miscellaneous small QoL improvements

ridiculousfish:
This is a grab-bag of some miscellaneous small changes. It clarifies some parameters in CreatorFunction, passes in the Context pointer, and fixes a footgun in HermesValueTraits.

### DIFF
--- a/include/hermes/VM/Callable.h
+++ b/include/hermes/VM/Callable.h
@@ -705,8 +705,12 @@ class NativeFunction : public Callable {
 /// Object.
 class NativeConstructor final : public NativeFunction {
  public:
+  /// A CreatorFunction is responsible for creating the 'this' object that the
+  /// constructor function sees.
+  /// \p proto is the '.prototype' property of the constructor and should be set
+  /// as the __proto__ for the nascent object.
   using CreatorFunction =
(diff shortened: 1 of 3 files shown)


include/hermes/VM/Callable.h
```diff
@@ -705,8 +705,12 @@ class NativeFunction : public Callable {
 /// Object.
 class NativeConstructor final : public NativeFunction {
  public:
+  /// A CreatorFunction is responsible for creating the 'this' object that the
+  /// constructor function sees.
+  /// \p proto is the '.prototype' property of the constructor and should be set
+  /// as the __proto__ for the nascent object.
   using CreatorFunction =
-      CallResult<PseudoHandle<JSObject>>(Runtime *, Handle<JSObject>);
+      CallResult<PseudoHandle<JSObject>>(Runtime *, Handle<JSObject> proto);
```
tmikov: We should really fix this naming. In many places this is called `parentHandle` and the actual field is called `parent_`. Though in NativeConstuctor we also call it `prototype` and in other places `prototypeHandle`.

tmikov:
@ridiculousfish can you edit the PR description to contain all the details from the individual commits. The problem is that PRs are imported as a single commit containing just the PR description. Reading the individual commit descriptions is much more informative, but that information will be lost.

PR #233, opened 2020-04-30
Don't include VM header in public API.

alloy:
As per [this comment](https://github.com/microsoft/react-native/issues/319#issuecomment-621389883), removing this inclusion to keep VM header requirement out of distributions.

### DIFF
--- a/API/hermes/hermes.h
+++ b/API/hermes/hermes.h
@@ -14,8 +14,11 @@
 #include <string>
 
 #include <hermes/Public/RuntimeConfig.h>
-#include <hermes/VM/GCExecTrace.h>
 #include <jsi/jsi.h>
+
+#ifdef HERMESVM_API_TRACE
+#include <hermes/VM/GCExecTrace.h>
+#endif
 
 struct HermesTestHelper;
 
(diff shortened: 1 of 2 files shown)


API/hermes/hermes.h
```diff
@@ -121,10 +124,12 @@ class HermesRuntime : public jsi::Runtime {
   /// it can be written into the trace for later replay.
   const ::hermes::vm::MockedEnvironment &getMockedEnvironment() const;
 
+#ifdef HERMESVM_API_TRACE
   /// Get a structure representing the execution history (currently just of
   /// GC, but will be generalized as necessary), to aid in debugging
   /// non-deterministic execution.
```
alloy, line 130: I wasn’t entirely sure if this comment is referring to aiding debugging _of_ tracing, in which case this may need to be `HERMESVM_API_TRACE_DEBUG`? For now I went with `HERMESVM_API_TRACE` because e.g. `SynthTrace.h` guards the inclusion of the `hermes/VM/GCExecTrace.h` header using that macro.

alloy:
Just saw https://github.com/facebook/hermes/commit/cdc4ea7dafbf7ebf7a96f64e264993fae6d1932e, makes sense 👍

PR #242, opened 2020-05-07
Miscellaneous CMake improvements

ridiculousfish:
This PR is a collection of changes to the CMake build, aimed at making it easier to integrate Hermes into other projects.


ridiculousfish:
A `config.h` would indeed be better than this and I'd be happy to switch to that. I didn't pursue that approach, because I thought it would be hard to implement it in the BUCK build.

tmikov:
Is it necessarily incompatible with Buck, I wonder. We can keep the current approach with Buck and just use a static empty config.h. I am leaning toward a config.h. We are already talking about a binary Hermes library distribution, which would have this exact problem - how would the distribution encode the flags that were used to build Hermes?

On the other hand, users of Hermes are supposed to go through JSI, which shouldn't be affected by this. Before we make a decision, can you share anything more about which defines specifically caused a problem or in what use case?

ridiculousfish:
Yes, I would be happy to provide more context, I will write it up.

ridiculousfish:
Closing this; I will republish it with a config.h approach and more context

PR #254, opened 2020-05-27
Fix Error message when passing 'undefined' to the constructor

### LINKED ISSUE #252: A bug about constructor Error
### Testcase:
``` javaScript:
var NISLFuzzingFunc = function (nislMutationParameter0) {
    var i = new Error(nislMutationParameter0);
    var tmp = i["message"];
    print(tmp);
    print(tmp.length);
    print(typeof tmp);
};
var nislMutationArgument0 = undefined;
var NISLCallingResult = NISLFuzzingFunc(nislMutationArgument0);
```
### Command:
``` shell
hermes testcase.js
```

### Output:
```
undefined
9
string
```

### Expected output:
```

0
string
```

According to ES6, when the value of `message`  is `undefined` in [19.5.1.1](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-error-message), The "message" property of Error instance should not be "undefined". 
In addition,  output of other engines such as v8, javascriptCore and spiderMonkey is showing in Expected output.

ahmedkrmn:
Fixes #252.
To solve this issue, this PR checks for the message in `JSError::setMessage()` function. 
If it's equal to `undefined`, it's changed to an empty string.

### DIFF
--- a/test/hermes/error.js
+++ b/test/hermes/error.js
@@ -37,6 +37,14 @@ e = new Error('another-message');
 print(e);
 //CHECK: Error: another-message
 
+e = new Error(undefined);
+print(e);
+//CHECK: Error
+
+e = new Error('undefined');
+print(e);
+//CHECK: Error: undefined
+
 e = new TypeError();
 print(e);
 //CHECK: TypeError
--- a/lib/VM/JSLib/Error.cpp
+++ b/lib/VM/JSLib/Error.cpp
@@ -107,7 +107,8 @@ static CallResult<HermesValue> constructErrorObject(
   JSError::setupStack(selfHandle, runtime);
 
   // new Error(message).
-  if (args.getArgCount() >= 1) {
+  // Only proceed when 'typeof message' isn't undefined.
+  if (!args.getArg(0).isUndefined()) {
     if (LLVM_UNLIKELY(
             JSError::setMessage(
                 selfHandle, runtime, runtime->makeHandle(args.getArg(0))) ==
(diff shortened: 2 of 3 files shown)


ahmedkrmn:
@tmikov thanks for your insights, but what do you think is the cause of the hex addresses output that I've mentioned?

ahmedkrmn:
@tmikov Thanks for your insights, will fix the commit soon.

tmikov:
LGTM. One more thing and it will be ready to go. Can you please add a test case in https://github.com/facebook/hermes/blob/master/test/hermes/error.js .

ahmedkrmn:
@tmikov 
Hi Mr. Tzvetan,
Can you please recheck the PR?

test/hermes/error.js
```diff
@@ -37,6 +37,14 @@ e = new Error('another-message');
 print(e);
 //CHECK: Error: another-message
 
+e = new Error(undefined);
+print(e);
```
tmikov, line 41: I think we need something like:
```
print(e.hasOwnProperty(e, 'message'));
```

ahmedkrmn: There were already some test cases tackling this so I thought it would be unnecessary to repeat them.

tmikov: This is specifically what this PR is doing. It is removing the "message" property if the parameter is `undefined`. We need a test for that specifically.

ahmedkrmn: You're right, I don't know how I missed this!
Have you added the test? It appears that the bot closed PR.

ahmedkrmn: I believe `e.hasOwnProperty(e, 'message')` always returns `false` whether the message is `undefined` or not.

PR #271, opened 2020-06-12
Stop madvise MADV_DONTNEED identifier hashes

ridiculousfish:
While loading a bytecode file, Hermes will populate the identifier table
with the precomputed hashes. It then issues an madvise MADV_DONTNEED call
as it will not need to read the hashes again.

This is fine and dandy if the bytecode file is mmap'd from disk. But if it
is from memory, MADV_DONTNEED has destructive semantics: the hashes are
*zeroed*. If you then run the same bytecode file again, all identifiers
will have zero hashes, and they will be treated as distinct from the first
time the bytecode file was run.

I propose just removing the madvise call. If that is too aggressive, it
could instead be controlled by a RuntimeModuleFlag.

Test Plan: a nice new test


ridiculousfish:
(Tracking down this bug was quite the journey!)

ridiculousfish:
Incidentally the test will only fail on Linux, because only Linux has destructive semantics for MADV_DONTNEED.

tmikov:
This is a great catch!

tmikov:
I think this was added as part of our effort to decrease memory usage. Since the decision was data driven, I think we may not want to completely undo it (but let's see first what @kodafb thinks).

A runtime flag also doesn't seem entirely safe.

I wonder whether there is an easy and performant way to check at runtime whether the memory is a file backed mmap. 

Make no mistake, we must fix this!

kodafb:
Oops. Nice catch indeed!

I'm not aware of an easy/portable way to detect whether the buffer is file-backed (we'd need to parse /proc/self/maps...)

Also, the amount of memory involved here is pretty small, and if there is memory pressure (and the buffer is file-backed) then the OS would likely evict these pages anyway since they are only accessed once.

So I'm fine with removing it.

tmikov:
Update: @kodafb accepted the diff!

PR #282, opened 2020-07-01
[WIP] Build all targets in linux and macOS CircleCI tests

neildhar:
Ensure that all targets can be built when running tests


mhorowitz:
As we expected, this is detecting a latent bug CI was not detecting before, which is the cause of he test-macos failure.  We'll wait for a fix for that to land before landing this.  Thanks!

PR #284, opened 2020-07-01
An implementation of Intl.getCanonicalLocales for Android

mganandraj:
This is an implementation of Intl.getCanonicalLocales API (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/getCanonicalLocales) using the platform libraries available in Android and tests. 

Most of the implementation details are documented inline with the code itself. In short, we had decided not to package any external internationalization libraries or data with Hermes, and leverage what the platform provides favoring space efficiency over correctness and ECMAScript spec compliance.


lib/Platform/Intl/java/com/facebook/hermes/intl/Intl.java
```diff
@@ -8,45 +8,306 @@
...
-
-  // Implementer note: This method corresponds roughly to
-  // https://tc39.es/ecma402/#sup-string.prototype.tolocaleuppercase
-  public static String toLocaleUpperCase(List<String> locales, String str)
-  {
-    return "uppered";
-  }
+
+    private static boolean isAlphaNum(String name, int min, int max) {
+        char[] chars = name.toCharArray();
```
willholen, line 24: Could you please use `.charAt` in a simple for loop instead to avoid allocating and copying an array on each invocation? I tried it in a basic benchmark and it showed a 12x improvement.

mganandraj: Hi @willholen .. Sure i'll do .. It looked so wasteful to convert the string to array, and i don't remember how i ended up with this .. Will make the changes in next iteration ..

android/.gitignore
```diff
@@ -1,2 +1,90 @@
 .gradle
 .externalNativeBuild
+
+# https://github.com/github/gitignore/blob/master/Android.gitignore
```
mhorowitz, line 4: It feels a bit strange to put this in every project, instead of in a system-wide or user-specific .gitignore file.

android/intltest/java/com/facebook/hermes/test/HermesIntlAndroidTest.java
```diff
@@ -17,14 +17,38 @@
...
     try (JSRuntime rt = JSRuntime.makeHermesRuntime()) {
-      rt.evaluateJavaScript(script);
+      evalScriptFromAsset(rt, "intl.js");
+    }
+  }
+
+  @Test
+  public void testIntlGetCanonicalNamesTestScriptFromAsset() throws IOException {
+    try (JSRuntime rt = JSRuntime.makeHermesRuntime()) {
+      evalScriptFromAsset(rt, "sta.js");
```
mhorowitz, line 44: I am concerned that this approach to testing will not scale.

Hermes currently has infrastructure to run against test262, but we expect developers to download the tests separately.  See https://github.com/facebook/hermes/tree/master/utils/testsuite.

In this case, it's a little different, since we need the tests to be around at build time, but there are more than 800 files in `test262/test/intl402`, and copying them all into the Hermes repo does not feel like the right answer.

A better solution would be to have the build system take a path to the testsuite as an option.  We already maintain a list of tests which we expect Hermes will not pass; we can maintain a second one for Intl.  (We can't use the same one, as all Intl tests need to be skipped on platforms without Intl.)

android/intltest/java/com/facebook/hermes/test/assets/intl_getCanonicalNames_2.js
```diff
@@ -0,0 +1,131 @@
...
+  var canonicalLocales = Intl.getCanonicalLocales(tag);
+  assert.sameValue(canonicalLocales.length, 1);
+  assert.sameValue(canonicalLocales[0], regularGrandfatheredUTS35[tag]);
+});
+
+
+// https://github.com/tc39/test262/blob/master/test/intl402/Intl/getCanonicalLocales/preferred-variant.js
+var canonicalizedTags = {
+  // "ja-latn-hepburn-heploc": "ja-Latn-alalc97-hepburn", // TODO::Hermes icu4j reurns ja-latn-hepburn-heploc.. while icu4c returns ja-Latn-alalc97-hepburn .. Not V8 parity.
+};
```
mhorowitz, line 73: Combining and modifying the upstream test files creates a burden for upgrading test262. I'd rather use a skiplist.

lib/Platform/Intl/java/com/facebook/hermes/intl/Intl.java
```diff
@@ -8,45 +8,306 @@
...
+        // The failed test cases are commented with TODO::Hermes tag in intl_getCanonicalNames_*.js
+        //
+        // Ref: https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/util/ULocale.html#forLanguageTag-java.lang.String-
+        // Ref: https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1Locale.html#af76028775e37fd75a30209aaede551e2
+
+        ULocale uLocale= ULocale.forLanguageTag(locale);
+        String tag = uLocale.toLanguageTag();
+        return tag;
+
+        // Note: TODO : An alternative implementation could use https://developer.android.com/reference/android/icu/util/ULocale.Builder to build the locale.. which does structural validation internally based on documentation.
```
mhorowitz, line 198: This seems like it would reduce the java code size, by delegating the validation to the library.  Is there a reason to prefer doing it the way it is here?

android/intltest/java/com/facebook/hermes/test/HermesIntlAndroidTest.java
```diff
@@ -17,14 +17,38 @@
...
+  private void evalScriptFromAsset(JSRuntime rt, String filename) {
+    try {
+      AssetManager assets = getInstrumentation().getContext().getAssets();
+      InputStream is = assets.open(filename);
+      String script = new BufferedReader(new InputStreamReader(is))
+              .lines().collect(Collectors.joining("\n"));
+      rt.evaluateJavaScript(script);
+    } catch (IOException ex) {
+      ex.printStackTrace();
+      assert (false);
```
mhorowitz, line 30: an exception will already cause a test failure, and allow additional tests to run and report status.  What is the reason to explicitly catch and assert here?

android/intltest/java/com/facebook/hermes/test/HermesIntlAndroidTest.java
mganandraj: Hi @mhorowitz ,

Sorry for a late response .. It's been quite a hectic last couple of weeks both at home and work :)

I'm back now and am making some progress fixing many of the corner cases following Firefox's implementation .. which in turn will make it easier for us to have a low overhead blacklist based tests ... I'll update this PR soon with those changes along with some performance fixups ..

android/.gitignore
mganandraj: Makes sense .. I've pulled out the .gitignore file ..

android/intltest/java/com/facebook/hermes/test/HermesIntlAndroidTest.java
mganandraj: Hi @mhorowitz 
We do have a blacklist based test coverage in the latest iteration. We've added the test files from test262 as it is without any changes, and apply the blacklist. It would be better to download the tests on the fly, but i'd like to go ahead checking in those, for now.

In order to be able to formalize the test as this, I had to make the implementation much richer. Currently we have a formal statement/definition on what do we don't support, for instance.

android/intltest/java/com/facebook/hermes/test/assets/intl_getCanonicalNames_2.js
mganandraj: Sure.. In the latets iteration, we don't make any changes to the test262 files.

lib/Platform/Intl/java/com/facebook/hermes/intl/Intl.java
mganandraj: In the latest iteration, we've switcher to Builder, but it comes with its own deficiencies. For e.g. the Builder implementation in icu4j don't look up the grandfathered tag mapping. But, it has the performance advantage as it doesn't require parsing the whole tag again, which we have already parsed.

This is a super critical aspect about this change. Currently, we are adding a set of autogenerated data into the binary for improving the correctness, but resulting in bigger binary size. Having predictable behaviour is important. I completely understood your concern. I think, we should review the details about the implementation in the coming weeks. For now, i'd like to go ahead with the current implementation.

android/intltest/java/com/facebook/hermes/test/HermesInstrumentationTest.java
```diff
@@ -72,12 +72,12 @@ public void testDateFormat() {
       }
 
       rt.evaluateJavaScript(
-          new StringBuilder()
-              .append("var date = new Date(Date.UTC(2012, 11, 20, 3, 0, 0));\n")
-              .append("d1 = date.toLocaleDateString();\n")
-              .append("d2 = date.toLocaleTimeString();\n")
-              .append("d3 = date.toLocaleString();\n")
-              .toString());
+              new StringBuilder()
```
mganandraj, line 75: These formatting changes are unintentional .. I'll revert them ..

mganandraj:
Hi .. Thanks a lot for taking a look.. While implementing the actual Intl object (Collator, NumberFormat Etc.), i realized that the code in this PR needs to be structured slightly differently to allow easy reuse and avoid repeated computations. Hence, pulling this change back temporarily so that i can share a new PR with restructured code. 

Sorry for any inconvenience.

lib/Platform/Intl/java/com/facebook/hermes/intl/Intl.java
mhorowitz: That sounds fine.  It will be easier to consider size/functionality tradeoffs when the system is more mature.

mhorowitz:
No worries.  Looking forward to the new PR!

mganandraj:
Hi @mhorowitz  .. Please find a new PR with getCanonicalLoales and Collator implementation here : https://github.com/facebook/hermes/pull/317
Thanks!

PR #287, opened 2020-07-08
Run e2e tests against RN master.

avp:
Summary: We should be able to run against RN master.

Differential Revision: D22440288


willholen:
Follow #294 for the thrilling conclusion.

PR #290, opened 2020-07-10
Add job to build macOS CocoaPods binaries

alloy:
## Summary

Adds a CI job to build binaries for macOS to be used via CocoaPods.

## Test Plan

```
$ mkdir test-hermes-build
$ cd test-hermes-build/
$ curl -L -O https://6909-154201259-gh.circle-artifacts.com/0/tmp/hermes/output/hermes-runtime-darwin-v0.5.0.tar.gz
$ tar -zxvf hermes-runtime-darwin-v0.5.0.tar.gz
$ tree .
.
├── LICENSE
├── README.md
├── destroot
│   ├── Library
│   │   └── Frameworks
│   │       ├── hermes.framework
│   │       │   ├── Headers
│   │       │   │   ├── CompileJS.h
│   │       │   │   ├── DebuggerAPI.h
│   │       │   │   ├── Public
│   │       │   │   │   ├── Buffer.h
│   │       │   │   │   ├── CrashManager.h
│   │       │   │   │   ├── CtorConfig.h
│   │       │   │   │   ├── DebuggerTypes.h
│   │       │   │   │   ├── GCConfig.h
│   │       │   │   │   ├── GCTripwireContext.h
│   │       │   │   │   └── RuntimeConfig.h
│   │       │   │   ├── SynthTrace.h
│   │       │   │   ├── SynthTraceParser.h
│   │       │   │   ├── TraceInterpreter.h
│   │       │   │   ├── TracingRuntime.h
│   │       │   │   ├── hermes.h
│   │       │   │   └── hermes_tracing.h
│   │       │   ├── Resources
│   │       │   │   └── Info.plist
│   │       │   ├── Versions
│   │       │   │   ├── 0
│   │       │   │   │   ├── Headers
│   │       │   │   │   │   ├── CompileJS.h
│   │       │   │   │   │   ├── DebuggerAPI.h
│   │       │   │   │   │   ├── Public
│   │       │   │   │   │   │   ├── Buffer.h
│   │       │   │   │   │   │   ├── CrashManager.h
│   │       │   │   │   │   │   ├── CtorConfig.h
│   │       │   │   │   │   │   ├── DebuggerTypes.h
│   │       │   │   │   │   │   ├── GCConfig.h
│   │       │   │   │   │   │   ├── GCTripwireContext.h
│   │       │   │   │   │   │   └── RuntimeConfig.h
│   │       │   │   │   │   ├── SynthTrace.h
│   │       │   │   │   │   ├── SynthTraceParser.h
│   │       │   │   │   │   ├── TraceInterpreter.h
│   │       │   │   │   │   ├── TracingRuntime.h
│   │       │   │   │   │   ├── hermes.h
│   │       │   │   │   │   └── hermes_tracing.h
│   │       │   │   │   ├── Resources
│   │       │   │   │   │   └── Info.plist
│   │       │   │   │   └── hermes
│   │       │   │   └── Current
│   │       │   │       ├── Headers
│   │       │   │       │   ├── CompileJS.h
│   │       │   │       │   ├── DebuggerAPI.h
│   │       │   │       │   ├── Public
│   │       │   │       │   │   ├── Buffer.h
│   │       │   │       │   │   ├── CrashManager.h
│   │       │   │       │   │   ├── CtorConfig.h
│   │       │   │       │   │   ├── DebuggerTypes.h
│   │       │   │       │   │   ├── GCConfig.h
│   │       │   │       │   │   ├── GCTripwireContext.h
│   │       │   │       │   │   └── RuntimeConfig.h
│   │       │   │       │   ├── SynthTrace.h
│   │       │   │       │   ├── SynthTraceParser.h
│   │       │   │       │   ├── TraceInterpreter.h
│   │       │   │       │   ├── TracingRuntime.h
│   │       │   │       │   ├── hermes.h
│   │       │   │       │   └── hermes_tracing.h
│   │       │   │       ├── Resources
│   │       │   │       │   └── Info.plist
│   │       │   │       └── hermes
│   │       │   └── hermes
│   │       └── hermes.framework.dSYM
│   │           └── Contents
│   │               ├── Info.plist
│   │               └── Resources
│   │                   └── DWARF
│   │                       └── hermes
│   ├── bin
│   │   ├── hbcdump
│   │   ├── hdb
│   │   ├── hermes
│   │   ├── hermesc
│   │   └── hvm
│   └── include
│       ├── hermes
│       │   ├── CompileJS.h
│       │   ├── DebuggerAPI.h
│       │   ├── Public
│       │   │   ├── Buffer.h
│       │   │   ├── CrashManager.h
│       │   │   ├── CtorConfig.h
│       │   │   ├── DebuggerTypes.h
│       │   │   ├── GCConfig.h
│       │   │   ├── GCTripwireContext.h
│       │   │   └── RuntimeConfig.h
│       │   ├── SynthTrace.h
│       │   ├── SynthTraceParser.h
│       │   ├── TraceInterpreter.h
│       │   ├── TracingRuntime.h
│       │   ├── hermes.h
│       │   └── hermes_tracing.h
│       └── jsi
│           ├── JSIDynamic.h
│           ├── decorator.h
│           ├── instrumentation.h
│           ├── jsi-inl.h
│           ├── jsi.h
│           ├── jsilib.h
│           └── threadsafe.h
├── hermes-runtime-darwin-v0.5.0.tar.gz
└── hermes.podspec

25 directories, 84 files
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,7 @@ workflows:
       - android
       - linux
       - macos
+      - macos-cocoapods
       - windows
       - npm:
           requires:
@@ -29,7 +30,7 @@ workflows:
             - windows
       - test-linux
       - test-macos
-      - test-macos-podspec
+      - test-macos-cocoapods
       - test-e2e:
           requires:
             - npm
@@ -225,6 +226,52 @@ jobs:
           paths:
             - .
 
+  macos-cocoapods:
+    macos:
+      xcode: "10.3.0"
+    environment:
+      - HERMES_WS_DIR: /tmp/hermes
+      - TERM: dumb
+      # Homebrew currently breaks while updating:
+      # https://discuss.circleci.com/t/brew-install-fails-while-updating/32992
+      - HOMEBREW_NO_AUTO_UPDATE: 1
+    steps:
+      - checkout
+      - run:
+          name: Install dependencies
+          command: |
+            brew install cmake ninja
+      - run:
+          name: Set up workspace
+          command: |
+            mkdir -p "$HERMES_WS_DIR" "$HERMES_WS_DIR/output" "$HERMES_WS_DIR/package-root"
+            ln -sf "$PWD" "$HERMES_WS_DIR/hermes"
+      - run:
+          name: Build Hermes for macOS as CocoaPods pod
+          command: |
+            cd "$HERMES_WS_DIR"
+            hermes/utils/build/configure.py --distribute --cmake-flags="$(ruby -rcocoapods-core -e 'load %{hermes/hermes.podspec}; puts HermesHelper.cmake_configuration') -DCMAKE_INSTALL_PREFIX:PATH=../destroot"
+            cmake --build ./build_release --target hermes-runtime-darwin-cocoapods-release
+      - run:
+          name: Copy artifacts
+          command: |
+            cd "$HERMES_WS_DIR"
+            cp "build_release/github"/hermes-runtime-darwin-*.tar.gz "output"
+      - run:
+          name: Checksum artifacts
+          command: |
+            cd "$HERMES_WS_DIR/output"
+            for file in *
+            do
+              shasum -a 256 "$file" > "$file.sha256"
+            done
+      - store_artifacts:
+          path: /tmp/hermes/output/
+      - persist_to_workspace:
+          root: /tmp/hermes/output/
+          paths:
+            - .
+
   test-macos:
     # CheckedMalloc.Death fails in release mode, so build a debug version
     macos:
@@ -254,7 +301,7 @@ jobs:
             cmake --build ./build
             cmake --build ./build --target check-hermes
 
-  test-macos-podspec:
+  test-macos-cocoapods:
     macos:
       xcode: "10.3.0"
     environment:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -203,7 +203,7 @@ jobs:
           name: Build Hermes for macOS
           command: |
             cd "$HERMES_WS_DIR"
-            hermes/utils/build/configure.py --distribute
+            hermes/utils/build/configure.py --distribute --cmake-flags='-DHERMES_BUILD_APPLE_FRAMEWORK:BOOLEAN=false'
             cmake --build ./build_release --target github-cli-release
       - run:
           name: Copy artifacts
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,7 @@ workflows:
       - android
       - linux
       - macos
-      - macos-cocoapods
+      - build-macos-runtime
       - windows
       - npm:
           requires:
@@ -30,7 +30,7 @@ workflows:
             - windows
       - test-linux
       - test-macos
-      - test-macos-cocoapods
+      - test-macos-runtime-build-and-cocoapods-integration
       - test-e2e:
           requires:
             - npm
@@ -226,7 +226,7 @@ jobs:
           paths:
             - .
 
-  macos-cocoapods:
+  build-macos-runtime:
     macos:
       xcode: "10.3.0"
     environment:
@@ -301,7 +301,7 @@ jobs:
             cmake --build ./build
             cmake --build ./build --target check-hermes
 
-  test-macos-cocoapods:
+  test-macos-runtime-build-and-cocoapods-integration:
     macos:
       xcode: "10.3.0"
     environment:
--- a/hermes.podspec
+++ b/hermes.podspec
@@ -11,8 +11,12 @@ module HermesHelper
     "command -v #{bin} > /dev/null 2>&1"
   end
 
+  def self.cmake_configuration
+    "-DHERMES_ENABLE_DEBUGGER:BOOLEAN=true -DHERMES_ENABLE_FUZZING:BOOLEAN=false -DHERMES_ENABLE_TEST_SUITE:BOOLEAN=false"
+  end
+
   def self.configure_command
-    "./utils/build/configure.py #{BUILD_TYPE == :release ? "--distribute" : "--build-type=Debug"} --cmake-flags='-DHERMES_ENABLE_DEBUGGER:BOOLEAN=true -DHERMES_ENABLE_FUZZING:BOOLEAN=false -DHERMES_ENABLE_TEST_SUITE:BOOLEAN=false -DCMAKE_INSTALL_PREFIX:PATH=../destroot' build"
+    "./utils/build/configure.py #{BUILD_TYPE == :release ? "--distribute" : "--build-type=Debug"} --cmake-flags='#{HermesHelper.cmake_configuration} -DCMAKE_INSTALL_PREFIX:PATH=../destroot' build"
   end
 end
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -710,3 +710,16 @@ add_custom_command(
 add_custom_target(
   github-cli-release
   DEPENDS ${HERMES_GITHUB_DIR}/${HERMES_CLI_GITHUB_FILE})
+
+add_custom_target(
+  hermes-runtime-darwin-cocoapods-release
+  DEPENDS install
+  COMMAND
+    mkdir -p ${HERMES_GITHUB_DIR}/package-root
+  COMMAND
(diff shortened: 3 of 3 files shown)


alloy:
@willholen So this is the correct package layout for CocoaPods to ingest directly (without a root directory), but I just realised that that may not be the best way to package it for the npm step. I’ll take a look at the contents of the other job artifacts to see what it should be.

mhorowitz:
Just to make sure I understand, with this change, in circleci, macos-cocoapods is analogous to android, in that those jobs build the necessary runtime artifacts for an application using Hermes, right?

If so, I wonder if it would be clearer to rename android to android-runtime, and macos-cocoapods to macos-runtime.

In the future, I think we should consider making the names of the cli and runtime artifacts and the circleci jobs which build them more consistent, but that need not delay this PR.  For now, I think the single rename above would clarify my immediate confusion.

alloy:
@mhorowitz Correct. I’ll update those job names 👍

alloy:
@mhorowitz I went a little further in renaming jobs, hope that’s ok.

alloy:
I’ll move on to the NPM package creation part, which I can either add to this PR or you can merge it as a self-contained change. Whichever comes first and is most convenient to you.

mhorowitz:
@alloy It looks like you didn't rename test-macos-cocoapods, and it's not obvious to the casual observer what this is testing (cli or runtime.  or both?)

One of the things which remains a bit confusing (although perhaps improved) is the correspondence between the circleci job names and the github release tarballs they generate.  The hermes-cli-* and hermes-runtime-* tarballs at https://github.com/facebook/hermes/releases correspond one to one with circleci tasks which generate them.  Your current names make it a bit more clear what's happening if you know this, but I think it would be nice if the circleci names and release files had more directly corresponding names.  With the PR as of now, hermes-cli-linux-... is built by the build-cli-tools-for-linux job, which is not unclear, but also unnecessarily different.  hermes-cli-darwin-... is built by build-cli-tools-for-macos, which adds a darwin vs macos difference.  The tarball names are based on CMAKE_SYSTEM_NAME, so keeping those suffixes will simplify the build files, but switching the literal parts would be ok with me. I don't have strong preferences on the specifics, but consistency is valuable.

alloy:
Agreed on consistency, just wasn’t sure how far I was supposed to go. I can do that, let me know if you want it in this PR or a follow-up one.

mhorowitz:
In this PR, I'm fine if all we do is use initial names for macos-cocoapods and test-macos-cocoapods which are more clear what they build/test, even if consistency is still lacking.  Improving consistency can be in this PR, or later, but in the interest of keeping PRs small and isolated, a follow-up may be better.  I don't see value in renaming existing jobs to something differently inconsistent in this PR.

alloy:
> In this PR, I'm fine if all we do is use initial names for macos-cocoapods and test-macos-cocoapods which are more clear what they build/test

@mhorowitz My bad, I reverted that change and redid it for just those two jobs 👍

.circleci/config.yml
```diff
@@ -225,6 +226,52 @@ jobs:
...
+          name: Install dependencies
+          command: |
+            brew install cmake ninja
+      - run:
+          name: Set up workspace
+          command: |
+            mkdir -p "$HERMES_WS_DIR" "$HERMES_WS_DIR/output" "$HERMES_WS_DIR/package-root"
+            ln -sf "$PWD" "$HERMES_WS_DIR/hermes"
+      - run:
+          name: Build Hermes for macOS as CocoaPods pod
```
mhorowitz, line 250: So the process for building a pod doesn't involve running the pod tool?

hermes.podspec
```diff
@@ -11,8 +11,12 @@ module HermesHelper
     "command -v #{bin} > /dev/null 2>&1"
   end
 
+  def self.cmake_configuration
+    "-DHERMES_ENABLE_DEBUGGER:BOOLEAN=true -DHERMES_ENABLE_FUZZING:BOOLEAN=false -DHERMES_ENABLE_TEST_SUITE:BOOLEAN=false"
+  end
+
   def self.configure_command
-    "./utils/build/configure.py #{BUILD_TYPE == :release ? "--distribute" : "--build-type=Debug"} --cmake-flags='-DHERMES_ENABLE_DEBUGGER:BOOLEAN=true -DHERMES_ENABLE_FUZZING:BOOLEAN=false -DHERMES_ENABLE_TEST_SUITE:BOOLEAN=false -DCMAKE_INSTALL_PREFIX:PATH=../destroot' build"
+    "./utils/build/configure.py #{BUILD_TYPE == :release ? "--distribute" : "--build-type=Debug"} --cmake-flags='#{HermesHelper.cmake_configuration} -DCMAKE_INSTALL_PREFIX:PATH=../destroot' build"
```
mhorowitz, line 19: Since I barely remember what little I ever knew about cocoapods, how do all the pieces here come together?

If I understand correctly, the plan is that we will provide an npm which includes the prebuilt hermes pod, and this will be consumed by mac apps which use hermes.  Is this right?

I see where circleci builds the pod above, but I also see code below (lines 44,46) where the same build commands appear in the podspec.  When are these commands used?

For an application which consumes hermes, how does cocoapods know where to find the prebuilt hermes framework?

.circleci/config.yml
alloy: A pod is either source or binaries with a podspec that knows how to include those. Typically the podspecs are served from a centralized spec repo, in which case you might use the `pod` tool to push the spec on release. In the case of RN, though, they are shipped within npm packages, so the only work that needs to be done is to build the binaries and package them.

.circleci/config.yml
```diff
@@ -225,6 +226,52 @@ jobs:
...
+          command: |
+            cd "$HERMES_WS_DIR"
+            hermes/utils/build/configure.py --distribute --cmake-flags="$(ruby -rcocoapods-core -e 'load %{hermes/hermes.podspec}; puts HermesHelper.cmake_configuration') -DCMAKE_INSTALL_PREFIX:PATH=../destroot"
```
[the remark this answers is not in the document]
alloy: No; the podspec includes this and is able to build the binaries in case somebody decides to install hermes from source. No interaction of the user is needed.

.circleci/config.yml
mhorowitz: Since the podspec seems to have code for building the Framework (although just by calling cmake), is there a reason to write out the similar logic here?  It just seems like a violation of DRY.

hermes.podspec
alloy: As mentioned above, this is here so users can install from source, if need be. The releases we cut just come with the artefacts in the same place as where this podspec expects them to be (e.g. `vendored_framework`)

alloy: And correct, the plan is to ship them in npm bundles, as most other RN deps are distributed. (I am personally not convinced we should keep doing it in such a roundabout way, but for now it works.)

.circleci/config.yml
alloy: There is no `pod` cli feature to invoke an arbitrary podspec’s  `prepare_command`. I did try to counter the DRY violation by refactoring the podspec such that we could load it here and get the configuration that matters (I don’t consider the build and destroot locations to be part of that), so those only need to be maintained in a single place.

hermes.podspec
mhorowitz: Oh, I think I see.  The app Podspec would point to `node_modules/hermes-engine-darwin` (maybe with some `..`s), then? I think that was the bit I hadn't figured out.

alloy: Yep, that’s it 👍 See the integration test for an example of what that syntax would look like (although in that test the binaries don’t exist yet and so it will build them through cmake) https://github.com/facebook/hermes/blob/f128811121ce711cd0b99c2abe7dfd649168cca3/test/ApplePlatformsIntegrationTestApp/Podfile#L6

alloy: For posterity, there’s no real need to distribute this as a npm package, but we’re doing this simply for consistency with the other RN deps (as discussed with @wilholen).

.circleci/config.yml
mhorowitz: Bummer.  I guess we could the logic in prepare_command into a shell script which is run by cocoapods or circleci.  but that can be a followup, too.  I've asked enough questions already.  Thanks for educating me.

mhorowitz:
At this point, I'm satisifed.  @willholen can you import one more time?

image

tmikov:
@alloy merging this PR broke debugging in my IDE (CLion). Any ideas how to fix it? I am completely blocked by this.

PR #296, opened 2020-07-16
[cmake] Add explicit option to build Apple dSYM

alloy:
## Summary

Add `HERMES_BUILD_APPLE_DSYM` option to build a dSYM bundle for the libhermes target on Apple platforms.

This will work with any of the build types and is off by default.

Installing the tools with the install/strip target will ensure all tools and the runtime lib are stripped of debug symbols, but leaving the dSYM bundle in tact.

### Build

```bash
./src/utils/build/configure.py --distribute --cmake-flags='-DHERMES_BUILD_APPLE_DSYM:BOOLEAN=true -DCMAKE_INSTALL_PREFIX:PATH=../destroot_release' build_release
cmake --build ./build_release
```

### Install without stripping

```bash
cmake --build ./build_release --target install
nm -a destroot_release/bin/hermesc | wc -l
   27943
```

### Install with stripping

```bash
cmake --build ./build_release --target install/strip
nm -a destroot_release/bin/hermesc | wc -l
     250
```

…and dSYM DWARF metadata is maintained:

```bash
dwarfdump --statistics destroot_release/Library/Frameworks/hermes.framework.dSYM
{"version":3,"file":"destroot_release/Library/Frameworks/hermes.framework.dSYM/Contents/Resources/DWARF/hermes","format":"Mach-O 64-bit x86-64","source functions":30305,"source functions with location":30302,"inlined functions":172725,"inlined funcs with abstract origins":172725,"unique source variables":79276,"source variables":353690,"variables with location":232195,"call site entries":186409,"scope bytes total":19161949,"scope bytes covered":10500176,"total function size":1763513,"total inlined function size":998375,"total formal params":300264,"formal params with source location":166067,"formal params with type":300264,"formal params with binary location":200407,"total vars":38809,"vars with source location":38385,"vars with type":38809,"vars with binary location":22161}
```

PR #297, opened 2020-07-16
[ci] Create hermes-engine-darwin npm package

alloy:
⚠️ _This includes the changes of #296, so I’ll have to rebase this PR once that lands. The changes that pertain to this PR are in https://github.com/facebook/hermes/commit/9da61f7a9da01b5bd49e95dca0b98a5204210790_

## Summary

Adds hermes-engine-darwin to the npm CI job.

Notably the symbolic links need to be stripped from the framework bundle, as npm by design does not support symbolic links to exist in packages. However, seeing as we already package distinct versions of the framework in distinct npm packages, there isn't any need to support multiple versions of Hermes in a single framework bundle; so this should be fine.

## Test Plan

TODO: I’ll need to backport all changes to Hermes v0.4.1 in order to pull it into RN macOS v0.62. In the meantime I want to see if CI is correctly cutting this package.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -712,9 +712,9 @@ add_custom_target(
   COMMAND
     cp -R -f ${CMAKE_INSTALL_PREFIX} ${HERMES_PKG_ROOT}/destroot
   COMMAND
-    rm -r ${HERMES_PKG_FRAMEWORK}/{Headers,Resources,hermes} && mv ${HERMES_PKG_FRAMEWORK}/Versions/0/* ${HERMES_PKG_FRAMEWORK}/ && rm -r ${HERMES_PKG_FRAMEWORK}/Versions && install_name_tool -id @rpath/hermes.framework/hermes ${HERMES_PKG_FRAMEWORK}/hermes
+    find ${HERMES_PKG_ROOT}/destroot/bin -type f ! -name 'hermesc' -delete
   COMMAND
-    cp ${CMAKE_CURRENT_SOURCE_DIR}/hermes.podspec ${CMAKE_CURRENT_SOURCE_DIR}/LICENSE ${CMAKE_CURRENT_SOURCE_DIR}/README.md ${HERMES_PKG_ROOT}
+    cp ${CMAKE_CURRENT_SOURCE_DIR}/hermes.podspec ${CMAKE_CURRENT_SOURCE_DIR}/LICENSE ${HERMES_PKG_ROOT}
   COMMAND
-    tar -C ${HERMES_PKG_ROOT}/ -czhf ${HERMES_GITHUB_DIR}/hermes-runtime-${HERMES_GITHUB_SYSTEM_NAME}-v${HERMES_RELEASE_VERSION}.tar.gz .
+    tar -C ${HERMES_PKG_ROOT}/ -czvf ${HERMES_GITHUB_DIR}/hermes-runtime-${HERMES_GITHUB_SYSTEM_NAME}-v${HERMES_RELEASE_VERSION}.tar.gz .
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -700,16 +700,21 @@ add_custom_target(
   github-cli-release
   DEPENDS ${HERMES_GITHUB_DIR}/${HERMES_CLI_GITHUB_FILE})
 
+set(HERMES_PKG_ROOT ${HERMES_GITHUB_DIR}/package-root)
+set(HERMES_PKG_FRAMEWORK ${HERMES_PKG_ROOT}/destroot/Library/Frameworks/hermes.framework)
+
 add_custom_target(
   hermes-runtime-darwin-cocoapods-release
   COMMAND
     ${CMAKE_COMMAND} --build . --target install/strip
   COMMAND
-    mkdir -p ${HERMES_GITHUB_DIR}/package-root
+    mkdir -p ${HERMES_PKG_ROOT}
   COMMAND
-    cp -R -f ${CMAKE_INSTALL_PREFIX} ${HERMES_GITHUB_DIR}/package-root/destroot
+    cp -R -f ${CMAKE_INSTALL_PREFIX} ${HERMES_PKG_ROOT}/destroot
   COMMAND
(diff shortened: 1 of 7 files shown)


alloy:
Note to self: need to fix the `rpath` now that the binary no longer exists in a nested version dir.

alloy:
@willholen Seeing as I don’t have the ability to SSH into the CI jobs, could you perhaps do so and debug [the prepack script](https://github.com/facebook/hermes/pull/297/files#diff-09eaea1b3e184fff561c520f2d1b4ddbR12)? It works fine locally, but fails on CI like so https://app.circleci.com/pipelines/github/facebook/hermes/1064/workflows/a5cc7cf1-7726-4fa3-aba2-82a98d917f62/jobs/7208/steps

willholen:
I downloaded the artifact from https://7273-154201259-gh.circle-artifacts.com/0/tmp/hermes/output/hermes-engine-darwin-v0.5.0.tgz and it looks complete as far as I can tell. 

@alloy The title still says "[DO NOT MERGE]". Is there still testing and tweaking remaining?

alloy:
@willholen Great! Yeah I first want to test integration into RN macOS, but have to do so with the other build of the 0.4 branch. Will update once I’ve done that.

alloy:
Ok, the framework works, but unfortunately removing the symbolic link meant lldb no longer automatically can find the dSYM bundle. So looks like I’ll have to revert the change of removing the symlink and instead rebuild the symlinks in a `postinstall` script of the packages `package.json`.

willholen:
@alloy Oh, that's frustrating :(

I don't know how CocoaPods use symlinks, but some generic thoughts are:
* If it's complicated by the version numbers, the symlink could just be hard coded to "current" or "0.0.0" or something since the package is already versioned through NPM
* Files could simply be duplicated (e.g. via `rsync -L`). It would take up a bit more space, but it would presumably cause fewer issues if a project is cloned on or copied via Windows
* The framework could be added to the NPM as a `.tar.gz` instead of as its unpacked contents. That way, MacOS can unpack it complete with symlinks, while Windows can ignore it (since it can't use it anyways) and have a smaller on-disk footprint.

alloy:
> The framework could be added to the NPM as a .tar.gz instead of as its unpacked contents. That way, MacOS can unpack it complete with symlinks, while Windows can ignore it (since it can't use it anyways) and have a smaller on-disk footprint.

Oh yeah, interesting thought—I like it :+1:

alloy:
Ok, fixed the issue by adding an extra layer of tarball, as suggested. This should be good to go now 👍

CMakeLists.txt
```diff
@@ -700,16 +700,21 @@ add_custom_target(
...
   COMMAND
     ${CMAKE_COMMAND} --build . --target install/strip
   COMMAND
-    mkdir -p ${HERMES_GITHUB_DIR}/package-root
+    mkdir -p ${HERMES_PKG_ROOT}
+  COMMAND
+    cp -R -f ${CMAKE_INSTALL_PREFIX} ${HERMES_PKG_ROOT}/destroot
   COMMAND
-    cp -R -f ${CMAKE_INSTALL_PREFIX} ${HERMES_GITHUB_DIR}/package-root/destroot
+    find ${HERMES_PKG_ROOT}/destroot/bin -type f ! -name 'hermesc' -delete
```
mhorowitz, line 715: This feels weird.  Instead of doing this, could you make 713 copy the one file you want?  I don't know what else is in destroot, but making the contents of the package inclusionary rather than exclusionary seems easier to understand.

npm/hermes-engine-darwin/unpack-tarball.js
```diff
@@ -0,0 +1,19 @@
+const path = require("path");
+const fs = require("fs");
+const child_process = require("child_process");
+
+if (
+  process.platform === "darwin" &&
+  !fs.existsSync(path.join(__dirname, "destroot"))
+) {
+  const tarball = fs.readdirSync(__dirname).find(function (entry) {
+    return entry.endsWith(".tar.gz");
```
mhorowitz, line 10: Can we also check startsWith("hermes-runtime-darwin-"), just in case something else ends up in here?

npm/hermes-engine-darwin/unpack-tarball.js
```diff
@@ -0,0 +1,19 @@
...
+  });
+  if (!tarball) {
+    throw new Error("Could not locate tarball");
+  }
+  child_process.execFileSync("/usr/bin/tar", [
```
mhorowitz, line 15: Why an explicit path?  Some people might install tools in /usr/local/bin, for example.  If you just say "tar", will node search the path?

CMakeLists.txt
alloy: Good point 👍

npm/hermes-engine-darwin/unpack-tarball.js
alloy: The system supplied version of `tar` is a known good version, so it seems more stable to _not_ accidentally use some other version of `tar` that the user might have installed.

To your second question, seems like Node will indeed search the `PATH` env variable when not give an absolute path.

npm/hermes-engine-darwin/unpack-tarball.js
alloy: I figured keeping this flexible would make it future-proof against renaming the tarball, as this happens in a different build job, but you’re right that it might be bad if some other artefact that this script doesn’t expect got included too.

npm/hermes-engine-darwin/unpack-tarball.js
willholen: @alloy I definitely see your point, though if a user explicitly wants to supply their own `tar` I think we should let them.

Our CircleCI jobs make some use of this kind of thing, and my `~/bin` has several overrides that I want used and therefore put first in my PATH.

alloy: I personally find it a little less intuitive for an opaque installation, such as untarring from inside a post-install step, to possibly break due to such customizabitility, but on the other hand I don’t really expect people to install a custom tar without knowing what they’re doing, so will make the change 👍

alloy: Done

alloy:
@willholen @mhorowitz Thanks for your help getting this over the finish line! :pray:

PR #304, opened 2020-07-22
fix(docs): Update README.md

iamdarshshah:
### Summary

- [x] Aligning header elements to the center
- [x] Updated Hermes logo position

> This makes it look a lot cleaner.


neildhar:
Hi, thanks for doing this! Unfortunately, we'd like to avoid putting HTML in the README so we can keep the plaintext version as readable as possible.

iamdarshshah:
Thanks @neildhar for reviewing it. Yes, I agree with your point. But if we look from the accessibility point of view, then using `Heading` tags or any `HTML` tags would be beneficial.

Would like to know your thoughts regarding accessibility. 

Thanks 👍

neildhar:
The actual rendering of the markdown in the browser is done in HTML and does use heading tags so that shouldn't be an issue.

PR #305, opened 2020-07-23
Added namespace std

mohitsinha04:
Have Added namespace std and eliminate std:: from vector for better Readability of Code.


tmikov:
Adding `using namespace std::` in a header is not a good idea.

PR #328, opened 2020-08-11
Added more links for better understanding and easy installation

Shanu1515:
## Summary

Added detailed instructions and installation process links for more clarity and information.

## Test Plan
Not required


Huxpro:
@Shanu1515 Thanks for supporting the Hermes project.  We've evaluated this PR, and we don't believe it improves the documentation, so we've decided not to accept it.

PR #332, opened 2020-08-17
Build Hermes for iOS 

grabbou:
## Summary

This PR allows building and running Hermes on iOS devices. It extends the prior work done by @alloy that enabled Hermes to run on MacOS.

In a nutshell, the Hermes podspec was extended to include a dedicated framework for iOS devices. That framework itself is a universal framework that includes binaries for different iOS architectures and an iPhone simulator. 

Every target is built into `build_<name>`, e.g. `build_iphoneos` and `build_iphonesimulator`. This convention (and helpers that I have created) allows for quick and easy extension to support iPad as well (and probably should happen as a part of this or next PR).

Here's dump of `destroot` folder to give you an idea how frameworks are located:
```
destroot/Library/Frameworks
├── iphoneos
│   ├── hermes.framework
│   └── hermes.framework.dSYM
├── iphonesimulator
│   ├── hermes.framework
│   └── hermes.framework.dSYM
└── macosx
    ├── hermes.framework
    └── hermes.framework.dSYM
```

Things work locally and the test mobile application prints "Hello World" just fine.

I have started an appropriate discussion regarding them around the lines where the todos are located.

As a follow-up, 

## Follow-up

- [ ] a PR to React Native to allow using Hermes as a default engine
- [ ] https://github.com/facebook/hermes/pull/332#issuecomment-678109796

## Todo

- [x] https://github.com/facebook/hermes/pull/332#issuecomment-678111329

## Test Plan

1. Clone the project
2. `cd test/ApplePlatformsIntegrationTestApp/` && `pod install`
3. Open `ApplePlatformsIntegrationTestApp.xcworkspace`
4. Choose `ApplePlatformsIntegrationTestMobileApp` target
5. Run on a device or a simulator


.circleci/config.yml
```diff
@@ -250,9 +250,7 @@ jobs:
       - run:
           name: Build Hermes for macOS as CocoaPods pod
           command: |
-            cd "$HERMES_WS_DIR"
-            hermes/utils/build/configure.py --distribute --cmake-flags="$(ruby -rcocoapods-core -e 'load %{hermes/hermes.podspec}; puts HermesHelper.cmake_configuration') -DCMAKE_INSTALL_PREFIX:PATH=../destroot"
-            cmake --build ./build_release --target hermes-runtime-darwin-cocoapods-release
+            exit 1
```
grabbou, line 253: Left as a todo. I need to replace `--cmake-flags="$(ruby -rcocoapods-core -e 'load %{hermes/hermes.podspec}; puts HermesHelper.cmake_configuration') -DCMAKE_INSTALL_PREFIX:PATH=../destroot"` with `./build-apple-framework.sh`.

It is trivial, but I want to take time to connect that with building iOS pieces as well.

.circleci/config.yml
```diff
@@ -317,10 +315,20 @@ jobs:
...
+      - run:
+          name: Test MacOS application
+          command: |
+            . ./run.sh
+            test_apple_target "ApplePlatformsIntegrationTestApp"
+      - run:
+          name: Test iPhone application
+          command: |
+            . ./run.sh
+            test_apple_target "ApplePlatformsIntegrationMobileTestApp"
```
grabbou, line 331: While this test runs just fine, it will fail. I don't know how to assert the output yet. Please check the comment in the `run.sh` file.

test/ApplePlatformsIntegrationTestApp/ApplePlatformsIntegrationTestMobileApp/AppDelegate.mm
```diff
@@ -0,0 +1,30 @@
...
+
+- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
+    std::string s = "print('Hello world -- son of Maia and Zeus');";
+    std::string url = "http://example/js";
+    facebook::hermes::HermesRuntime::DebugFlags flags;
+    auto runtime = facebook::hermes::makeHermesRuntime();
+    runtime->debugJavaScript(s, url, flags);
+    exit(0);
+    return YES;
+}
```
grabbou, line 28: This is just a simple copy-paste from MacOS application. While it works, it is quite tricky to access console output (just like from a Mac app) and check its contents.

Any ideas here? I'd prefer to keep it simple and not go into a fully-fledged test suite. 

@alloy

test/ApplePlatformsIntegrationTestApp/ApplePlatformsIntegrationTestMobileApp/AppDelegate.h
```diff
@@ -0,0 +1,14 @@
+//
+//  AppDelegate.h
+//  ApplePlatformsIntegrationTestMobileApp
+//
+//  Created by Michał Grabowski on 17/08/2020.
+//
```
alloy, line 6: Can you remove these?

CMakeLists.txt
```diff
@@ -58,6 +58,10 @@ if (POLICY CMP0026)
   cmake_policy(SET CMP0026 OLD)
 endif()
 
+# Has to be set before `project` as per documentation
+# https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html
+set(CMAKE_OSX_SYSROOT ${HERMES_APPLE_TARGET_PLATFORM})
```
alloy, line 63: What does this do when building eg Android?

test/ApplePlatformsIntegrationTestApp/ApplePlatformsIntegrationTestMobileApp/AppDelegate.mm
alloy: Good question. So currently you’re not doing any testing yet, right? Or did I overlook it?

CMakeLists.txt
grabbou: No effect (as other "OSX" including flags)

test/ApplePlatformsIntegrationTestApp/ApplePlatformsIntegrationTestMobileApp/AppDelegate.mm
grabbou: My only testing is printing this hello world string (just like on MacOS). It works, but I just realized it's not as easy to get the console output here. 

Possible ways to go forward:
- get the console output (there are some hacks)
- rewrite it to use XCTest or something similar

CMakeLists.txt
grabbou: (haven't tested that, just referring to what cmake docs say. However, there were times when they didn't match the reality)

test/ApplePlatformsIntegrationTestApp/ApplePlatformsIntegrationTestMobileApp/AppDelegate.h
grabbou: Done

test/ApplePlatformsIntegrationTestApp/ApplePlatformsIntegrationTestMobileApp/AppDelegate.mm
grabbou: Resolved by providing XCTest.

grabbou:
Update: I have provided cross-platform test suite using XCTest. It contains two separate targets that share the test suite. As a result, a Mac application and iOS application are no longer present. This approach is cleaner than looking at `stdout`, which wasn't as easy and clean to implement on iOS side.

grabbou:
Update: I have shipped CircleCI updates.

Notable changes:
- unified CircleCI setup across both "test" and "build" jobs that target Apple
- used cache to preserve build folders from "test" and successfully use them in "build" to save A LOT of time (around 30 minutes from past builds)
- cleaned up the use of `HERMES_WS_DIR` and doing a lot of `cd` throughout CircleCI in favor of more declarative `working_directory` and just using the default directory too
- added few comments to places that confused me

I see a lot of improvements that potentially could be done to CircleCI configuration, but want to keep that for another PR.

grabbou:
Looks like e2e tests are failing (on master too). Rather than that, this PR is ready to review & merge. I just downloaded and checked the final artifact (`tar.gz`) and it contains both frameworks for iPhone and Mac.

hermes.podspec
```diff
@@ -29,26 +17,36 @@ Pod::Spec.new do |spec|
   spec.license     = { type: "MIT", file: "LICENSE" }
   spec.author      = "Facebook"
   spec.source      = { git: "https://github.com/facebook/hermes.git", tag: "v#{spec.version}" }
-  spec.platforms   = { :osx => "10.14" }
+  spec.platforms   = { :osx => "10.14", :ios => "10.0" }
```
grabbou, line 20: I think we should update this to something more accurate. What is the minimum that Hermes can run? With React Native being its primary consumer, I think we could align that requirement with what React Native supports.

alloy: Yup, align with min supported target of RN 👍

utils/build-apple-framework.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
```
alloy, line 1: Does this script really require `bash` instead of just `sh`?

alloy: I guess it might be for features such as testing if a command exists? It’s no biggie, every macOS installation should support this anyways.

alloy:
The test plan steps are outdated and, because of how git only tracks files, it’s also necessary to do some manual cleaning first:

1. Clone the project
1. `cd test/ApplePlatformsIntegrationTestApp`
1. `rm -rf ApplePlatformsIntegrationTestApp.*`
1. `pod install --verbose`
1. `open ApplePlatformsIntegrationTests.xcworkspace`
1. Choose `ApplePlatformsIntegrationTestMobileApp` target
1. Run on a device or a simulator (Command+U)

alloy:
Currently in the process of testing this on a iOS device, but had to jump through some hoops and update Xcode, but alas I will have to continue that tonight when I get back to work.

utils/build-apple-framework.sh
willholen: This script uses the `function` keyword, `local`, `[[ ]]` and arrays, all of which are bash extensions

alloy:
Alright, I was able to verify that it works on device too. I made a few additions that I’ve PR’ed [here](https://github.com/callstack/hermes/pull/1).

utils/build-apple-framework.sh
alloy: I should rephrase; I meant more, are those features all needed? But again, it actually doesn’t really matter, all macOS installations will have the dependency.

grabbou:
We are going to address the overall size of the artifact in a follow-up PR. One of the ideas is to separate Mac and iPhone frameworks into two separate `npm` packages. There are still a few questions around how CocoaPods handles that case that we want to check before doing any progress in that matter.

Anyway, since this is not going to ship immediately to React Native (and any other users), we don't see this being a blocker for merging this PR.

In a follow-up, I'd like to make some real measurements of the size of the artifact before/after optimizations to be sure what we're going to ship to the users.

grabbou:
Before this PR is ready to be merged, we need to provide a build phase script that strips `x86_64` slice from the binary. It is only used for being able to run the application with Hermes on an iPhone Simulator. Not doing so will result in an automatic rejection of the application from the App Store.

hermes.podspec
grabbou: Turns out its already aligned: https://github.com/facebook/react-native/blob/master/React.podspec#L39

utils/build-apple-framework.sh
grabbou: That's a good question. I started off with the script from React Native and left that header.

grabbou:
Update: Turns out CocoaPods strip invalid architectures out of the box: https://github.com/CocoaPods/CocoaPods/blob/2fb2db21016f82806d3940013bda30ffa941151b/lib/cocoapods/generator/embed_frameworks_script.rb#L123-L138.

utils/build-apple-framework.sh
alloy: Okeydokey

alloy (approved):
Having tested this on device and having reviewed the changes one more time, I’d say this looks good to me 👍 

@mhorowitz @willholen Could you import this?

alloy:
Note that if https://github.com/facebook/hermes/pull/335 gets merged before this, then this PR will need to be updated to do the same for iOS.

Huxpro:
> Note that if #335 gets merged before this, then this PR will need to be updated to do the same for iOS.

@alloy oh I didn't noticed this comment. @grabbou would you mind doing the necessary updates?

grabbou:
Updated to reflect changes in #335 in the way that work with this PR. Also, resolved some conflicts with `master` branch as a result. Should be good to go now!

alloy:
Nice; looks good to me 👍

grabbou:
I don't know why `test-apple-runtime` started breaking, will take a look.

grabbou:
The failure should be gone now. Ready to go.

alloy:
Ace. @mhorowitz can you re-import?

When this is in I can leverage it to build a fat binary for macOS x86_64 and ARM64.

EhsanSarshar:
can I now use Hermes with this pull request

kelset:
@EhsanSarshar not yet, you will have to wait for 0.64.0

prabhuvi:
How can I build hermes.framework as a static library. It's a dynamic lib right now and I am unable to figure out the changes required to generate static lib. Can someone please help?

neildhar:
@prabhuvi Hermes provides a target `hermesapi` which produces a static library with the same contents as `libhermes`. The iOS build of Hermes used by React Native is maintained by RN, and they may be able to offer better guidance on whether/how this is supported.

prabhuvi:
Thank you @neildhar. Currently if I look for the contents for hermes.framework file, it has a dynamic library. I would like to retain the cocoapod changes and have hermes-engine pod build hermes.framework through script phase but the framework/library inside should be static.

PR #335, opened 2020-08-27
[podspec] Lower macOS deployment target to 10.13

alloy:
## Summary

This follows the recent change in RN macOS:
https://github.com/microsoft/react-native-macos/pull/555

Also set this constraint during actual build of the framework.

## Test Plan

Builds fine locally and should also be green on CI.


alloy:
@willholen @Huxpro It would be great if we can also create a new v0.4.x release of this to make the instructions for enabling hermes in RN macOS easier, but only if it doesn’t cost you too much time.

PR #338, opened 2020-09-03
Intl APIs

mganandraj:
## Summary

This change implements Intl::NumberFormat APIs using libraries available in Android. The changes are made on top of the Collator implementation, hence the commit contains those changes as well, even though there is another active PR with collator changes.

On Android platform, the availability of Intl libraries varies wildly. Before API 24, Android platform used to expose a subset of ICU APIs through java.text namespace. From API 24 onwards, Android ships with a bigger subset of ICU APIs under android.icu namespace. But some important APIs such as LocaleMatcher, NumberFormatter etc. are still missing (until API 29 where things started improving ... ).

With this implementation, we try to get the best out of what is available in API 24. It is possible to directly manipulate the format patterns to achieve even better conformance, but that's going to be a lot more work. For devices running API 23 or lower, we try to support the basics, but the code is not tested.

## Test Plan

I've used test262 test data to test the implementation throughout. The test code is part of the checkin but not the test262 test data itself.


lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterICU.java
```diff
@@ -0,0 +1,358 @@
...
+            case PERCENT:
+                platformNumberFormatterICU = PlatformNumberFormatterICU.createDecimalFormatForPercent(localeObject, style);
+                break;
+            case DECIMAL:
+            case UNIT:
+                platformNumberFormatterICU = PlatformNumberFormatterICU.createDecimalFormatGerneric(localeObject, notation, compactDisplay, style);
+                break;
+            default:
+                throw new JSRangeErrorException("Unrecognized formatting style requested.");
+        }
```
mhorowitz, line 292: This pattern using nested case statements with helper functions feels like it will create an unnecessarily large footprint.  For most cases, you could just call NumberFormat.getInstance() once, after choosing an appropriate value for the ICU style.  Then you could have some special cases before and after for the weird cases.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterICU.java
```diff
@@ -0,0 +1,358 @@
...
+                    formatWidth = MeasureFormat.FormatWidth.WIDE;
+                    break;
+                case NARROW:
+                    formatWidth = MeasureFormat.FormatWidth.NARROW;
+                    break;
+                case SHORT:
+                default:
+                    formatWidth = MeasureFormat.FormatWidth.SHORT;
+                    break;
+            }
```
mhorowitz, line 176: One of the cool features of java enums is they can have data fields and methods.  So you could store the MeasureFormat.FormatWidth value as data on the enum itself, and replace this switch statement with a simple method call which is just an accessor.  See [an example here](https://gist.github.com/mhorowitz/81a1841036b46ed7fc3ea11306e44ebe) or at https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html.  In a very rudimentary test, this was quite a bit smaller than using a switch to look up a per-enum value.

There are a lot of places where you switch over an enum, where you could instead return a per-enum value.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterICU.java
```diff
@@ -0,0 +1,358 @@
...
+                    currencySymbol = currencyCode;
+                    break;
+                case SYMBOL:
+                case NARROWSYMBOL:
+                default:
+                    currencySymbol = currency.getName((ULocale) mLocaleObject.getLocale(), Currency.SYMBOL_NAME, null);
+                    break;
+            }
+
+            mDecimalFormat.getDecimalFormatSymbols().setCurrencySymbol(currencySymbol);
```
mhorowitz, line 67: getDecimalFormatSymbols is documented to return a copy of the symbols, so this seems like it shouldn't work.  Has it been tested?

lib/VM/JSLib/Intl.cpp
```diff
@@ -35,8 +37,15 @@ namespace {
 CallResult<std::u16string> stringFromJS(
     Runtime *runtime,
     PseudoHandle<> value) {
+  Handle<> valueHandle = runtime->makeHandle(std::move(value));
+  TagKind kind = valueHandle.get().getTag();
+  if(kind != StrTag && kind != ObjectTag ) {
```
avp, line 42: Please avoid directly comparing tags like this unless absolutely necessary.

Use the `isString()` and `isObject()` functions instead. (`valueHandle->isString()` will do the trick)

lib/VM/JSLib/Intl.cpp
```diff
@@ -165,9 +181,6 @@ CallResult<HermesValue> partsToJS(
   uint64_t index = 0;
   for (auto &part : *result) {
     CallResult<Handle<JSObject>> partRes = partToJS(runtime, std::move(part));
-    if (LLVM_UNLIKELY(partRes == ExecutionStatus::EXCEPTION)) {
```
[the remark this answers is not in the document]
mganandraj: This was a mistake.. Reverted .. Thanks!

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformCollator.java
```diff
@@ -0,0 +1,150 @@
...
+
+    public static List<String> filterLocales (List<String> requestedLocales, String matcher) throws JSRangeErrorException {
+        ArrayList<String> supportedLocales = new ArrayList<>();
+        for (String candidateLocale : requestedLocales) {
+            ArrayList<String> canonicalCandidateLocaleList = new ArrayList<>();
+            canonicalCandidateLocaleList.add(candidateLocale);
+
+            LocaleMatcher.LocaleMatchResult localeMatchResult;
+            // TODO :: We need to refactor this to avoid platform checks ..
+            // Note : "getAvailableLocales" call is expensive and we should try to reuse the results ..
```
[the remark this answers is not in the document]
mganandraj: Totally agreed.. I've made substantial refactoring aroudn the locale matching and resolution in the latest iteration .. Please take a look and let me know.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateFormatter.java
```diff
@@ -0,0 +1,43 @@
...
+            result.resolvedDesiredLocale = localeMatchResult.matchedRequestedLocale;
+
+            // Note that we copy only collation extension as we know that "collation" is the only relevant keyword for icu.RuleBasedCollator;
+            // ; as returned by RuleBasedCollator.getKeywords()
+            // TODO: Spec requires & to be future proof, this should be dynamic.
+            // copyCollationExtension(result.resolvedDesiredLocale, result.resolvedLocale);
+
+            return result;
+        } else {
+            Locale[] availableLocalesArray = android.icu.text.DateFormat.getAvailableLocales();
```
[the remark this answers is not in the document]
mganandraj: THis code is gone with the latest refactorings .. Please take a look and let me know.

lib/VM/JSLib/Intl.cpp
```diff
@@ -251,6 +264,8 @@ const OptionData kCollatorOptions[] = {
 };
 
 const OptionData kDTFOptions[] = {
+        {u"dateStyle", platform_intl::Option::Kind::String, 0},
+        {u"timeStyle", platform_intl::Option::Kind::String, 0},
```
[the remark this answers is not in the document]
mganandraj: I was referring to MDN initially to understand the API surface and met with these options there .. 
After reading your comment I searched around and realized that it is in proposal state : https://tc39.es/proposal-intl-datetime-style/
I still tried to keep it but is making the code a bit complicated when dealing with "hour" field .. Hence, I'm currently removing all the code for these options ..

lib/VM/JSLib/Intl.cpp
```diff
@@ -307,6 +322,10 @@ CallResult<platform_intl::Options> normalizeOptions(
     const OptionData optionData[]) {
   platform_intl::Options ret;
 
+  if(options->isNull())
```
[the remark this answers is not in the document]
mganandraj: Awesome .. Please fix this !

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterICU.java
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        return platformNumberFormatterICU;
+    }
+
+
+    public static PlatformNumberFormatterICU createDecimalFormatForCurrency(ILocaleObject localeObject, IPlatformNumberFormatter.CurrencySign currencySign, IPlatformNumberFormatter.Style style) throws JSRangeErrorException {
+        switch (currencySign) {
+            case ACCOUNTING:
+                return new PlatformNumberFormatterICU((DecimalFormat) android.icu.text.NumberFormat.getInstance((ULocale) localeObject.getLocale(), android.icu.text.NumberFormat.ACCOUNTINGCURRENCYSTYLE),
```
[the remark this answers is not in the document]
mganandraj: Added dynamic type check all the casts .. Thanks!

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatter.java
```diff
@@ -0,0 +1,59 @@
...
+
+import android.icu.util.ULocale;
+import android.os.Build;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class PlatformNumberFormatter {
+
+    // This is an expensive implementation.. Must be optimized.
```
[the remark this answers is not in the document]
mganandraj: It is gone after the latest refactoring .. Please take a look and let me know ..

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatter.java
```diff
@@ -0,0 +1,59 @@
...
+
+            if(localeMatchResult.matchedLocale != null && !localeMatchResult.isDefault) {
+                supportedLocales.add(localeMatchResult.matchedRequestedLocale.toCanonicalTag());
+            }
+        }
+
+        return supportedLocales;
+    }
+
+    public static PlatformCollator.LocaleResolutionResult resolveLocales(List<String> locales, String localeMatcher) throws JSRangeErrorException {
```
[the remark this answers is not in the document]
mganandraj: It is gone after the latest refactoring .. Please take a look and let me know ..

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterAndroid.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.Currency;
+import java.util.Locale;
+
+import static com.facebook.hermes.intl.IPlatformNumberFormatter.Style.CURRENCY;
+
+public class PlatformNumberFormatterAndroid implements IPlatformNumberFormatter {
+
+    private java.text.Format  mFinalFormat;
+    private DecimalFormat mDecimalFormat;
+    private ILocaleObject mLocaleObject;
```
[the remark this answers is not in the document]
mganandraj: Fixed.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterAndroid.java
```diff
@@ -0,0 +1,152 @@
...
+import java.text.AttributedCharacterIterator;
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
+import java.text.NumberFormat;
+import java.util.Currency;
+import java.util.Locale;
+
+import static com.facebook.hermes.intl.IPlatformNumberFormatter.Style.CURRENCY;
+
+public class PlatformNumberFormatterAndroid implements IPlatformNumberFormatter {
```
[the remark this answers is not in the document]
mganandraj: In the latest iteration, all three services are following a very similar pattern with a lot of code shared between them.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterICU.java
```diff
@@ -0,0 +1,358 @@
...
+import static android.icu.text.NumberFormat.SCIENTIFICSTYLE;
+import static com.facebook.hermes.intl.IPlatformNumberFormatter.Style.CURRENCY;
+
+// An implementation of NumberFormat services using ICU4J APIs available in Android from API 24 onwards.
+// We could have a much richer implementation on newer API versions with // Note :: https://developer.android.com/reference/android/icu/number/NumberFormatter which was introduced in API 30.
+//
+// The implementation has at least the following deficiencies,
+// 1. SignDisplay attribute implementation is partly hacky, and doesn't work in many cases.
+// 2. formatToParts does report the whole formatted string as a "literal"
+public class PlatformNumberFormatterICU implements IPlatformNumberFormatter {
```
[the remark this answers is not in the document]
mganandraj: Sticking to "ICU" everywhere .. Thanks!

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterICU.java
```diff
@@ -0,0 +1,358 @@
...
+
+    private java.text.Format mFinalFormat;
+    private DecimalFormat mDecimalFormat;
```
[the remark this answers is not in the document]
mganandraj: Fixed.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterICU.java
mganandraj: Fixed most of the clutter in the latest iteration .. Please take a look and let me know.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterICU.java
mganandraj: Thanks! I'm heavily leveraging this pattern in the latest iteration ..

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterICU.java
mganandraj: My assumption was that it returns a reference.. I guess it was not tested .. I've changed the code to set the updated symbols back to the format object ..

lib/VM/JSLib/print.cpp
```diff
@@ -33,6 +35,14 @@ CallResult<HermesValue> print(void *, Runtime *runtime, NativeArgs args) {
     llvh::outs() << StringPrimitive::createStringView(
                         runtime, runtime->makeHandle(std::move(*res)))
                         .getUTF16Ref(tmp);
+
```
mganandraj, line 38: This change was added for debugging ... And it's not intended for final checkin ..

API/hermes/CMakeLists.txt
```diff
@@ -93,8 +93,8 @@ elseif("${CMAKE_CXX_COMPILER_ID}" MATCHES "MSVC")
 endif()
 
 set_target_properties(libhermes PROPERTIES
-  COMPILE_FLAGS "${compile_flags}"
-  LINK_FLAGS "${link_flags}"
+  COMPILE_FLAGS "${compile_flags} -Wl,--hash-style=both"
```
mganandraj, line 96: This was needed to get the libs loaded on older platforms .. but it's not specific to intl code.. Hence, there is something i'm missing and it is probably not the best ways to do it ..

CMakeLists.txt
```diff
@@ -163,7 +163,7 @@ set(HERMES_STATIC_LINK OFF CACHE BOOL
 set(HERMES_USE_STATIC_ICU OFF CACHE BOOL
   "Force static linking of ICU. May only work on GNU/Linux.")
 
-set(HERMES_ENABLE_DEBUGGER_DEFAULT OFF)
+set(HERMES_ENABLE_DEBUGGER_DEFAULT ON)
```
mganandraj, line 166: I'll revert this change ..

android/.gitignore
```diff
@@ -1,2 +1,90 @@
 .gradle
 .externalNativeBuild
+
+# https://github.com/github/gitignore/blob/master/Android.gitignore
```
mganandraj, line 4: This is not intented for checking .. Ill revert it ..

android/build.gradle
```diff
@@ -40,14 +40,15 @@ buildscript {
     compileSdkVersion = 28
     abis = project.hasProperty('abis') ?
          project.getProperty("abis").split(",") :
-         ["arm64-v8a", "armeabi-v7a", "x86_64", "x86"]
+         ["x86"]
```
mganandraj, line 43: TO be reverted ..

lib/VM/JSLib/Intl.cpp
mganandraj: Thanks!

lib/Platform/Intl/java/com/facebook/hermes/intl/UnicodeExtensionKeys.java
```diff
@@ -0,0 +1,177 @@
...
+
+    public static String COLLATION_CASEFIRST = "colcasefirst"; // TODO:: double check this
+    public static String COLLATION_CASEFIRST_CANON = "kf";
+
+    // TODO :: Build/Look-for a BiMap datastructure.
+    private static HashMap<String, String> s_canonicalkey_icukey_map;
+    private static HashMap<String, String> s_icukey_canonicalkey_map;
+
+    static {
+        s_canonicalkey_icukey_map = new HashMap<String, String>() {{
```
mganandraj, line 36: TODO:: These maps should be part of the Intl service instance .. not global .. so that their cost is per instance .. and garbage collectable ..

mhorowitz:
@mganandraj I'm working through the code, but I also tried running the tests.  Most pass (great!) but I still see some failures: https://gist.github.com/mhorowitz/2eb4758d066dc3ae2a3e84765f01b992

are these expected?

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterICU.java
```diff
@@ -0,0 +1,330 @@
...
+        }
+
+        return this;
+    }
+
+    @Override
+    public PlatformNumberFormatterICU setSignDisplay(IPlatformNumberFormatter.SignDisplay signDisplay) {
+        if (mNumberFormat instanceof android.icu.text.DecimalFormat) {
+            android.icu.text.DecimalFormat decimalFormat = (android.icu.text.DecimalFormat)mNumberFormat;
+            android.icu.text.DecimalFormatSymbols symbols = decimalFormat.getDecimalFormatSymbols();
```
mhorowitz, line 126: https://developer.android.com/reference/java/text/DecimalFormat#getDecimalFormatSymbols() says this returns a copy.  Do we have tests that this works correctly?

lib/Platform/Intl/java/com/facebook/hermes/intl/notes.md
```diff
@@ -0,0 +1,9 @@
+Known Issues
+
+1. "intl402/Collator/ignore-invalid-unicode-ext-values.js" is failing because Array.sort is not stable.
+
+To reproduce, repeat this statement multiple times : 
+
+console.log(testArray.sort(new Intl.Collator().compare));
+
+2. Spec requires the object returned by Intl.Collator.prototype.resolvedOptions() to be in specific order. Because we use JavaHashMaps in Java code and std::unordered_map in the C++ layer, we can't produce deterministic ordering of keys. This fails "intl402/Collator/prototype/order.js"
```
mhorowitz, line 9: Nice catch, I hadn't realized there were were ordering constraints on the resolved options.

I see two potential fixes here.
1. we could make C++ optionsToJs take an extra argument giving the options order, and then it can add them to the JS object in that order.
2. we could LinkedHashMap in Java for resolved options, which preserves insertion order, and in the C++ side, introduce a new OrderedOptions = std::vector<std::pair<std::u16string, Option>> which is used for returning resolved options.

I think 2 is a bit nicer, as it doesn't require maintaining a separate ordering list.

mhorowitz (commented):
I've made a few more comments, but this is looking better.  There are some aesthetic and subjective things that we could do differently, but I don't think it makes sense to block landing on doing them.

However, there are still a bunch of test failures, both as you have excluded, and some additional ones I linked above.  I haven't looked very closely at them to understand how significant they are.  How concerned are you by the failures that are there?  Would you prefer to land this PR, and then fix the failures later, or fix them before landing?

lib/Platform/Intl/java/com/facebook/hermes/intl/notes.md
mganandraj: I second that #2 is nicer .. In the latest iteration I've fixed the resolvedOptions function of all services to create LinkedHashMap .. the c++ side change is not made though.

lib/Platform/Intl/PlatformIntlAndroid.cpp
```diff
@@ -93,9 +93,8 @@ jni::local_ref<JOptionsMap> optionsToJava(const Options &options) {
     } else if (kv.second.isNumber()) {
       jvalue = jni::autobox(static_cast<jdouble>(kv.second.getNumber()));
     } else {
-      assert(kv.second.isString() && "Option is not valid type");
+      assert(kv.second.isString());
```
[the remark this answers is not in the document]
mganandraj: Ah .. Somehow I keep reintroducing this !

lib/Platform/Intl/java/com/facebook/hermes/intl/OptionHelpers.java
```diff
@@ -0,0 +1,98 @@
...
+                throw new JSRangeErrorException("String option expected but not found");
+        }
+
+        return value;
+    }
+
+    public static <T extends Enum<?>> T searchEnum(Class<T> enumeration, Object value) {
+
+        if(JSObjects.isUndefined(value)) {
+            // TODO :: there must be a better way to do this.
```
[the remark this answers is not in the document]
mganandraj: Bingo.. Thanks!

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformNumberFormatterICU.java
mganandraj: We are modifying the currency symbol in the copy and setting it back to DecimalFormat object calling its "setDecimalFormatSymbols" .. SignDisplay is one attribute which try to do our best with the APIs that we got, but still  we do very poorly .. We've it explicitly mentioned in the tests in HermesIntlNumberFormatTest.java (https://github.com/facebook/hermes/pull/338/files#diff-334df24cf4d5b7537b266fd3c4c4151f)

We've better API support available post from API level 30 for SignDisplay : https://developer.android.com/reference/android/icu/number/NumberFormatter.SignDisplay which we can look into later ..

mganandraj:
> I've made a few more comments, but this is looking better. There are some aesthetic and subjective things that we could do differently, but I don't think it makes sense to block landing on doing them.
> 
> However, there are still a bunch of test failures, both as you have excluded, and some additional ones I linked above. I haven't looked very closely at them to understand how significant they are. How concerned are you by the failures that are there? Would you prefer to land this PR, and then fix the failures later, or fix them before landing?

Ah .. I wasn't seeing these failures likely because my VM was based on API30 .. But i could reproduce these on a new VM with API26 .. THese are due to wierd exceptions on exotic locales .. I've now patched to code to fallback to default locale or in the worst case to "en" .. and all the (whitelisted) tests are passing on API 26 and API 24 ... 

In general, I believe we do fairly well on API 24 and forward ... I'll try to list down the deficiencies that are on top of my mind, but i've to sit down and create an exhausitve list which i'll do very soon, but i haven't got a chance to do yet .. Sorry .. But i've invested a few hours in categorizing test failures in code .. Please take a look at HermesIntlCollatorTest.java, HermesIntlNumberFormatTest.java etc. please.

1. NumberFormatting has the following issues 
   a. In handling "signDisplay" attribute
   b. formatToParts don't return parts with "Unit" style
2. The keys in resolvedOptions are not ordered.
3. There's a few deviations from what test262 expects, because we rely on what ICU provides. For e.g. GMT vs UTC. which we don't try to explicitly match with the test dataset.
4. I'm not sure how people use "Intl.getCanonicalLocale" method. If it's supposed to be a stable identifier across versions, so that people can store it in database, we don't return stable values across versions, because ICU doesn't provide us stable values across. Our custom implementation prior to v24 does return stable values, but the values may differ from ICU's locale id in special cases.

The above is by no means an exhaustive list. I'll try to improve this situation in the coming days.

But I believe our current implementation will do well for majority of the user scenarios and locales around the world.

android/intltest/java/com/facebook/hermes/test/HermesIntlDateFormatTest.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    public void testIntlDateTimeFormat_prototype_toStringTag() throws IOException {
+
+        String basePath = "test262-main/test/intl402/DateTimeFormat/prototype/toStringTag";
+
+        Set<String> whilteList = new HashSet<>();
+
+        Set<String> deviations = new HashSet<>(Arrays.asList(
+                "toString.js", //Expected SameValue(«[object Object]», «[object Intl.DateTimeFormat]») to be true
```
mhorowitz, line 195: This was a due to a spec change: https://tc39.es/ecma402/#annex-incompatibilities

android/intltest/java/com/facebook/hermes/test/HermesIntlGetCanonicalLocalesTest.java
```diff
@@ -0,0 +1,94 @@
...
+    private static final String LOG_TAG = "HermesIntlGetCanonicalLocalesTest";
+
+    @Test
+    public void testIntlGetCanonicalLocales() throws IOException {
+        String basePath = "test262-main/test/intl402/Intl/getCanonicalLocales";
+        Set<String> whiteList = new HashSet<>();
+        Set<String> blackList = new HashSet<>(Arrays.asList("Locale-object.js"
+                , "canonicalized-tags.js" // All except one tag (cmn-hans-cn-u-ca-t-ca-x-t-u) passes. icu4j adds an extra 'yes' token to the unicode 'ca' extension!
+                , "complex-region-subtag-replacement.js" // We don't do complex region replacement.
+                , "has-property.js" // Test needs Proxy,
```
mhorowitz, line 45: Proxy is now enabled by default

android/intltest/java/com/facebook/hermes/test/HermesIntlGetCanonicalLocalesTest.java
```diff
@@ -0,0 +1,94 @@
...
+public class HermesIntlGetCanonicalLocalesTest extends HermesIntlTest262Base {
+
+    private static final String LOG_TAG = "HermesIntlGetCanonicalLocalesTest";
+
+    @Test
+    public void testIntlGetCanonicalLocales() throws IOException {
+        String basePath = "test262-main/test/intl402/Intl/getCanonicalLocales";
+        Set<String> whiteList = new HashSet<>();
+        Set<String> blackList = new HashSet<>(Arrays.asList("Locale-object.js"
+                , "canonicalized-tags.js" // All except one tag (cmn-hans-cn-u-ca-t-ca-x-t-u) passes. icu4j adds an extra 'yes' token to the unicode 'ca' extension!
```
mhorowitz, line 43: Since this affects more than one test, is this added 'yes' something we can easily work around?

android/intltest/java/com/facebook/hermes/test/HermesIntlGetCanonicalLocalesTest.java
```diff
@@ -0,0 +1,94 @@
...
+                , "complex-region-subtag-replacement.js" // We don't do complex region replacement.
+                , "has-property.js" // Test needs Proxy,
+                , "non-iana-canon.js" // All except one tag (de-u-kf) passes. icu4j adds an extra 'yes' token to the unicode 'kf' extension !
+                , "preferred-variant.js" // We don;t do variant replacement
+                , "transformed-ext-canonical.js" // We don't canonicalize extensions yet.
```
mhorowitz, line 48: Do we plan to?

android/intltest/java/com/facebook/hermes/test/HermesIntlGetCanonicalLocalesTest.java
```diff
@@ -0,0 +1,94 @@
...
+                , "transformed-ext-invalid.js"  // We don't canonicalize extensions yet.
+                , "unicode-ext-canonicalize-region.js"  // We don't canonicalize extensions yet.
+                , "unicode-ext-canonicalize-subdivision.js"  // We don't canonicalize extensions yet.
+                , "unicode-ext-canonicalize-yes-to-true.js"  // We don't canonicalize extensions yet.
+                , "unicode-ext-key-with-digit.js"  // We don't canonicalize extensions yet.
+                , "grandfathered.js" // icu4j doesn't perform all grandfathered tag replacements.
+                , "preferred-grandfathered.js" // icu4j doesn't perform all grandfathered tag replacements.
+                , "invalid-tags.js" // // icu4j !
```
mhorowitz, line 56: Can you elaborate?

android/intltest/java/com/facebook/hermes/test/HermesIntlNumberFormatTest.java
```diff
@@ -0,0 +1,218 @@
...
+                "constructor-order.js",  // When strictly following spec, the currency checks comes before unit check and the test will throw RangeError on currency validation before reaching the cdoe which throws TypeError on seeing undefined unit. But, we have a part of option validation in C++ code which throws all TypeErrors, which results in the TypeError getting thrown.
+                "currency-digits.js", // Didn't get correct minimumFractionDigts for currency AFN. Expected SameValue(«0», «2») to be true
+                "constructor-unit.js" // com.facebook.hermes.intl.JSRangeErrorException: Unknown unit: acre-per-acre .. We support only units directly known to https://developer.android.com/reference/android/icu/util/MeasureUnit .. MeasureFormat.format requires an instance of MeasureUnit.
+        ));
+
+        Set<String> testIssues = new HashSet<>(Arrays.asList(
+                "proto-from-ctor-realm.js", // Property '$262' doesn't exist
```
mhorowitz, line 52: defined in harness/detachArrayBuffer-host-detachArrayBuffer.js

android/intltest/java/com/facebook/hermes/test/HermesIntlNumberFormatTest.java
```diff
@@ -0,0 +1,218 @@
...
+                "constructor-locales-get-tostring.js", // Property 'Proxy' doesn't exist
+                "subclassing.js", // Compiling JS failed: 18:1:invalid statement encountered. Buffer size 986 starts with: 2f2f20436f7079726967687420323031
```
mhorowitz, line 54: this is because Hermes doesn't support classes yet.

mhorowitz (commented):
> But i've invested a few hours in categorizing test failures in code .. Please take a look at HermesIntlCollatorTest.java, HermesIntlNumberFormatTest.java etc. please.

Sure, this is useful.  I've made a bunch of comments inline.  it seems quite a few failure would be fixed by rebasing, or understanding why some harness functions aren't working.

I agree that a list of issues which seem top of mind to you would be helpful.  In addition to this, it would be helpful to document what failures are implementation bugs we plan to fix later, what is due to limitations or bugs in android.icu, what is due to limitations in Hermes, what is due to test262 bugs, etc.

Once we have such a list, we can decide what we believe needs to be fixed before landing this PR, and what can wait for later.

I think more generally, we also need a developer-facing doc which describes which features aren't expected to work correctly.  Most JavaScript developers would have a hard time figuring this out from the C++ and Java code.  This could be a followup.

> I'm not sure how people use "Intl.getCanonicalLocale" method. If it's supposed to be a stable identifier across versions, so that people can store it in database, we don't return stable values across versions, because ICU doesn't provide us stable values across. Our custom implementation prior to v24 does return stable values, but the values may differ from ICU's locale id in special cases.

I'm not sure, either.  In https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers, referenced by https://tc39.es/ecma402/#sec-canonicalizeunicodelocaleid, canonicalization seems to be mostly about order and case of various tokens in a locale identifier.  Can you give an example of returns which aren't stable across versions?  That would help me understand what we're dealing with.

mganandraj:
Hi @mhorowitz .. I think the PR is in a good shape for the initial commit .. Could you please let me know if you have any strong concerns ? If not, could you please signoff/merge the changes ?

android/intltest/build.gradle
```diff
@@ -8,17 +8,53 @@
...
+  id("de.undercouch.download")
+}
+
+import de.undercouch.gradle.tasks.download.Download
 
 buildDir = "${rootProject.ext.hermes_ws}/build_android/intltest"
 buildDir.mkdirs()
 
+def facebookBuild = System.getenv("FACEBOOK") ?: "0"
+def hermesHostBuild = System.getenv("HERMES_WS_DIR")
```
mganandraj, line 22: These lines should be removed prior to checkin.

android/intltest/java/com/facebook/hermes/test/HermesIntlDateFormatTest.java
```diff
@@ -0,0 +1,214 @@
...
+        String basePath = "test262-main/test/intl402/DateTimeFormat/prototype/format";
+
+        Set<String> whiteList = new HashSet<>();
+
+        Set<String> deviations = new HashSet<>(Arrays.asList(
+                "proleptic-gregorian-calendar.js", // Strangely icu4j returned date for -8640000000000000 (beginning of ECMAScript time) is November 271817 .. The expected is April 271822 .. Need to dig in.
+                "time-clip-near-time-boundaries.js" //Expected a RangeError but got a ReferenceError
+        ));
+
+        Set<String> unSupportedList = new HashSet<>(Arrays.asList(
```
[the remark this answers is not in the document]
mganandraj: I've added better comments around the rationale behind not supporting them. I believe, you made the same assumption when writing the C++ Intl code, which i presumed based on the set of attributes that you'd included in the code. 
I'm happy to continue contributing in future towards implementing the newer editions of the standard. Of course if you are happy with the quality of my contribution so far :)

android/intltest/java/com/facebook/hermes/test/HermesIntlGetCanonicalLocalesTest.java
mganandraj: This has helped in trimming down our blacklists .. But, i had to add a small snippet to Intl.cpp to allow for holes in the proxy based array inputs .. Please take a look at it to see whether it's correct.

And, it looks like the "has" trap is not getting called when accessin the proxy (when used as locales arguments),as required by the test262 test.

android/intltest/java/com/facebook/hermes/test/HermesIntlNumberFormatTest.java
mganandraj: That one doesn't seem to be the one.. 

But https://github.com/tc39/test262/blob/20e31011058783d9bdeb4c34a8aacbb632355e2e/implementation-contributed/v8/test262/harness-adapt.js#L107 looks more like the one, which adds $262 to global with RealmOperators .. 

Do we have something like it for Hermes?

android/intltest/java/com/facebook/hermes/test/HermesIntlGetCanonicalLocalesTest.java
mganandraj: Done ..

android/intltest/java/com/facebook/hermes/test/HermesIntlGetCanonicalLocalesTest.java
mganandraj: Currently we don't plan to .. Im not sure whether how important it is to stick to test data in these scenarios .. More stricture canoicalization is mroe compuration and space intensive .. which doesn't worth unless it's critical ..

android/intltest/java/com/facebook/hermes/test/HermesIntlNumberFormatTest.java
mhorowitz: It looks like we do, it's here: https://github.com/facebook/hermes/blob/ad7229435fdd6ab45beff44731a955067c1c072a/utils/testsuite/testsuite.py#L130-L137

Unfortunately, it's inlined in the py file right now.  We should be able to pull it into a standalone js file you can use.

mganandraj: Let me try it today.

android/intltest/java/com/facebook/hermes/test/HermesIntlDateFormatTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+    public void testIntlDateTimeFormat_prototype_format() throws IOException {
+
+        String basePath = "test262-main/test/intl402/DateTimeFormat/prototype/format";
+
+        Set<String> whiteList = new HashSet<>();
+
+        Set<String> deviations = new HashSet<>(Arrays.asList(
+                "proleptic-gregorian-calendar.js", // Strangely icu4j returned date for -8640000000000000 (beginning of ECMAScript time) is November 271817 .. The expected is April 271822 .. Need to dig in.
+                "time-clip-near-time-boundaries.js" //Expected a RangeError but got a ReferenceError
```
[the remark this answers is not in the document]
mganandraj: It looked like a bug in the ICU .. The same could be reproduced in the Java debugger too .. I need to dig in ... I'll make a note of it in the list of items to follow up .. Thanks!

mhorowitz (approved):
The Notes file is great, thank you!  Can we flesh out item 3 under Future Work?  It would be helpful to know what we expect to fix, and what we can't, for whatever reason.  Eventually, it would be nice to have this in a format more easily usable from GitHub.  Markdown is ideal, but I don't know how practical it is.  I found some converters and plugins for Word, but your document is pretty complicated, so I don't know how well they will work.  If we can export to an .html file and include that in our docs somehow, that would work, too.

It looks like there's a bunch of unintentional whitespace changes.  I'll be running a local linter over the code anyway, so that should clean it up without you needing to do anything.

That said, please don't change any of this in this PR.  This is passing internal tests and ready to land, but I need to restart the process when you update the PR, so any further changes should be in a new PR on a branch after this one lands.

lib/VM/JSLib/Intl.cpp
```diff
@@ -202,14 +209,24 @@ CallResult<std::vector<std::u16string>> normalizeLocales(
...
           createListFromArrayLike(
               localeObj,
               runtime,
               *lengthRes,
-              [&ret](Runtime *runtime, uint64_t index, PseudoHandle<> value) {
-                CallResult<std::u16string> strRes =
-                    stringFromJS(runtime, std::move(value));
+              [&ret, isProxy](Runtime *runtime, uint64_t index, PseudoHandle<> value) {
+
+    // When the locales list is a proxy object, gaps are allowed.
```
mhorowitz, line 220: As far as I can tell, gaps are allowed for non-proxies, too.

mganandraj: If I remember correctly, firefox throws range exception with normal arrays with gaps .. I couldn't get to the spec which covers it.

mganandraj:
> The Notes file is great, thank you! Can we flesh out item 3 under Future Work? It would be helpful to know what we expect to fix, and what we can't, for whatever reason. Eventually, it would be nice to have this in a format more easily usable from GitHub. Markdown is ideal, but I don't know how practical it is. I found some converters and plugins for Word, but your document is pretty complicated, so I don't know how well they will work. If we can export to an .html file and include that in our docs somehow, that would work, too.
> 
> It looks like there's a bunch of unintentional whitespace changes. I'll be running a local linter over the code anyway, so that should clean it up without you needing to do anything.
> 
> That said, please don't change any of this in this PR. This is passing internal tests and ready to land, but I need to restart the process when you update the PR, so any further changes should be in a new PR on a branch after this one lands.

Thanks ! I'll flesh out the details and convert the doc to a more friendly format in the coming days soon .. ( I think the whitespace changes were introduced by some last minute editing in VsCode .. Thanks for taking care of them ! )

lib/VM/JSLib/Intl.cpp
mhorowitz: I tried
`new Intl.Collator([, "en-US", , "fr-FR"]).resolvedOptions().locale`
on node, firefox, and jsc, and all returned "en-US" without throwing.  I believe the relevant spec is https://tc39.es/ecma402/#sec-canonicalizelocalelist step 7.b-c, which skips elements not present.

mganandraj: new Intl.Collator([, "en-US", undefined , "fr-FR"]).resolvedOptions().locale is throwing a TypeError with chrome : { Language ID should be string or object. } .. but new Intl.Collator([, "en-US", , "fr-FR"]).resolvedOptions().locale is not .. 

From a developer perspective, both "['a',,'c'][1]" and "['a',undefined,'c'][1]" evaluates to undefined .. But in the locales array, a gap and "explicit undefined" doesn't mean the same !

mganandraj:
> > The Notes file is great, thank you! Can we flesh out item 3 under Future Work? It would be helpful to know what we expect to fix, and what we can't, for whatever reason. Eventually, it would be nice to have this in a format more easily usable from GitHub. Markdown is ideal, but I don't know how practical it is. I found some converters and plugins for Word, but your document is pretty complicated, so I don't know how well they will work. If we can export to an .html file and include that in our docs somehow, that would work, too.
> > It looks like there's a bunch of unintentional whitespace changes. I'll be running a local linter over the code anyway, so that should clean it up without you needing to do anything.
> > That said, please don't change any of this in this PR. This is passing internal tests and ready to land, but I need to restart the process when you update the PR, so any further changes should be in a new PR on a branch after this one lands.
> 
> Thanks ! I'll flesh out the details and convert the doc to a more friendly format in the coming days soon .. ( I think the whitespace changes were introduced by some last minute editing in VsCode .. Thanks for taking care of them ! )

@mhorowitz Do I need to do anything to resolve the Facebook internal linter/build/test errors ?

lib/VM/JSLib/Intl.cpp
mhorowitz: There's a difference between undefined and having a gap, though.  They do behave differently as described in the spec, but I don't see any expected difference in behavior for a Proxy vs a plain array.  In your example, `a[1]` are both undefined, but `Reflect.has(a, 1)` are not the same.

mhorowitz:
> @mhorowitz Do I need to do anything to resolve the Facebook internal linter/build/test errors ?

No, everything is ok.   For some reason, the bot is just not updating that on GitHub.

PR #345, opened 2020-09-09
Add arm64 slice to macOS framework

alloy:
## Summary

This adds a arm64 slice to the macOS framework artefact for the upcoming ARM based Macs.

## Test Plan

As tested on a DTK machine:

Screenshot 2020-09-14 at 17 45 06

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,7 +41,7 @@ workflows:
 # Default settings for Apple jobs (apple-runtime, test-apple-runtime)
 apple_defaults: &apple_defaults
   macos:
-    xcode: "10.3.0"
+    xcode: "12.0.0"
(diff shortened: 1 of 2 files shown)


tmikov:
OK, let us know when.

alloy:
@tmikov Ok, this is good to go 👍 

I’ll have to backport this to 0.4.x for RN macOS v0.62.x (which is our current stable release) and 0.5.x (which will soon be needed for when we release RN macOS v0.63.x).

alloy:
@tmikov Hmm, working through the v0.4.x backport, I just realized that the CLI tools are built in a different job that does not make use of the same build artefacts and thus doesn’t build for arm64. I can make a separate PR for that or amend this one, whichever makes the most sense for you.

PR #352, opened 2020-09-15
[circle] Build macOS CLI artefacts for arm64

alloy:
## Summary

As noted on https://github.com/facebook/hermes/pull/345#issuecomment-692228581, that PR did not update the CLI artefacts to build for arm64.

I think there’s an opportunity to reduce the amount of separate builds and package both the runtime package and CLI from the same artefacts while also reducing build time, but for now I’d like to ensure the CLI tools at least also come with arm64 slices.

## Test Plan

Tested on a DTK machine:

```
$ ditto --arch arm64 package/osx-bin/hermes package/osx-bin/hermes-arm64                 
$ file package/osx-bin/hermes-arm64                                                      
package/osx-bin/hermes-arm64: Mach-O 64-bit executable arm64
$ echo "'use strict'; function hello() { print('Hello World'); } hello();" | ./package/osx-bin/hermes-arm64 
>> Hello World
undefined
>> 
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -196,7 +196,7 @@ jobs:
 
   macos:
     macos:
-      xcode: "10.0.0"
+      xcode: "12.0.0"
     environment:
       - HERMES_WS_DIR: /tmp/hermes
       - TERM: dumb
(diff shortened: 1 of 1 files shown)


alloy:
@tmikov @Huxpro It would be good to include this before pushing out a v0.7.0 npm package.

.circleci/config.yml
```diff
@@ -196,7 +196,7 @@ jobs:
 
   macos:
     macos:
-      xcode: "10.0.0"
+      xcode: "12.0.0"
```
tmikov, line 199: Do we need this at all, since it is set as default above (same for some other settings like TERM)?

alloy: Definitely seems like we should be able to leverage that indeed. Before I forget this and the issue mentioned in the OP, I filed a ticket https://github.com/facebook/hermes/issues/353

PR #363, opened 2020-09-21
feat: build missing architectures on iOS

grabbou:
Turns out we need one more architecture for certain simulators (required by Xcode) to ship with the framework.


alloy:
Ah yes, old devices. This arch is only needed for iOS 10, right?

alloy:
> @alloy my understanding is limited here, but from my research, this seems to be mainly used on 32 bit simulator devices on iOS 10.

According to [this wikipedia entry](https://en.wikipedia.org/wiki/IOS_10#Supported_devices) iOS 10 is indeed the last OS that supports 32-bit devices, so until RN drops iOS 10 support we need to include this.

PR #364, opened 2020-09-21
fix: broken Hermes.podspec post-install in release

grabbou:
Utilities are not part of a final package. We need to check for their presence while Hermes is being installed, otherwise, this will throw an error and make installation impossible.


alloy:
Makes sense 👍

PR #365, opened 2020-09-21
feat: enable bitcode

grabbou:
Bitcode is turned on by default in React Native and so, setting it here as well.


tmikov:
There needs to be an option HERMES_ENABLE_BITCODE in the main CMake file, no?

Huxpro:
@grabbou could you define the new option in the main file, similar to this: https://github.com/facebook/hermes/blob/070ffddf819831772e0f53445c952774f32db92b/CMakeLists.txt#L146 ?

grabbou:
@Huxpro will do, I will also add some other missing flags that we have added the other day in a separate PR.

grabbou:
@tmikov just saw your comment. Didn't need to set it, so not sure whether it "has" to be there by the means of working / not working. Would appreciate a bit of clarification so that I can deepen my understanding of CMake! 

I have tested this on my RNTester application (part of React Native PR) and it worked with bitcode enabled.

tmikov:
@grabbou looks good now. The main reason to do it is that we want all flags controlling the build to be listed and documented in one place.

Issue #368, opened 2020-09-22
Optimise the size of iOS artifacts


grabbou:
## Problem

Right now, the artifact itself is rather large after all architectures, dSYMs, and bitcode were included - **2GB**. We need to find a way to reduce this to a reasonably small size, so that it doesn't affect React Native users.

The goal is to reduce the size so that it comes w/o penalty for new React Native applications.

## Solution

- Optimise size of `hermes-engine-darwin` npm package
  - Turn off bitcode (**2GB** with bitcode enabled for three architectures, **12MB** without bitcode). While it saves the size, it makes it impossible to build React Native with `bitcode`, which comes "on" by default. Potential change in the default template possible as long as it makes sense to do so. We can also disable `bitcode` for `iphonesimulator` architectures as I think it makes no sense to ship that for non-release architectures (won't be able to build for simulator in the release tho)
  - Decide what to do with symbols - `dSYM` weights **200mb**
- Make it optional package, so that users will have to install `hermes-engine-darwin` npm package on demand (unlike for Android) - size constraint doesn't affect the initial setup time. Potential degraded DX as it requires many actions to be done to enable Hermes
- Do not use `npm` for `hermes-engine-darwin`, but use `CocoaPods` dependency management mechanism. We can publish full Hermes to CocoaPods repository and it will be automatically installed as soon as React Native users change the flag in a Podfile to `enable_hermes`. This is done by removing `path` to a `Hermes` podspec that points to `node_modules`.
- Build Hermes in the user land (don't ship artifacts)

Opening this issue to discuss potential ways going forward and options that we have.

alloy:
We definitely do not need bitcode for the simulator. IIRC, Xcode doesn’t even actually build bitcode until an archive release is built.

Symbols should be made available, at the very least for release builds in order to symbolicate crash reports. These could potentially be made available separately for when needed, but we have to take into account the possibility that RN users will not know/remember to do so and in turn struggle with unhelpful crash reports.

I am in favour of not using npm. At first it seemed like it might be good to stay consistent for now, but it’s becoming clear that it’s only adding friction, with eg mow needing to decide when to install it.

(FWIW I believe we also shouldn’t install the Android runtime by default.)

Building hermes in user-land would solve the artefact size issue, but isn’t trivial to convert (although everything is doable) and would increase the build time of RN projects.

grabbou:
> Are you saying Xcode won’t allow a simulator release build without the slice present?

[...]

PR #354, opened 2020-09-16
Do not build cli tools for iphone targets

alloy:
## Summary

Currently (in `v0.7.0`) we build the CLI tools for both `iphone` and `macos` targets, which means that, when installing into the same location, cmake will skip installing the CLI tools from the second build. In our current situation, the second installation is the macOS build, which is really the only target from which we want CLI tools.

```
$ tar -zxvf hermes-engine-darwin-0.7.0.tgz
$ file package/destroot/bin/hermesc
package/destroot/bin/hermesc: Mach-O universal binary with 3 architectures: [arm_v7:Mach-O executable arm_v7] [arm_v7s:Mach-O executable arm_v7s] [arm64]
package/destroot/bin/hermesc (for architecture armv7):    Mach-O executable arm_v7
package/destroot/bin/hermesc (for architecture armv7s):    Mach-O executable arm_v7s
package/destroot/bin/hermesc (for architecture arm64):    Mach-O 64-bit executable arm64
```

Put differently, nobody runs `hermesc` on iOS, we always only want the macOS builds of these.

----

I opted to make building CLI tools a simple cmake option and then use that from the apple builds to control wether or not to include them.

## Test Plan

With the `hermes-engine-darwin-0.7.0.tgz` artefact of this `npm` CI job:

```
$ tar -zxvf hermes-engine-darwin-0.7.0.tgz
$ file package/destroot/bin/hermesc
package/destroot/bin/hermesc: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64]
package/destroot/bin/hermesc (for architecture x86_64):	Mach-O 64-bit executable x86_64
package/destroot/bin/hermesc (for architecture arm64):	Mach-O 64-bit executable arm64
```

PR #376, opened 2020-09-30
feat: only do bitcode for iphone

grabbou:
## Summary

Bitcode is not needed for `iphonesimulator` - this PR makes bitcode to be included with iOS only. Reduces the package size by ~1GB.

Related to #368


Huxpro:
@grabbou looks like there's a merge conflict after #354 getting landed. Could you rebase over it?

grabbou:
Done 👌

PR #390, opened 2020-10-11
Add macOS arm64 support to 0.5.x

alloy:
## Summary

Build macOS artefacts for macOS 10.13 and with arm64 slices.

## Test Plan

Screenshot 2020-10-22 at 19 27 07

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,10 +63,10 @@ endif()
 # - npm/package.json
 # - hermes.podspec
 project(Hermes
-        VERSION 0.5.2
+        VERSION 0.5.3
         LANGUAGES C CXX)
 # Optional suffix like "-rc3"
-set(VERSION_SUFFIX "-rc1")
+set(VERSION_SUFFIX "")
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules/")
 
--- a/hermes.podspec
+++ b/hermes.podspec
@@ -7,12 +7,14 @@ module HermesHelper
   # BUILD_TYPE = :debug
   BUILD_TYPE = :release
 
+  DEPLOYMENT_TARGET = "10.13"
+
   def self.command_exists?(bin)
     "command -v #{bin} > /dev/null 2>&1"
   end
 
   def self.cmake_configuration
-    "-DHERMES_ENABLE_DEBUGGER:BOOLEAN=true -DHERMES_ENABLE_FUZZING:BOOLEAN=false -DHERMES_ENABLE_TEST_SUITE:BOOLEAN=false -DHERMES_BUILD_APPLE_FRAMEWORK:BOOLEAN=true -DHERMES_BUILD_APPLE_DSYM:BOOLEAN=true"
+    "-DHERMES_ENABLE_DEBUGGER:BOOLEAN=true -DHERMES_ENABLE_FUZZING:BOOLEAN=false -DHERMES_ENABLE_TEST_SUITE:BOOLEAN=false -DHERMES_BUILD_APPLE_FRAMEWORK:BOOLEAN=true -DHERMES_BUILD_APPLE_DSYM:BOOLEAN=true -DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=#{DEPLOYMENT_TARGET} -DCMAKE_OSX_ARCHITECTURES:STRING=x86_64;arm64"
   end
 
   def self.configure_command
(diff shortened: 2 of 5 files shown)


hermes.podspec
```diff
@@ -22,14 +24,14 @@ end
 
 Pod::Spec.new do |spec|
   spec.name        = "hermes"
-  spec.version     = "0.5.1"
+  spec.version     = "0.5.2-rc.1"
```
alloy, line 27: I changed this to be inline with the current version in this branch, but this should likely change before the actual release?

alloy:
Ok, build is green. Will test on a DTK later today.

alloy:
@Huxpro This is tested and good to go. The one thing that would need a change is the version, I guess it can just be v0.5.2 or v0.5.3?

Huxpro:
@alloy hey i'm coming back to this!

> The one thing that would need a change is the version, I guess it can just be v0.5.2 or v0.5.3?

Hm i'm not sure how does the arm64 support work to tell this. What does adding `arm64` to the [CMAKE_OSX_ARCHITECTURES](https://cmake.org/cmake/help/v3.14/variable/CMAKE_OSX_ARCHITECTURES.html#variable:CMAKE_OSX_ARCHITECTURES) do? It seems like there is still only one `destroot/Library/Frameworks/hermes.framework/Versions/0/hermes` in the `hermes-runtime-darwin` tarball built from the CircleCI of this PR and it's still an x86_64 executable? Well actually, I realized I don't know how is the `hermes-runtime-darwin` thing being used at all...apparently it's different with the `libhermes.so` used by Android.

Basically, we used `rc1` in the `0.5.2-rc1` for Proxy because it's an opt-in npm package instead of being a default. `npm install hermes-engine@~0.5.0` can pick `0.5.2` but not `0.5.2-rc1`. 

So if the npm package after this PR can work for both x86 and ARM out of the box I will say `0.5.3`. If it's not going to work with x86 I will say `0.5.3-blah` (`rc1`, or `mac-arm`, idk) and let RN macOS ARM somehow opt in the `hermes-engine@0.5.3-blah`.

alloy:
> What does adding `arm64` to the [CMAKE_OSX_ARCHITECTURES](https://cmake.org/cmake/help/v3.14/variable/CMAKE_OSX_ARCHITECTURES.html#variable:CMAKE_OSX_ARCHITECTURES) do? It seems like there is still only one `destroot/Library/Frameworks/hermes.framework/Versions/0/hermes` in the `hermes-runtime-darwin` tarball built from the CircleCI of this PR, and it is still an x86_64 executable?

Mach-O binaries can be [“fat” binaries](https://en.wikipedia.org/wiki/Fat_binary), which mean they can contain multiple slices to support different architectures.

In the case of these artefacts:

```bash
$ file hermes-runtime-darwin-v0.5.2-rc1/destroot/bin/hermesc 
hermes-runtime-darwin-v0.5.2-rc1/destroot/bin/hermesc: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64]
hermes-runtime-darwin-v0.5.2-rc1/destroot/bin/hermesc (for architecture x86_64):	Mach-O 64-bit executable x86_64
hermes-runtime-darwin-v0.5.2-rc1/destroot/bin/hermesc (for architecture arm64):	Mach-O 64-bit executable arm64
$ file hermes-runtime-darwin-v0.5.2-rc1/destroot/Library/Frameworks/hermes.framework/Versions/0/hermes 
hermes-runtime-darwin-v0.5.2-rc1/destroot/Library/Frameworks/hermes.framework/Versions/0/hermes: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit dynamically linked shared library x86_64] [arm64]
hermes-runtime-darwin-v0.5.2-rc1/destroot/Library/Frameworks/hermes.framework/Versions/0/hermes (for architecture x86_64):	Mach-O 64-bit dynamically linked shared library x86_64
hermes-runtime-darwin-v0.5.2-rc1/destroot/Library/Frameworks/hermes.framework/Versions/0/hermes (for architecture arm64):	Mach-O 64-bit dynamically linked shared library arm64
```

> So if the npm package after this PR can work for both x86 and ARM out of the box I will say `0.5.3`.

Yup, it will work for both archs 👍 

> Basically, we used `rc1` in the `0.5.2-rc1` for Proxy because it's an opt-in npm package instead of being a default. `npm install hermes-engine@~0.5.0` can pick `0.5.2` but not `0.5.2-rc1`.

Ah gotcha, good for me to at least understand the reason for that being a prerelease :+1:

hermes.podspec
Huxpro, line 27: Let's either revert this change and make another commit (like https://github.com/facebook/hermes/commit/49c2875605e190f9d8fdf66c72a30bbce5db6739) to bump all version occurrences to 0.5.3 for consistency, or I think it would be also nice to just bump all versions occurrences to 0.5.3 directly in this PR (and we can directly tag this as `v0.5.3`).

Huxpro (changes requested):
@alloy thanks for the elaboration! TIL the fat binaries!

Let's finalize this as `0.5.3` then ;)

hermes.podspec
alloy: Cool, will update 👍

alloy: Done 👍

alloy:
Updated the versions to 0.5.3.

CMakeLists.txt
```diff
@@ -63,7 +63,7 @@ endif()
 # - npm/package.json
 # - hermes.podspec
 project(Hermes
-        VERSION 0.5.2
+        VERSION 0.5.3
         LANGUAGES C CXX)
 # Optional suffix like "-rc3"
 set(VERSION_SUFFIX "-rc1")
```
Huxpro, line 69: Oops, sorry for not mentioning this `"-rc1"` need to be reset to `""` as well, otherwise artifacts would be named with `-rc1` hence the current CircleCI failures. It was set in https://github.com/facebook/hermes/commit/c9d1d8335e8f34242a26646665228d522b7df2ae.

Huxpro (changes requested):

CMakeLists.txt
alloy: Ah my bad, updated 👍

PR #392, opened 2020-10-14
doc: update yarn command

jeremybarbet:
## Summary

`yarn install` has been replaced by `yarn add`

### DIFF
--- a/doc/ReactNativeIntegration.md
+++ b/doc/ReactNativeIntegration.md
@@ -93,10 +93,10 @@ command:
 
 ```shell
 # Install the Hermes NPM to be used by the project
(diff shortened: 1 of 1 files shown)


Huxpro:
@jeremybarbet thanks for catching this!

Out of curiosity, was `yarn` okay with `yarn install <package>`? Would it be complaining that you should use `yarn add` and stoped there, or would it be automatically ran in a sort of `npm`-compatible mode?

jeremybarbet:
@Huxpro Unfortunately there was no gracious fallback, it just stopped the script and I had to run the command line again and replaced with `add`.

I'm still using v1 `yarn@1.22.5`, not sure if it's the same for the v2.

Issue #353, opened 2020-09-16
Opportunities to refactor the CI config


alloy:
## Problem

1. The config can be DRYer
2. There’s probably opportunity to reduce overall build time

## Solution

1. See https://github.com/facebook/hermes/pull/352/files#r488813262
2. The macOS CLI artefacts for the CLI package are built separately from the macOS runtime artefact, but probably don’t need to be.

alloy:
cc @grabbou

PR #393, opened 2020-10-14
Rework iOS & Mac CI setup to fix creating artefacts

grabbou:
There are a few things that I have reworked in this PR, and I have done my best to explain them in as much details as possible.

## The challenge

First, let me start with the root cause of why I started working on this PR:

> The `hermes-engine-darwin-vX.X.X.tgz` archive created by CircleCI (and later published to npm or CocoaPods) contains iPhone only architectures, and does not contain MacOS and iPhone Simulator ones.

The reason for that is simple 👇

On CircleCI, we use [`cmake` to build a special `hermes-runtime-darwin-cocoapods-release` target](https://github.com/facebook/hermes/blob/master/.circleci/config.yml#L292). In order to actually build it, we need to tell `cmake` the build folder to use.

The challenge is that we have three separate build folders - `build_iphoneos`, `build_iphonesimulator` and `build_macosx` - each, for a different platform that we support. Thanks to that, we can avoid different issues (especially between iPhone/iPhone Simulator and Mac) when building different platforms back and forth, as well as share the cache on subsequent builds. We can be also more explicit when invoking `cmake`, which helps to locate the correct build folder, e.g. [when building CLI tools](https://github.com/facebook/hermes/pull/393/commits/8c3f2fe793e6fc40b46d81fcafa2dcf6161c4e3b).

Side effect of that design decision is that **we can't just use `cmake` to build an archive with all platforms and architectures we want**. It just doesn't work (or I am not aware of a way to make it work).

## The solution

Do not build a custom `cmake` target (named "hermes-runtime-darwin-cocoapods-release"). Use plain bash to build Hermes for each platform we want and then, manually copy and package that into an archive. 

We already use that same approach while in development - there is a custom post install phase in `hermes-engine.podspec` It consists of `build_apple_framework` that builds Hermes for a given platform with flags/settings specified and `create_universal_framework` that combines multiple platforms into a single framework (e.g. iPhone and iPhone Simulator).

## Additional benefits

Together with aforementioned, I have implemented several additional improvements.

1. There is no `spec.prepare_command` in a Hermes podspec

We no longer implicitly build a framework when running `pod install` in a `test/ApplePlatformsIntegrationTestApp`. We now need to explicitly run e.g. `./utils/build-mac-framework.sh`.

As a result, I have decided to remove that step from a podspec. We no longer need to conditionally include/remove it, depending on a presence of an environmental variable.

2. Better CircleCI configuration with shared cache

Since scripts were extracted to separate files, they can be now directly run from a CircleCI. We don't need a custom `cmake` target, we can use the same code that is already used in development (and has been tested to work flawlessly).

Thanks to a few different scripts, we can have a set of CircleCI jobs, each responsible for a different aspect of building and packaging Hermes for Darwin:
- build_apple_platforms - to build MacOS, iPhone, iPhone Simulator and CLI
- test_apple_platforms - to run tests on previously built frameworks
- package_apple_platforms - to package frameworks into archives

The build action is performed only once, as a part of `build_apple_platforms` and is cached for subsequent runs and following jobs in the pipeline, to reduce the amount of time needed to test and verify results.

## How to test it

Copy `Hermes-engine-darwin-vx.x.x.tar.gz` from CircleCI `npm` job and paste it into your `Pods` folder. It will replace currently released version to CocoaPods that doesn't contain all the platforms.

Re-run on iPhone Simulator. Linking error with missing files/classes for given architecture will be gone.

--

Fixes https://github.com/facebook/hermes/issues/353 (we no longer build CLI separately).


.circleci/config.yml
```diff
@@ -269,31 +229,66 @@ jobs:
...
+            sudo gem install cocoapods
+      - run:
+          name: Build the iOS frameworks
+          command: ./utils/build-ios-framework.sh
+      - run:
+          name: Build the Mac frameworks
+          command: ./utils/build-mac-framework.sh
+      - run:
+          name: Build the CLI for Mac
+          command: cmake --build ./build_macosx --target github-cli-release
```
grabbou, line 255: This replaces entire `macos` job: https://github.com/facebook/hermes/pull/393/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47L198

.circleci/config.yml
```diff
@@ -269,31 +229,66 @@ jobs:
...
       - save_cache:
-          key: v1-repo-{{ .Environment.CIRCLE_SHA1 }}
+          key: v3-repo-{{ .Environment.CIRCLE_PR_NUMBER }}
```
grabbou, line 257: Changes to cache key are temporary to avoid rebuilding the whole framework (which takes a long time). I'd like to change this cache key to something that actually indicates whether there were changes that need to be rebuilt or not.

.circleci/config.yml
```diff
@@ -269,31 +229,66 @@ jobs:
...
+            mkdir -p /tmp/cocoapods-package-root/destroot
+            cp -R ./destroot /tmp/cocoapods-package-root/destroot
+            cp hermes-engine.podspec LICENSE /tmp/cocoapods-package-root
+
+            tar -C /tmp/cocoapods-package-root/ -czvf /tmp/hermes/output/hermes-runtime-darwin-v$(get_release_version).tar.gz .
       - run:
-          name: Copy artifacts
+          name: Package the CLI
           command: |
-            cp "build_iphoneos/github"/hermes-runtime-darwin-*.tar.gz /tmp/hermes/output
```
grabbou, line 296: This was a bug from previous PR: wrong build folder was supplied for the CLI tools.

.circleci/config.yml
```diff
@@ -309,27 +304,27 @@ jobs:
...
-            mkdir -p "$HERMES_WS_DIR"
-            ln -sf "$PWD" "$HERMES_WS_DIR/hermes"
-      - run:
-          name: Run MacOS regression tests in debug mode
-          command: |
-            cd "$HERMES_WS_DIR"
-            hermes/utils/build/configure.py
-            cmake --build ./build
-            cmake --build ./build --target check-hermes
+  # test-macos:
```
grabbou, line 307: WIP: Need to uncomment this job.

.circleci/config.yml
grabbou: In a future PR, I'd like to split this into `build-cli-tools.sh` for the CLI itself and just use plain Bash to build and package the CLI into `tgz` in a `package_apple_platforms` job. 

Unfortunately, this will take a bit more time. It is also not necessary to be implemented in this PR. Leaving it for the future.

grabbou:
Note: there is one WIP around CLI that I need to resolve.

.circleci/config.yml
grabbou: CC: @alloy 

I believe your PR fixed it for local builds, still leaving this error on the CI: https://github.com/facebook/hermes/pull/354

grabbou:
@alloy do you still need `npm` releases or we're good to drop that and just stick with a CocoaPods package?

grabbou:
Potential improvement for the future: don’t build all architectures on CircleCI. Instead, build just the ones we need for testing. Have a condition that if we’re running on master, do the full build (or when we’re running under special commit that bumps release versions). This is something to consider for the future.

grabbou:
PR is ready to be reviewed.

.circleci/config.yml
Huxpro: Yeah I'm cool with that. Let's revert it to something stricter when the new setup is getting stable.

Huxpro (commented):
Lgtm! @alloy do you have any pending reviews? 
Otherwise I can resolve the nits and internal linter failures myself and accept it.

Btw, It seems like this also fixed the `test-apple-runtime` currently failing in the master. I can't figure out what part of code you changed actually fix it and I'm wondering.

grabbou:
> Btw, It seems like this also fixed the test-apple-runtime currently failing in the master. I can't figure out what part of code you changed actually fix it (or, what was the causing the failure) and I'm wondering.

No idea! FWIW the build now takes place under `build-apple-runtime` so that's the one that's likely to fail. I am not sure I have pulled all your latest changes tho.

.circleci/config.yml
grabbou: I like that plan!

grabbou:
@Huxpro I had a minute this morning and decided to apply tweaks myself, so just needs importing (hopefully) on your side!

.circleci/config.yml
```diff
@@ -269,31 +230,70 @@ jobs:
...
+      # when it is not necessary
+      - restore_cache:
+          key: v3-repo-{{ .Environment.CIRCLE_SHA1 }}
+      - run:
+          name: Set up workspace
+          command: mkdir -p /tmp/hermes/output
+      - run:
+          name: Install dependencies
+          command: |
+            brew install cmake
```
alloy, line 248: Was this intentional?

```suggestion
            brew install cmake ninja
```

hermes-engine.podspec
```diff
@@ -32,34 +29,4 @@ Pod::Spec.new do |spec|
...
-
-        create_universal_framework "iphoneos" "iphonesimulator"
-      fi
-
-      # If MacOS framework does not exist, build one
-      if [ ! -d destroot/Library/Frameworks/macosx/hermes.framework ]; then
-        build_apple_framework "macosx" "x86_64;arm64" "#{HermesHelper::OSX_DEPLOYMENT_TARGET}"
-      fi
-    EOS
-  end
```
alloy, line 64: I think we should keep a `prepare_command` step so people can still build the pod directly from the repo, if needed.

alloy (changes requested):
Very nice 👌 

As mentioned, the one thing we should keep is a working `prepare_command` script that is able to build the artefacts directly from the repo.

alloy:
> @alloy do you still need `npm` releases and `hermes-engine-darwin` package or we're good to drop that and just stick with a CocoaPods `hermes-runtime-engine` package?

I’m unsure what the ramifications are. You only mean to stop creating new ones, not to pull the existing ones, right? If so, yup that’s good 👍

alloy:
> Btw, It seems like this also fixed the `test-apple-runtime` currently failing in the master. I can't figure out what part of code you changed actually fix it (or, what was causing the failure) and I'm wondering.

I wonder if it might be related to invoking `cmake` recursively leads to it swallowing all the output and making CI timeout? At least that appeared to be the case in some of my tests.

hermes-engine.podspec
grabbou: I was hoping that running `./utils/build-mac-framework.sh` will be enough to build directly from the repo, rather than keeping this + custom ENV to disable it in release versions. 

I am happy to keep it tho!

grabbou:
> I’m unsure what the ramifications are. You only mean to stop creating new ones, not to pull the existing ones, right? If so, yup that’s good 👍

Yes, all new ones will be released to CocoaPods and no new `npm` packages to be created.

hermes-engine.podspec
alloy: When someone runs `pod install` they do not expect to have to manually invoke anything else, so let just do that for them by invoking the right script here?

grabbou: Sounds good to me.

Huxpro:
The PR is merged!

However, the [build-apple-runtime is now failed](https://app.circleci.com/pipelines/github/facebook/hermes/1531/workflows/68182766-f05e-45db-a322-67b3dfc82202/jobs/11559). So apparently this PR did not fix the `test-apple-runtime` currently failing in the master. Maybe the CircleCI was not really testing the PR against master code? Idk.

But it was expected and it's okay for now. This PR should work fine with the 0.7.0 branch that we need for the RN 0.64.

grabbou:
> Maybe the CircleCI was just not really testing the PR against master code hence the false postive? I dunno.

I didn't start this branch with latest master updates most likely

PR #396, opened 2020-10-23
fix: bring back podspec preapre_command

grabbou:
This PR brings back `spec.prepare_command` to make it easier to use Hermes while in development.

Ref: https://github.com/facebook/hermes/pull/393#discussion_r509340544


hermes-engine.podspec
```diff
@@ -29,4 +29,18 @@ Pod::Spec.new do |spec|
...
+    spec.prepare_command = <<-EOS
+      # When true, debug build will be used.
+      # See `build-apple-framework.sh` for details
+      DEBUG=#{HermesHelper::BUILD_TYPE == :debug}
+
+      # Build iOS framework
+      ./utils/build-ios-framework.sh
+      
+      # Build Mac framework
+      ./utils/build-mac-framework.sh
```
grabbou, line 43: The previous `if` logic is now in each file, so nothing changed, just organised.

alloy (approved):
Yup, this is perfect 👌 Thanks! 🙏

PR #399, opened 2020-10-31
Memory profiler patches

dulinriley:
## Summary

Cherry pick the following commits into release-v0.7:
These are all about the memory profiler, and also ensure the JSI API is compatible with RN 0.64.

* b7177cba82b06e6f482b6597b7195f5daa7757a8
* ea2b5aa5a7a792cf599f5b06b5503c280d364d85
* 801aa8a14bae9ec221b76259739ae8ea298bd3e4
* 73594b5f568567c7469c88316ca698ef6df21685
* 1d8d495fe6f1fc9e31e4403644634c77aaf36d6f
* cfaf270740b165c5810a6c1ec29e41deaf2cab90

Include some merge conflict fixes.

## Test Plan

CI will build a Hermes library to embed into an app.
Download that into a test app, connect Chrome following the instructions here:
https://reactnative.dev/docs/hermes#debugging-hermes-using-google-chromes-devtools

Right now the memory profiler crashes, but it's not crashing from a linker error, just a bug in the
memory profiler that will be fixed somewhere else, and will not block shipping a v0.7.2 or RN v0.64


dulinriley:
Ok I verified that RNTester worked with this release. I'm seeing a crash when using the Heap profiler, but that might be an unrelated bug that hasn't been resolved in hermes master either.
This should be good to merge.

dulinriley:
I'll need to include two other commits as well:
* 1d8d495fe6f1fc9e31e4403644634c77aaf36d6f
* cfaf270740b165c5810a6c1ec29e41deaf2cab90

I'll update this PR with those two added

Huxpro:
With the great help from @alloy, I can verify that this PR (prior to the 2 new commits) did resolve the `facebook::jsi` errors from the RN iOS Hermes build (tested on `RNTesterPods.xcworkspace` with `USE_FRAMEWORKS=1 USE_HERMES=1` setting).

Before:
image

After:
image

We are in the good track! Let's wait the CI to build the artifacts with the 2 new commits and try one more time.

dulinriley:
Oh whoops I left a conflict marker in one of those commits. I'll edit and force-push some new commits, which will restart the CI jobs.

dulinriley:
Ok force-pushed after verifying `cmake --build ../build -t check-hermes` succeeded, now the CI should have the right stuff.

dulinriley:
Ok downloaded the new npm and tested with RNTester. Seems to work fine, although heap snapshots seem to be throwing an exception, and the memory profiler is still not working. None of these issues are worth blocking the React Native release, although I'll see if I can attach a debugger

There's no difference between `jsi.h` and `instrumentation.h` between the repos anymore, so this should be safe to land.

Huxpro:
For visibility, I tested the artifact from the updated PR. Seems like there is some other RN related breakages. But at least we've got passed JSI/Hermes errors hence we decided to merge it.

image

This is some errors that I don't know so cc @alloy @grabbou for helps. I'll re-upload the Github Release of v0.7.0 one more time so you guys can easily purge the CocoaPod cache to see if it repro and how to resolve it.

- [ ] release re-uploaded

grabbou:
Awesome, thank you. I will test it on my branch and I think we can treat CI failures as a good indicator of what is there to be done.

@Huxpro It may be worth to just release another `patch` update. The 0.7.1 is already used in React Native `master` and while this is not publicly consumed anywhere outside, in the nearest future, I think it would be better to keep releasing traditional patches.

PR #410, opened 2020-11-19
Bringing up E2E test job in Hermes Github repo.

mganandraj:
## Summary

This change adds a new Job to Hermes CI pipeline to run the Android E2E tests. The primary motivation behind the change is to have E2E verification of the new Intl/ECMA402 APIs on Android platform as the implementation calls Android APIs.

A couple of fixes in the tests are made to ensure that the job runs cleanly.

## Test Plan

Only test code which is passing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -609,6 +609,8 @@ jobs:
             cd "$HERMES_WS_DIR/hermes/android"
             # We need to build only x86 for running the tests. This saves a lot of computing resources.
             echo 'abis=x86' >> gradle.properties
+            adb logcat -c || true
             ./gradlew :intltest:preparetest262 && ./gradlew :intltest:connectedAndroidTest
+            adb logcat -d > /tmp/hermes/output/adblog.txt
       - store_artifacts:
           path: /tmp/hermes/output/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,7 @@ workflows:
       - test-e2e:
           requires:
             - npm
+      - test-e2e-intl
 
 # Default settings for Apple jobs (apple-runtime, test-apple-runtime)
 apple_defaults: &apple_defaults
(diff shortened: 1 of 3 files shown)


mhorowitz:
This looks great!  I don't fully understand it, but I can see the tests pass in circleci, and the output looks right.

I did notice one odd thing.  I'll point it out in case it's concerning, but I don't really know what it means.  In the job output at https://app.circleci.com/pipelines/github/facebook/hermes/1655/workflows/9ea6c655-b00c-467f-91f2-13007d482b33/jobs/12358, the "Run emulator in background" step has a grey dash instead of a green check, and the logs end with "Build was canceled".  Maybe this is normal behavior for a background job?

mganandraj:
> This looks great! I don't fully understand it, but I can see the tests pass in circleci, and the output looks right.
> 
> I did notice one odd thing. I'll point it out in case it's concerning, but I don't really know what it means. In the job output at https://app.circleci.com/pipelines/github/facebook/hermes/1655/workflows/9ea6c655-b00c-467f-91f2-13007d482b33/jobs/12358, the "Run emulator in background" step has a grey dash instead of a green check, and the logs end with "Build was canceled". Maybe this is normal behavior for a background job?

The task you mentioned is marked "background" which essentially runs asynchronously, which makes the emulator boot process run in parallel to the rest of the job execution.  ..  Later, another task waits for the boot completion and triggers the gradle task which builds and tests .. The emulator background task gets cancelled when the job ends .. 

There may be a way to join back to a background task in CircleCI, which i haven't looked into yet. Alternatively we could make the emulator task foreground at the expense of increased job execution time. I think, we could leave it as it is as it is not affecting the job status

mhorowitz:
> There may be a way to join back to a background task in CircleCI, which i haven't looked into yet. Alternatively we could make the emulator task foreground at the expense of increased job execution time. I think, we could leave it as it is as it is not affecting the job status

I agree, it's fine for now.  I just wanted to make sure it wasn't going to cause any unexpected issues.  Thanks for explaining.

PR #426, opened 2020-12-04
Intl API: JNI initialization, Proguard annotations and intl build definitions.

mganandraj:
## Summary

This PR has fixes to Intl API implementation to make it work properly under release configurations. With this PR, we also start building the Intl code into hermes release packages.

Essentially, the PR has the following changes,

1. Add Proguard annotations to keep the JNI referenced Java symbols from being stripped away in minified packages
2. Ensure fbjni is initialized when loading libhermes.so.
3. Add runtime config switch to control the inclusion of Intl object in the global. We keep it OFF by default.
4. Changes in the gradle build definitions to start building C++ and Java Intl code into the hermes release package. We use product flavors to clearly differentiate the Intl enabled packages for now being. We can probably clean things up once the Intl APIs are stabilized.
5. Updated the documentation (Unfortunately still a docx file, at \lib\Platform\Intl\java\com\facebook\hermes\intl\HermesIntlAPIsImplementationNotes.docx) to include the impact on the size of the application package due to this change.

As the runtime switch is currently turned OFF, the real impact of this change is going to be the increased package size. I'm pasting the measured numbers here for quick reference,

Impact on Application Size
The following numbers are measured using a test application which takes dependency on the Hermes library to evaluate a JavaScript snippet. Essentially, enabling Intl APIs adds 57-62K per ABI.
Product APK Size	NOINTL	INTL	DIFF	PERC
ARM64	1,672,235	1,729,579	57,344	3.43%
ARM	1,471,539	1,528,883	57,344	3.90%
X86_64	1,844,255	1,901,599	57,344	3.11%
X86	1,950,739	2,012,179	61,440	3.15%

The overhead is contributed by both compiled native C++ and Java bits. 
The uncompressed size of the Hermes shared library got bigger as follows,

libhermes.so Size	NOINTL	INTL	DIFF	PERC
ARM64	2,473,760	2,551,592	77,832	3.15%
ARM	1,696,672	1,754,016	57,344	3.38%
X86_64	2,633,528	2,711,368	77,840	2.96%
X86	2,859,916	2,945,936	86,020	3.01%

And the Java bits got bigger as well,
Java Size	NOINTL	INTL	DIFF	PERC
classes.jar (in hermes.aar)	    559	120975	120,416	21541.32%
classes.dex (intltestapp.apk)	160708	234808	74,100	46.11%

Please note that the application dex file contains non-hermes class files too.
And finally, this is the increase in the final npm package,
NPM Package	NOINTL	INTL	DIFF	PERC
	214447973	219291220	4,843,247	2.26%

I'm not including the other performance metrics (Memory footprint, latency etc.) with this change because,
1. Intl APIs are currently disabled by default. 
2. I've measured some metrics and didn't see anything egregious, but I know a couple of potential improvement opportunities, which I want to plug before sharing the numbers, which i'll do as part of enabling Intl by default.

## Test Plan
I've a test application which was extremely useful in diagnosing and debugging hermes on Android, which I'll include in another PR.
There is no change in the product code which can affect the functionality, and all the tests are passing.


ar-g:
Hi, this is not related to PR, but I couldn't find a better place.

I wanted to test our project with Intl, so I've built the latest Hermes master and added it to RN 0.64-rc1. It still crashing with Intl not found. 

Do I need to add any extra configuration? I see there's an extra flag passing 

```
externalNativeBuild {
        cmake {
          arguments "-DHERMES_ENABLE_INTL=True"
        }
      }
```

API/hermes/JNI_OnLoad.cpp
```diff
@@ -0,0 +1,5 @@
+#include <fbjni/fbjni.h>
+
+extern "C" jint JNIEXPORT JNICALL JNI_OnLoad(JavaVM *jvm, void *reserved) {
```
mhorowitz, line 3: This is here presumably to implement "Ensure fbjni is initialized when loading libhermes.so."  But this is a bit surprising to me, as hermes.so isn't directly loaded by Java, so I would not think this would be called.

I'm guessing you did this to fix a problem of some sort, but it's not clear what the problem was.  Can you say what you were trying to fix with this?

mhorowitz (commented):
Generally, this looks good!  One question below.

API/hermes/JNI_OnLoad.cpp
mganandraj: Hi @mhorowitz 

As we discussed offline, it was required because hermes is not built to depend on libfbjni.so, but instead directly compiling the fbjni c++ sources into the libhermes.so. This results in libhermes.so having a private copy of the jvm pointer.

But, later when I tried with react-native, the fbjni calls to Java code seems to work even without this initialization. Hence, there is something going on which i haven't understood well enough.

Please let me know if you want me to try to look into the details here.

mhorowitz: @mganandraj I think your observation about libhermes including its own fbjni is on point and relevant.  I think the correct long-term fix is for Hermes to depend on prebuilt external fbjni from maven as RN does, so they can both share the same libfbjni.so.  But this is not a simple fix, so for now, including JNI_OnLoad in libhermes is an ok workaround, but we should try to fix this for real, too.

mhorowitz:
Trying to land, I'm getting a lot of errors like
```
/tmp/tmp.Xk98BTFios/hermes/lib/Platform/Intl/java/com/facebook/hermes/intl/NumberFormat.java:15: error: package com.facebook.proguard.annotations does not exist
import com.facebook.proguard.annotations.DoNotStrip;
                                        ^
```
which makes sense as hermes doesn't have com.facebook.proguard.

The android CI test is failing, but it's due to a problem which is already fixed.  If you rebase to master, then you should get more accurate CI results.  I don't know if you need to change gradle to refer to the right path for the proguard files, or what.

mganandraj:
> Trying to land, I'm getting a lot of errors like
> 
> ```
> /tmp/tmp.Xk98BTFios/hermes/lib/Platform/Intl/java/com/facebook/hermes/intl/NumberFormat.java:15: error: package com.facebook.proguard.annotations does not exist
> import com.facebook.proguard.annotations.DoNotStrip;
>                                         ^
> ```
> 
> which makes sense as hermes doesn't have com.facebook.proguard. I'm not quite sure how to fix this.
> 
> Our internal tests run
> 
> ```
> $GRADLE -Pabis=x86,arm64-v8a hermes:build
> $GRADLE -Pabis=armeabi-v7a build intl:build
> $GRADLE -Pabis=x86_64 build intltest:build
> ```
> 
> And it appears the failure is on the second command. Hopefully you can repro the failure with that.

The DoNotStrip is brought in through dependency to "com.facebook.yoga:proguard-annotations:1.14.1" .. 

"intl" project has this dependency defined by is under "androidTestImplementation" which seems to be causing issue here even though i still can't reproduce it locally .. I'm pushing another update removing bulding "intl" project as it is no longer needed, and the second PR has the changes to clean it up.

mhorowitz:
https://github.com/facebook/hermes/pull/426/files/3c667a277f3f8518c98f277d3b4d275aa4dfc147..f30044da909a700241ba536605ab541f8b98871d shows changes besides the ones we discussed.

Some of these changes are obvious, but some are not.  Can you remove the changes which aren't strictly necessary to fix the land failures?  We can always make more changes later, but I'd prefer to avoid creeping changes into a PR this late without some discussion of why.

mganandraj:
> Some of these changes are obvious, but some are not. Can you remove the changes which aren't strictly necessary to fix the land failures? We can always make more changes later, but I'd prefer to avoid creeping changes into a PR this late without some discussion of why. This risks slowing down the process, and I only have one more work day left this year.

Almost all changes are purely for silencing linter  .. None are necessary (Made supposedly to make the pull smoother), hence removed all changes other than removing "intl" module from building.

lib/VM/JSLib/GlobalObject.cpp
```diff
@@ -754,12 +754,15 @@ void initGlobalObject(Runtime *runtime, const JSLibFlags &jsLibFlags) {
...
   // Define the global Intl object
   // TODO T65916424: Consider how we can move this somewhere more modular.
-  runtime->ignoreAllocationFailure(JSObject::defineOwnProperty(
-      runtime->getGlobal(),
-      runtime,
-      Predefined::getSymbolID(Predefined::Intl),
-      normalDPF,
-      intl::createIntlObject(runtime)));
+  
+  if (LLVM_UNLIKELY(runtime->hasES6Intl())) {
```
Huxpro, line 758: Hey @mganandraj, I'm a little confused by the purpose of having this runtime config. 

If Hermes would be compiled with the Intl flag on for OSS, then the binary size has been already increased. Does unset this runtime config help with any savings?

Chaidez5:
Development of the world

PR #427, opened 2020-12-04
A test application for Hermes on Android and some cleanups

mganandraj:
## Summary

This change contains some test utilities and some android build definition cleanups. Specifically,

1. A new test app (intltestapp) which has
1a. A REPL activity which uses (copies!) the driver code from the native hermes (tools/hermes) repl application. We should find a way to share the code and avoid the duplication, which hopefully can be done later. The tool has some nice additions such as command history. I think this will aid big time in our future diagnosis and debugging.
1b. An EVAL activity which directly links to the hermes aar library (instead of linking directly against external libraries) like the external consumers do.
2. Removing the "intl" project which no longer serves any purpose.

## Test Plan
No change in the product code.


ar-g:
Hi, this is not related to PR, but I couldn't find a better place.

I wanted to test our project with Intl, so I've built the latest Hermes master and added it to RN 0.64-rc1. It still crashing with Intl not found. 

Do I need to add any extra configuration? I see there's an extra flag passing 

```
externalNativeBuild {
        cmake {
          arguments "-DHERMES_ENABLE_INTL=True"
        }
      }
```

mganandraj:
> Hi, this is not related to PR, but I couldn't find a better place.
> 
> I wanted to test our project with Intl, so I've built the latest Hermes master and added it to RN 0.64-rc1. It still crashing with Intl not found.
> 
> Do I need to add any extra configuration? I see there's an extra flag passing
> 
> ```
> externalNativeBuild {
>         cmake {
>           arguments "-DHERMES_ENABLE_INTL=True"
>         }
>       }
> ```

Hi, This CPP preprocessor flag (HERMES_ENABLE_INTL) should be set to have Intl objects and constructors available in global .. 
Plus the PR:https://github.com/facebook/hermes/pull/426 needs to go in to make Intl work in release configurations .. 
Please let me know if you have any questions! Thanks!

mhorowitz:
@mganandraj can you give some more background about this app?  It's hard to understand the purpose of it.  My questions are based on what we were able to surmise, but it's possible I've misinterpreted the intent here.

* Can you create a screen video demonstrating the main use cases, so we can more easily understand the intent?
* It seems to be both an ad hoc testing REPL, and an automated test helper.
* As an ad hoc tool, it seems useful.  But as you say, we should try to address the code duplication.
* As an automated tool, there is no documentation or usage of the robots.py file.  Does this tool enable any kinds of automation not enabled by the com.facebook.hermes.test testing infra?  Do we want two different kinds of automation?
* If we do want two kinds of automation, should this new one be integrated with CircleCI?

Thanks!

mganandraj:
> @mganandraj can you give some more background about this app? It's hard to understand the purpose of it. My questions are based on what we were able to surmise, but it's possible I've misinterpreted the intent here.
> 
> * Can you create a screen video demonstrating the main use cases, so we can more easily understand the intent?
> * It seems to be both an ad hoc testing REPL, and an automated test helper.
> * As an ad hoc tool, it seems useful.  But as you say, we should try to address the code duplication.
> * As an automated tool, there is no documentation or usage of the robots.py file.  Does this tool enable any kinds of automation not enabled by the com.facebook.hermes.test testing infra?  Do we want two different kinds of automation?
> * If we do want two kinds of automation, should this new one be integrated with CircleCI?
> 
> Thanks!

Hi @mhorowitz, 

Thanks a lot for taking a look !

Here is a quick video on the REPL app usage: https://1drv.ms/f/s!AjHU7UiFktXxl4UiKWNuSijOSINftA

Essentially, having an ad-hoc REPL app enables us to quickly try and debug various combinations and parameterizations of scripts. The existing command line REPL application (named hermes) doesn't allow us to debug any Android specific code such as the Intl implementation has. This tool is essentially an Android application shell around the existing REPL tool. 

The tool had another entry point which loads libhermes.so in release mode just like other applications (such as react-native). I'd removed that entry point  to keep this PR smaller and manageable. Also, I think its value will go down once this Intl implementation is stabilized,

I used it for some automated perf measurements using the python script included in the PR. The repl application can accept a script to be evaluated through adb intent, which is utilized by the script. The script then issues adb "dumpsys" commands to query the memory footprints, and write the processed output to file to be easily consumed (for e.g. copy/paste to excel to build charts) . This script can be extended in future to add more capabilities and run as part of the CI, for e.g. performance regression tests.

Currently, the script is not parameterized at all, and the scripts and the workflow is hardcoded. I've added a short comment on top to explain what it currently does. 

Also, i've made some changes in the latest iteration to avoid most of the code duplication.

tools/hermes/evaluate-line.js
```diff
@@ -14,15 +14,28 @@ C_STRING((function() {
...
-    colors.red = '\033[31m';
-    colors.green = '\033[32m';
-    colors.yellow = '\033[33m';
-    colors.blue = '\033[34m';
-    colors.magenta = '\033[35m';
-    colors.cyan = '\033[36m';
-    colors.white = '\033[37m';
 
-    colors.reset = '\033[0m';
+    if (typeof _replterminaltype !== "undefined" && _replterminaltype === "android") {
```
avp, line 18: This is supposed to be a property on the global object, right?

```suggestion
    if (globalThis._replterminaltype === "android") {
```

android/intltestapp/cpp/ReplProxy.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+        vm::Handle <vm::JSObject> global = mRuntime->getGlobal();
+        auto propRes = vm::JSObject::getNamed_RJS(
+                global, mRuntime.get(), vm::Predefined::getSymbolID(vm::Predefined::eval));
+        if (propRes == vm::ExecutionStatus::EXCEPTION) {
+            mRuntime->printException(
+                    llvh::outs(), mRuntime->makeHandle(mRuntime->getThrownValue()));
+            return "error getting 'eval' from global";
+        }
+        auto evalFn = mRuntime->makeHandle<vm::Callable>(std::move(*propRes));
```
avp, line 101: Is it possible to cache `evalFn` when we create the Runtime so we don't have to look it up each time?

android/intltestapp/cpp/ReplProxy.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+        bool hasColors = true;
+
+        // evaluate-line.js will color the output appropriately based on the available of this property in the global object.
+        vm::HermesValue terminalTypePropName = vm::HermesValue::encodeStringValue(vm::StringPrimitive::createNoThrow(mRuntime.get(), "_replterminaltype").get());
+        vm::HermesValue terminalTypePropVal = vm::HermesValue::encodeStringValue(vm::StringPrimitive::createNoThrow(mRuntime.get(), "android").get());
+        if(vm::ExecutionStatus::EXCEPTION ==
+            global->putComputed_RJS(global,
```
avp, line 109: This should probably only happen once at construction instead of every time we run a line of JS.

avp (commented):
@mganandraj since this is just a nice-to-have testing app for Android, that's great (I've certainly thought that would be useful in the past).

I think we should rename it from `intltestapp` to `ReplApp` or something similar to make it more obvious that it's generally useful, not just for `Intl`.

Some minor comments as well.

Huxpro:
@mganandraj the video looks awesome! I also feel this could be a more general purpose REPL app, despite it's mainly for Intl test automation for now.

I'd love to use it and contribute to it for fun ;) Specifically I want to make it looks nicer with monospaced font and closer to Hermes website color schema, but I need to learn some native Android dev first 😜. It might be helpful to have a README to guide people on whatever they need to know to contribute to it ;)

android/intltestapp/cpp/ReplProxy.cpp
```diff
@@ -0,0 +1,195 @@
...
+                                                   makeNativeMethod("initHybrid",
+                                                           ReplProxy::initHybrid),
+                                                   makeNativeMethod("nativeEvalScript",
+                                                           ReplProxy::nativeEvalScript),
+                                                   makeNativeMethod("nativeCollect",
+                                                           ReplProxy::nativeCollect),
+                                                   makeNativeMethod("nativeHeapStats",
+                                                           ReplProxy::nativeHeapStats)
+                                           });
+    }
```
Huxpro, line 28: This indentation looks really weird. Can we reuse the same formatting setup as Hermes C++ source?

android/intltestapp/src/main/res/values/strings.xml
```diff
@@ -0,0 +1 @@
+<resources></resources>
```
Huxpro, line 1: Is this useful?

android/intltestapp/cpp/ReplProxy.cpp
mhorowitz: When I import this, I'll run it through our linter.

tools/hermes/evaluate-line.js
mganandraj: It worked without ! And i believe the undefined variables will be looked up in global object when in non-strict mode ! Anyways, i fixed the code to make it explicit in the latest iteration ..

android/intltestapp/cpp/ReplProxy.cpp
mganandraj: Done in the latest iteration ..

android/intltestapp/cpp/ReplProxy.cpp
mganandraj: It's the default indentation chosen by the default Android Studio formatting .. It is definitely Wierd ! We need to pick a suitable clang formatting rules .. I'll look into it when making the next change ..

android/intltestapp/src/main/res/values/strings.xml
mganandraj: It should be deletable.. Will do when touching the source next time.

mganandraj:
> @mganandraj the video looks awesome! I also feel this could be a more general purpose REPL app, despite it's mainly for Intl test automation for now.
> 
> I'd love to use it and contribute to it for fun ;) Specifically I want to make it looks nicer with monospaced font and closer to Hermes website color schema, but I need to learn some native Android dev first 😜. It might be helpful to have a README to guide people on whatever they need to know to contribute to it ;)

You are absolutely welcome to contribute and let me know if you need any help from my side .. The current UI is cooked up hastily and and it looks ugly !

A README is a good idea, but it's not clear to me what to put in it at this moment. I'll add one the next time i touch the code.

Huxpro (changes requested):
> A README is a good idea, but it's not clear to me what to put in it at this moment. I'll add one the next time i touch the code.

Maybe something like...
1. an introduction of what this app does.
2. how to build and run it. It's not as obvious as just open it in Android Studio and it just works. it requires following <https://hermesengine.dev/docs/react-native-integration> to setup a workspace with a release build and all the right env vars.
3. something about `robot.py` maybe?

> As an automated tool, there is no documentation or usage of the robots.py file. Does this tool enable any kinds of automation not enabled by the com.facebook.hermes.test testing infra? Do we want two different kinds of automation?
If we do want two kinds of automation, should this new one be integrated with CircleCI?

I also want to second @mhorowitz in the above questions. You did describe what `robots.py` does conceptually but I'm still not sure what's the expected behaviors that one can observe and how to interpret the results?  How can someone make sure their changes to the app won't break it? 

I'm also getting `FileNotFound` error when trying to run `robot.py` locally.

android/intltestapp/cpp/ReplProxy.cpp
Huxpro: I'll run this PR through our internal linters to make sure everything is formatted properly before merging it. But I'm not sure how to enforce those rules externally :(

android/replapp/cpp/ReplProxy.cpp
```diff
@@ -0,0 +1,207 @@
...
+    vm::HermesValue getEvaluateLineFn(const std::shared_ptr<vm::Runtime>& runtime) {
+
+        vm::GCScope gcScope(runtime.get());
+        runtime->getHeap().runtimeWillExecute();
+
+        vm::Handle <vm::JSObject> global = runtime->getGlobal();
+
+        // evaluate-line.js will color the output appropriately based on the available of this property in the global object.
+        vm::HermesValue terminalTypePropName = vm::HermesValue::encodeStringValue(vm::StringPrimitive::createNoThrow(runtime.get(), "_replterminaltype").get());
+        vm::HermesValue terminalTypePropVal = vm::HermesValue::encodeStringValue(vm::StringPrimitive::createNoThrow(runtime.get(), "android").get());
```
Huxpro, line 51: You can't pass `vm::HermesValue` around across `putComputed_RJS` which might allocate (hence result in GC and move objects). Instead, `vm::StringPrimitive::createNoThrow` already returns `Handle` and you can directly pass those `Handle` without the need to wrap them in `makeHandle` at callsite.

android/replapp/scripts/buildall.sh
```diff
@@ -0,0 +1,14 @@
+export HERMES_WS_DIR=~/code/hermes_ws_perf/
```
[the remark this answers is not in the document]
mganandraj: Ah .. I was using this scripts some time back, but is not required to be submitted. Deleted it in the latest iteration. Thanks!

android/replapp/cpp/ReplProxy.cpp
mganandraj: Thanks ! Done.

PR #440, opened 2021-01-05
Fix Internal Bytecode OSS breakage

Huxpro:
## Summary

The CircleCI is currently broken on the trunk due to missing 
internal bytecode when building apple frameworks.

This diff set up the two staged build to recover the CI by 
1. build a native hermesc to precompile internal bytecode
2. build the cross-compiled hermes lib. 

Noted that Apple cross-compilation is different from other cross-compilation, 
so instead we use `HERMES_APPLE_TARGET_PLATFORM` to tell if an import hermesc 
should be used.

## Test Plan

CircleCI.

N.B. local `circleci` cli tool doesn't work with macOS environment.

You should notice a `hermesc` is built in a native setting first:

```
-- CMAKE_HOST_SYSTEM_NAME = Darwin
-- CMAKE_SYSTEM_NAME = Darwin
-- HERMES_APPLE_TARGET_PLATFORM =
-- CMAKE_CROSSCOMPILING = FALSE
-- Build files have been written to: /path/to/hermes/build_host_hermesc
```

then the apple universal framework should be built in a cross-compilation setting 
successfully (instead of failing and complaining missing internal bytecode)

```
-- CMAKE_HOST_SYSTEM_NAME = Darwin
-- CMAKE_SYSTEM_NAME = Darwin
-- HERMES_APPLE_TARGET_PLATFORM = ...
-- CMAKE_CROSSCOMPILING = TRUE
```

For some reasons the sh doesn't work on my local but it works on the CI.

```
λ ./utils/build-mac-framework.sh
...
ld: dynamic main executables must link with libSystem.dylib for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

cc @alloy @grabbou


Huxpro:
Wow build-apple-runtime and test-apple-runtime are all passed in the first run! Undraft it ;)

alloy (approved):
Great 👌

Issue #358, opened 2020-09-20
Assertion Failure at HandleRootOwner.h:413


HongxuChen:
## Bug Description

When running `hermes $JSFILE` with some crafted JSFILEs, `hermes` may trigger the assertion failure.

- [x] I have confirmed this bug does not occur with JSC

Hermes version: 0.7.0 (git ab7c2cfa)
React Native version (if any):
Android version (if any):
Platform (most likely one of arm64-v8a, armeabi-v7a, x86, x86_64):  x86_64

## Steps To Reproduce

1. Craft a javascript file `test.js` with the following content:
```javascript
h = {}
p = new Proxy({}, h)
h.__proto__ = p;
try { p._ } catch {}
```

2. run `./hermes test.js` and it reports:

```
hermes: /home/hongxu/work/hermes/hermes-asan/include/hermes/VM/HandleRootOwner.h:413: hermes::vm::PinnedHermesValue* hermes::vm::GCScope::newHandle(hermes::vm::HermesValue): Assertion `getHandleCountDbg() < handlesLimit_ && "Too many handles allocated in GCScope"' failed.
```

## The Expected Behavior
No assertion triggered.

HongxuChen:
see also #279

mhorowitz:
closed by https://github.com/facebook/hermes/commit/4108bf9b66d16dbddde55e951c247d70aaedc64e

WilliamParks:
The following code produces the same assertion error on commit b0102191e58f82de45ce16c6ba3b2ab74feeb537

```
var a = new Proxy([],{valueOf:13.37});
var b = new Proxy(a,Reflect);
for (var c in b) {
}
```

WilliamParks:
Here are some additional samples, which trigger the assertion against commit 7d4cb023f9cfc87d2b0e8144ccad3bd329269af8
```
function main() {
    var v1 = "9";
    var v4 = 0 < 10;
    while (v4 != 1024) {
        v1 += "$1";
        "".replace(/^(.*)/,v1);
        v4++;
    }
}
main();
```

```
var v0 = {};
var v3 = {__v_2:1};
var v4 = v3;
var v7 = {};
var v8 = new Proxy(v7,v4);
v0.__proto__ = v8;
function v9() {
    return v0;
}
v4.getPrototypeOf = v9;
var v11 = Object.prototype;
var v12 = v11.isPrototypeOf;
v12.call(0,v8);
```
```
[...]

Issue #279, opened 2020-06-28
Assertion `getHandleCountDbg() < handlesLimit_ && "Too many handles allocated in GCScope"' failed.


Changochen:
Git commit: `35da03ff3cd5c7c5391e809d8d1d99d8f227be0b`

CMD:`hermes poc.js`

POC:
```javascript
b = false
function c() {
  if (!b) {
    a = Array(3000)
    this[2] = a
  }
  b = true
}
JSON.parse("[1, 2, []]", c)
```

dulinriley:
I've confirmed it's happening for me as well on master.

dulinriley:
Fixed in 1be5a7bdccf47636ad93a0365be07c2b673be7d3, thanks for reporting this!

PR #441, opened 2021-01-05
Add Fuzzilli Instrumentation

shouc:
## Summary
Fuzzilli (https://github.com/googleprojectzero/fuzzilli) instrumentation is set up and you can now use Fuzzilli to fuzz this project. This achieves a lot higher coverages than LibFuzzer's. Previous bugs found by Fuzzilli are #358 and #279.


WilliamParks:
@shouc Have you considered adding this to Fuzzilli as well? 

For the FB folks, Fuzzilli found CVEs 2020-1912 and 2020-1914.

shouc:
> @shouc Have you considered adding this to Fuzzilli as well?
> 
> For the FB folks, Fuzzilli found CVEs 2020-1912 and 2020-1914.

Not sure what FB folks would like. If it is merged then we can add it to Fuzzilli by just adding a new profile.

lib/VM/JSLib/GlobalObject.cpp
```diff
@@ -764,6 +777,11 @@ void initGlobalObject(Runtime *runtime, const JSLibFlags &jsLibFlags) {
         intl::createIntlObject(runtime)));
   }
 #endif
+
+#ifdef HERMES_ENABLE_FUZZILLI
```
dulinriley, line 781: Instead of defining these functions inside of Hermes internally, I recommend instead defining them in fuzzili.cpp.
That should remove the need for adding the `HERMES_ENABLE_FUZZILLI` preprocessor variable and changing all the internal string and function files.

CMakeLists.txt
```diff
@@ -152,8 +152,15 @@ set(HERMES_ENABLE_UNDEFINED_BEHAVIOR_SANITIZER OFF CACHE BOOL
 set(HERMES_ENABLE_THREAD_SANITIZER OFF CACHE BOOL
   "Enable -fsanitize=thread")
 
-set(HERMES_ENABLE_FUZZING OFF CACHE BOOL
-  "Enable fuzzing")
+# Enable Trace PC Guard
+set(HERMES_ENABLE_TRACE_PC_GUARD OFF CACHE BOOL
```
dulinriley, line 156: What does `-fsanitize-coverage=trace-pc-guard` do? Should it be enabled for any fuzzing mode?
I would say just manage this flag with HERMES_ENABLE_FUZZILLI instead of making a separate variable for it. If libfuzzer needs it as well you can add the flag if either is defined.

tools/fuzzilli/fuzzilli.cpp
```diff
@@ -0,0 +1,172 @@
...
+using facebook::jsi::HostObject;
+using facebook::jsi::JSIException;
+using facebook::jsi::Object;
+using facebook::jsi::PropNameID;
+using facebook::jsi::Runtime;
+using facebook::jsi::String;
+using facebook::jsi::StringBuffer;
+using facebook::jsi::Value;
+
+//
```
dulinriley, line 22: Is this body of code some well-known thing? I have no idea what it's doing or why.
Are there any libraries that provide this for us like something clang provides?

tools/fuzzilli/fuzzilli.cpp
```diff
@@ -0,0 +1,172 @@
...
+      }
+      fflush(stdout);
+      fflush(stderr);
+      auto status = ((exceptionThrew ? 1 : 0) & 0xff) << 8;
+      CHECK(write(REPRL_CWFD, &status, 4) == 4);
+      __sanitizer_cov_reset_edgeguards();
+    }
+
+  } else {
+    // peacefully quit
```
dulinriley, line 168: Make this an early return case instead, to reduce indentation.

tools/fuzzilli/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
dulinriley, line 1: We made the fuzzers directory so that you can place different types of fuzzers in a single directory.
Just move fuzzili.cpp there and add this cmake stuff to the existing cmakelists.txt file

tools/fuzzilli/fuzzilli.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+//
+// END FUZZING CODE
+//
+
+int main(int argc, char** argv){
+
+  if (argc == 2 && strcmp(argv[1], "--replr") == 0){
+    // replr mode
+    auto runtime = makeHermesRuntime();
```
dulinriley, line 116: Inject the fuzzili function here instead, it'll be a lot fewer changes to Hermes files.

```suggestion
    auto runtime = makeHermesRuntime();
    runtime->global().setProperty(
      *runtime,
      "fuzzili",
      Function::createFromHostFunction(
        *runtime,
        facebook::jsi::PropNameID::forAscii(*runtime, "fuzzilli"),
        1,
        [](Runtime& rt,
             const Value& thisArg,
             const Value* args,
             size_t numArgs) -> Value { ... }));
```

utils/build/configure.py
```diff
@@ -69,6 +69,7 @@ def parse_args():
     parser.add_argument("--enable-asan", dest="enable_asan", action="store_true")
     parser.add_argument("--enable-ubsan", dest="enable_ubsan", action="store_true")
     parser.add_argument("--enable-tsan", dest="enable_tsan", action="store_true")
+    parser.add_argument("--enable-trace-pc-guard", dest="enable_trace_pc_guard", action="store_true")
```
dulinriley, line 72: You can just use `--cmake-flags` for this kind of ad-hoc stuff

dulinriley (changes requested):
Thanks for doing this! I think you can simplify it a deal by injecting the "fuzzili" function in "fuzzili.cpp" instead of in the Hermes VM files.

CMakeLists.txt
shouc: https://clang.llvm.org/docs/SanitizerCoverage.html It is for generating coverage information by inserting `__sanitizer_cov_trace_pc_guard` defined below to every edges. There are other use cases (coverage testing, etc.). The older version of Libfuzzer benefits from this flag. I feel it might be better to keep it as an independent flag.

tools/fuzzilli/fuzzilli.cpp
shouc: This provides the coverage info. It is adapted from Google V8.

shouc: I've made it clearer.

dulinriley:
Thanks for working on this, it looks good.
It appears that along the way the injection of the "fuzzili" function was lost, I can't find it anymore. Is that intentional?
I can import it and commit it to master if you think it's ready.

We'd also love to land this support into Fuzzili itself. Does Google run Fuzzili continuously? Perhaps as a part of OSS-fuzz which we already use?

shouc:
For instrumentation, it is intentional and made to be js function `FuzzilliCrash(int)` instead. I can also start a PR to oss-fuzz if you would like.

dulinriley:
Yes a PR to fuzzili and oss-fuzz would be great. Please add me as a subscriber to those. I'll import and commit this

tools/fuzzers/fuzzilli/fuzzilli.cpp
```diff
@@ -0,0 +1,214 @@
...
+//
+// END FUZZING CODE
+//
+
+
+int main(int argc, char **argv) {
+  if (argc != 2 || strcmp(argv[1], "--replr") != 0)
+    return 0;
+  // replr mode
+  auto runtime = makeHermesRuntime();
```
dulinriley, line 137: I'm not familiar with what a "REPLR" server means, but if the intention is to test each script separately, it would be better to instantiate the runtime once per script.
If you share a runtime amongst multiple scripts, then one script may only fail because a previous script ran in the same process.

shouc: REPLR is meant to use one runtime for a lot of scripts to improve efficiency. I checked V8 and JSC's impl and it seems that a common way to do separation is by 
```js
function x(){ [PAYLOAD] }; x();
```
It is updated now.

WilliamParks: Anything that could reasonably be corrupted during a run needs to be reset, in each loop of the REPRL (read evaluate print reset loop).

The issue is if an input corrupts some internal state but doesn't crash, a later round will crash and the fuzzer will misattribute it to the later input, without knowing about the previous round. Reseting the state each time ensures a crash is caused solely by that sample, and can be reproduced during triage.

shouc: I don't think there is a mechanism here that allows resetting. Do you think it is reasonable to just restart runtime each time?

shouc: I checked QJS and JSC's instrumentations. Both of them do not restart/reset the runtime.

WilliamParks: I'm looking at jsc.cpp, with the patch applied. The do-while loop is put just above here: https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/jsc.cpp#L3388, and thus creates a new global object each time here: https://github.com/WebKit/WebKit/blob/main/Source/JavaScriptCore/jsc.cpp#L3398

I could be wrong, but that's my understanding.

shouc: Okay you are right. If there is not any way to clean up the context (@dulinriley is that true?), then we can recreate the runtime every time. I don't see too much overhead so it should be fine...

dulinriley: Currently there's no way to clean up the context, so yes recreating the runtime on each step is the right thing to do. Initializing a Hermes runtime takes < 15 ms on an x64 desktop I believe, so the cost should be minimal

shouc: Done

dulinriley:
Sorry about the delay, I'm trying to commit this now. Hopefully it'll be in master soon

PR #446, opened 2021-01-28
Windows Logging and removing dependency on ICU and Shell32

mganandraj:
## Summary

This change removes dependency on ICU.dll and Shell32.dll when running on Windows. It ensures that Hermes works on all Windows devices, including Xbox. Also, this change adds ETW based logging.

In order to remove dependency on ICU, this change adds an implementation of globalization stubs using Windows.Globalization (or Windows NLS) functions. Please note that this change is not related for ECMA402 implementation.

In order to avoid load dependency on Shell32, we are stubbing out some code in the LLVM source.

## Test Plan
Ensured that the Hermes dll build and loads on windows without loading icu.dll and shell32.dll.


lib/Platform/Logging.cpp
```diff
@@ -20,13 +24,32 @@
 
 namespace hermes {
 
+#if defined(_MSC_VER)
+// Define the GUID to use in TraceLoggingProviderRegister
+// {0x5bf36bca, 0x6f36, 0x44a7, {0x83, 0x6f, 0x4f, 0x41, 0xad, 0x75, 0x86,
+// 0x32}}
```
dulinriley, line 30: What's the source of this GUID? Is it something meaningful?
Does it have some format like 123456abc-1234-1234-1234567 that's more human readable and copyable?

lib/Platform/Logging.cpp
```diff
@@ -36,12 +59,25 @@ void hermesLog(const char *componentName, const char *fmt, ...) {
...
       OS_LOG_TYPE_INFO,
       "%{public}s: %{public}s",
       componentName,
       buffer.get());
+#elif defined(_MSC_VER)
+  static std::once_flag g_initEtw_once;
+  std::call_once(
+      g_initEtw_once, []() { TraceLoggingRegister(g_hTraceLoggingProvider); });
+
+  TraceLoggingWrite(
```
dulinriley, line 75: If `TraceLoggingWrite` has a version that takes varargs like `vfprintf` below, you can save some malloc calls in the `va_copy` path here.
If not, this is fine.

utils/tracing/win/Start-Tracing.ps1
```diff
@@ -0,0 +1,28 @@
+param(
```
dulinriley, line 1: Can you add
```
Set-StrictMode -Version Latest
```
to all of these powershell files?

utils/tracing/win/Start-Tracing.ps1
```diff
@@ -0,0 +1,28 @@
+param(
+   [bool]$IncludeGeneralTrace = $False
```
dulinriley, line 2: I'm not a pwsh expert, but can you add some more info to these?
```
[Parameter(ParameterSetName='Default', Position=0, ValueFromPipeline=$true, ValueFromPipelineByPropertyName=$true)]
    [bool]
    ${IncludeGeneralTrace},
```
Not exactly those, but let's add all the validation and help we can to make it easy to use.
That's one of the advantages over bash!

utils/tracing/win/Start-Tracing.ps1
```diff
@@ -0,0 +1,28 @@
+param(
+   [bool]$IncludeGeneralTrace = $False
+)
+
```
dulinriley, line 4: Can you use the `begin`, `process`, `end` paradigm? That might even remove the need for a separate `Stop-Tracing.ps1`

utils/tracing/win/Start-Tracing.ps1
```diff
@@ -0,0 +1,28 @@
+param(
+   [bool]$IncludeGeneralTrace = $False
+)
+
+$scriptPath = split-path -parent $MyInvocation.MyCommand.Definition
```
dulinriley, line 5:
```suggestion
$scriptPath = Split-Path -parent $MyInvocation.MyCommand.Definition
```

utils/tracing/win/Stop-Tracing.ps1
```diff
@@ -0,0 +1,44 @@
+param(
```
dulinriley, line 1: All the same style comments from Start-Tracing.ps1 apply here

dulinriley (commented):
I'm just commenting on `lib/Platform/Logging.cpp` because I wrote that originally,
and the powershell files because I like powershell.

external/llvh/lib/Support/Windows/Path.inc
```diff
@@ -1182,6 +1182,7 @@ void closeFile(file_t &F) {
 }
 
 std::error_code remove_directories(const Twine &path, bool IgnoreErrors) {
+#if !defined(WINUWP)
```
tmikov, line 1185: This looks OK, there is only a small detail. All changes to LLVH should be recorded in the `external/llvh/patches/` directory. Basically:
```
cd external/llvh
git-diff --relative > patches/winup.patch
```

lib/Platform/Logging.cpp
mganandraj: Hi Dulin .. This is a random generated GUID, which globally identifies our "Facebook.hermes" ETW provider uniquely. The same GUID should be provided in the WPR profile (utils\tracing\win\hermes.wprp) to be used when starting and stopping the session using WPR .. 

My understanding is that, this format must be used when setting GUID variable in C/C++ source as the structure looks like, (There are some macros to work with registry format, declspect format etc. but they are more blackbox for me .. )

[REF: Windows Kits\10\Include\10.0.19041.0\shared\guiddef.h]
typedef struct _GUID {
  unsigned long  Data1;
  unsigned short Data2;
  unsigned short Data3;
  unsigned char  Data4[8];
} GUID;

lib/Platform/Logging.cpp
mganandraj: TraceLoggingWrite takes varargs .. but what we need is a version of TraceLoggingString that takes varargs, which i can't find so far :(

utils/tracing/win/Start-Tracing.ps1
mganandraj: Me either :) Did some refactorings .. Let me know your thoughts on the new iteration ..

tmikov (commented):
winuwp.patch appears to be in 16-bit Unicode. Can you get it as plain text please?

mganandraj:
> winuwp.patch appears to be in 16-bit Unicode. Can you get it as plain text please?

Sure! I could smell s

> winuwp.patch appears to be in 16-bit Unicode. Can you get it as plain text please?

Sure .. Sorry .. I didn't notice that earlier !

mganandraj:
> @mhorowitz has imported this pull request. If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D26131827).

Hi @mhorowitz The details on what tests/linter errors are currently failing is not available to me ! Please let me know if there is anything i can do to unblock this PR.

mhorowitz:
> Hi @mhorowitz The details on what tests/linter errors are currently failing is not available to me ! Please let me know if there is anything i can do to unblock this PR.

The problems were internal.  I believe I've fixed them.  I'll confirm on Monday.

lib/Platform/Unicode/PlatformUnicodeWinGlob.cpp
```diff
@@ -0,0 +1,449 @@
...
+        LOCALE_NAME_USER_DEFAULT,
+        dwFlags,
+        reinterpret_cast<LPCWSTR>(buf.data()),
+        buf.size(),
+        reinterpret_cast<LPWSTR>(output.data()),
+        output.size(),
+        nullptr /* lpVersionInformation */,
+        nullptr /* lpReserved */,
+        0 /* sortHandle */);
+    assert (charsWritten == 0);
```
mhorowitz, line 423: After resolving our internal errors, I started seeing failures in the xplat/hermes/test/BCGen/HBC/deltamode/string-functions-update.js test:

    Assertion failed: charsWritten == 0, file xplat\hermes\lib\Platform\Unicode\PlatformUnicodeWinGlob.cpp, line 428

The [documentation for LCMapStringEx](https://docs.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-lcmapstringex) says "This function returns 0 if it does not succeed." so I think this assert may be backwards.  I'm going to try flipping it.  If you think this fix is wrong, or I should do something else, please let me know.

You should be able to run these tests in your hermes checkout with `ninja check-hermes` or similar.

mganandraj: Hi @mhorowitz 

Oh right .. we should assert (charsWritten > 0) ..

I'm able to run the tests now, Thanks !

Should I send another iteration with this change ?

mhorowitz: I made the change myself, but there were some test failures.  I've pasted them here: https://gist.github.com/mhorowitz/ef8e11c780213654d9ad2d8b54460737
I hope you'll see them, too, because they will need to be fixed.  You can include the change above with other fixes.  Thanks!

mganandraj: Marc, I've updated the code and test (This is controversial.. Looking forward for your feedback :)) ..

There were some silly bugs in the code ... Always great to see automated tests catching those !

mhorowitz:
Still seeing test failures: https://gist.github.com/mhorowitz/5932cc020d3e873485085f5cb8ac5150

mganandraj:
> Still seeing test failures: https://gist.github.com/mhorowitz/5932cc020d3e873485085f5cb8ac5150



> Still seeing test failures: https://gist.github.com/mhorowitz/5932cc020d3e873485085f5cb8ac5150

Ah .. I didn't notice these checks .. (I was wrongly assuming the failures were preexisting :( )

// CHECK-NEXT: ß
print('A\u180e\u03a3\u180e'.toLocaleLowerCase() === 'a\u180e\u03c2\u180e');
// CHECK-NEXT: true
print('A\u180e\u03a3\u180eB'.toLocaleLowerCase() === 'a\u180e\u03c3\u180eb');
// CHECK-NEXT: true

As commented in the native test, Windows always capitalizes \u03a3 to \u03c3 .. But ICU seems to be doing context specific mappings here .. Not sure which one is correct, or is one more correct than other ..

I think, ICU is more uptodate, and it is being increasingly adopted as standard these days .

Could you advise how can we make progress here ? 

In the worst case, we have to keep this code in our windows specific fork of Hermes !

mhorowitz:
The existing test code passes on Hermes+ICU, Node, JSC, and Spidermonkey.
I am not a Greek scholar (😂) but Wikipedia says " The letter sigma ⟨Σ⟩ has two different lowercase forms in its standard variant, ⟨σ⟩ and ⟨ς⟩, with ⟨ς⟩ being used in word-final position and ⟨σ⟩ elsewhere."  Given the tests above, it looks like the different behavior is correct in the Greek language.  You cited CaseFolding.txt in your code comments, but there's also https://github.com/unicode-org/icu/blob/3ca5d8ca1559fb52d08bad61e350f0285f3edb37/icu4c/source/data/unidata/SpecialCasing.txt#L211 which adds an additional rule for final sigma.  If there's no way to get this behavior right in Windows, then we'll have to come up with a way to make the tests platform-dependent.

mhorowitz:
I had a realization.  Before this diff, Windows was using ICU for Unicode, and this was giving correct behavior.  Your summary implies that not all Windows device support ICU, hence the desire to switch to Windows.Globalization instead.  But, that makes this diff a regression on some machines.

Would it be reasonable to continue using ICU on Windows platforms which support it?  The upside is better spec compatibility when available, and no regression.  The downside is different behavior on different Windows machines.  I'm not sure which I prefer.

tmikov:
It feels like different behavior on different Windows machines is a recipe for trouble. If there are only a few such cases, can we just handle them as special exceptions in the new code? BTW, I wonder what the behavior of Chakra is?

mhorowitz:
Different behavior on Mac and Windows sounds like a recipe for trouble, too.  If different means "Laptop and Xbox", do most people even know Xbox runs a Windows variant?

As for a special exception, I don't know if it's plausible.  You can see what I said above about the one case we know about: "The letter sigma ⟨Σ⟩ has two different lowercase forms in its standard variant, ⟨σ⟩ and ⟨ς⟩, with ⟨ς⟩ being used in word-final position and ⟨σ⟩ elsewhere."  For us to special case this would require us to know what the end of a word looks like.  This seems awfully brittle.

That said, I suspect that getting this wrong on some platforms would not cause very much trouble.  I imagine Greeks must be frustrated, but used to it by now.

tmikov:
For now I will assume that detecting and using ICU if available, and otherwise automatically falling back to slightly less accurate implementation, is out of the question. Using ICU dynamically will affect all platforms and even doing it only on Windows is not going to be pretty.

Instead, I believe we need to think about two build flavors. The question is which one is the default and how do we deal with the test errors in the non-ICU one.

It is a given that people that want the non-ICU build because they know what they are doing, can always get it.

mganandraj:
Some variance is inevitable as we use different versions of ICU when using the binaries/data shipped with OS, as we do on mobile. But, things will eventually converge as the time progresses. 

But, convergence may not happen with WinGlob as Windows is also increasingly favoring ICU AFAIK. Chakra's original Intl implemenation was based on WinGlob, but they switched to an ICU based implementation recently.

Theoretically, we want to have a primary and runtime fallback implemenation, but that will add unnecessary complexity and performance issues. I agree that we should consider an alternative build flavor with WinGlob for special consumers such as Xbox. And an ICU based implementation should be default.

I'll pull back this PR for now.

Thanks!

Issue #326, opened 2020-08-10
Standalone hermes artifact


nlgtuankiet:
### Context
As a developer, I want to update small parts of my native app (Android + iOS) logic without update the app. The idea is to write these logics/functions in JS then use a runtime to evaluate the result
```java
SomeRuntime.evaluate("1 + 1"); // return "2"
```

### Exiting solution problem
on the Android side, there are some libraries/runtimes that support the use case, but they all have some limitations.
Duktape: lightweight (size), functional but performance is not that great
https://github.com/square/duktape-android
J2V8: offer the best performance, but heavy in size (8mb in split APK)
https://github.com/eclipsesource/J2V8
and some others but not worth to mention

Performace matter because this "function" is invoked for about 100 times per second in a single background thread.
Note: my app is 80% native and 20% react, so reuse the Hermes engine is a good option

### Solution
Make Hermes become a standalone library, support both apps that use React and app that doesn't.

tmikov:
Hi, Hermes already is a standalone library - although its primary use is in ReactNative apps, it has no dependencies on React Native itself.  `libhermes.so` simply exports a JSI interface, which is loaded and used by React Native, and can similarly be loaded and used by any app. 

With that said, our focus currently is React Native, so we don't have a lot of documentation for non-RN use cases. 

For some mode details, please see this comment: https://github.com/facebook/hermes/issues/43#issuecomment-512558215

nlgtuankiet:
for some reason I couldn't find any reference to `libhermes.so` in the react native sdk. The only reference I found is 
```groovy
task packageReactNdkLibs(dependsOn: buildReactNdkLib, type: Copy) {
    from("$buildDir/react-ndk/all")
    into("$buildDir/react-ndk/exported")
    exclude("**/libjsc.so")
    exclude("**/libhermes.so")
}
```

Could you please show me how to load the `libhermes.so` or obtain an instance of the JSI interface?

tmikov:
Take a look here: https://github.com/facebook/react-native/tree/b8621f5d303442ab78dc5d745cfc86a941d4737c/ReactAndroid/src/main/java/com/facebook/hermes/reactexecutor

Let me know if that helps.

nlgtuankiet:
[...]

PR #454, opened 2021-02-10
Cmake project integration

JoshMarler:
## Summary

This PR resolves a set of cmake issues I've hit whilst integrating libhermes into a C++ project.
A discussion of the issues is available here: https://github.com/facebook/hermes/issues/326

## Test Plan

Build using the standard workflow below working:

```
hermes/utils/build/configure.py
cd build
ninja
```

Integrating libhermes in C++ project working here: https://github.com/nick-thompson/blueprint/pull/227/files

### DIFF
--- a/API/hermes/CMakeLists.txt
+++ b/API/hermes/CMakeLists.txt
@@ -82,6 +82,7 @@ hermes_link_icu(libhermes)
 
 # Export the required header directory
 target_include_directories(libhermes PUBLIC ..)
+target_include_directories(libhermes PUBLIC ../../public/)
 
 if("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang" OR
     "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
--- a/API/hermes/CMakeLists.txt
+++ b/API/hermes/CMakeLists.txt
@@ -84,6 +84,11 @@ hermes_link_icu(libhermes)
 target_include_directories(libhermes PUBLIC ..)
 target_include_directories(libhermes PUBLIC ../../public/)
 
+target_compile_definitions(libhermes
+  PRIVATE
+  CREATE_SHARED_LIBRARY=1
+)
+
 if("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang" OR
     "${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU")
   # Export all API symbols
--- a/API/jsi/jsi/CMakeLists.txt
+++ b/API/jsi/jsi/CMakeLists.txt
@@ -6,10 +6,15 @@
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
-add_library(jsi
-        jsi.cpp)
+add_library(jsi STATIC jsi.cpp)
 
 target_include_directories(jsi PUBLIC ..)
+
+set_target_properties(
+  jsi
+  PROPERTIES
+  POSITION_INDEPENDENT_CODE ON
+)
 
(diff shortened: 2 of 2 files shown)


API/hermes/CMakeLists.txt
```diff
@@ -84,6 +84,11 @@ hermes_link_icu(libhermes)
 target_include_directories(libhermes PUBLIC ..)
 target_include_directories(libhermes PUBLIC ../../public/)
 
+target_compile_definitions(libhermes
+  PRIVATE
+  CREATE_SHARED_LIBRARY=1
```
tmikov, line 89: I am trying to figure out why this is necessary. Aren't we already setting this earlier on line 58?

tmikov: Oh, I get it. But how did it work before?

JoshMarler: I believe it only applies to the `hermesapi` target at the moment. 

In our app we're simply using `target_link_libraries(app libhermes)` I'm not sure if this is the expected usage? 

It appears as though `hermesapi` is the static library equivalent of `libhermes`?

neildhar: @tmikov I believe it currently works because we set `-fvisibility=default` on line 95 below, so we end up exporting everything (I don't know about the Windows case though).

JoshMarler: I think that flag is clang/gcc only? Adding the above triggers the `dllexport` for hermes header symbols. It's defined for the `hermesapi` target but not `libhermes`. Anyone able to confirm whether `hermesapi` is intended to basically be the static equivalent to the `libhermes` target?

tmikov: Yeah, that's how `hermesapi` is used internally. It pulls in all dependencies. IIRC, originally `libhermes` pulled in `hermesapi` as a library (which is how it should be in theory), but we ran into problems and so we just duplicated the same source files and dependencies in `libhermes`.

JoshMarler: Excellent thanks for the confirm. That saves me a fair amount of hassle.

JoshMarler: I suspect that's the root cause of the windows symbol export issues then. Perhaps a thing that was missed when linking libhermes against hermesapi was removed.

JoshMarler:
Out of interest did something stop this PR from being merged?

JoshMarler:
@tmikov , is there any chance this is going in? Would be cool to remove the divergence from our fork at some point. Just looking to bump to latest Hermes release for all the Hades goodness and having diverged brings in some small annoyance.

tmikov:
I don't remember what is holding it up, I will take a look.

JoshMarler:
FWIW I've just come back to a project using Hermes and the issue with having to manually add the below header rather than it being part of the publicly exposed headers for the two library targets still exists. 

```
target_include_directories(MyTarget PUBLIC hermes/public/)
```

neildhar:
Hi @JoshMarler, thanks for sharing that. I believe https://github.com/facebook/hermes/pull/694 should fix your issue, but it would be great if you can confirm that patching in the changes makes things work without needing to explicitly include the public directory.

JoshMarler:
Hey @neildhar, Thanks for the update! Apologies for the delays. Will test this shortly and feed back.

JoshMarler:
@neildhar can confirm your change does indeed resolve the issue. Can I ask when the next release branch/tag with this change is likely to land? 

Thanks again for picking this up!

neildhar:
@JoshMarler Thanks for verifying. the change is now merged, since you're building from source, you could start using it immediately.

PR #472, opened 2021-03-17
Remove any usage of FASTCOMP by default in builds

dulinriley:
Summary:
Since fastcomp was deprecated in emscripten v2 and above,
change the default in most places (configure.py and CMake) to be
"upstream".
This shouldn't have any effect on existing builds because they all
manually specified upstream anyways.
This change will prevent accidentally using fastcomp in the future,
although it's still possible to use it if you downgrade to emscripten v1.*

Reviewed By: neildhar

Differential Revision: D27128367


dulinriley:
Just to test the CI

Issue #468, opened 2021-03-14
hermes.framework is incompatible with iOS Simulator on Apple M1 Macs


Arkkeeper:
## Bug Description

There are such lines in **utils/build-ios-framework.sh**:

`build_apple_framework "iphoneos" "armv7;armv7s;arm64" "$ios_deployment_target"`
`build_apple_framework "iphonesimulator" "x86_64;i386" "$ios_deployment_target"`
`create_universal_framework "iphoneos" "iphonesimulator"`

But on M1 Macs, arm64 architecture should be used both for **"iphoneos"** and for **"iphonesimulator"**, otherwise build will fail and RN won't compile with Hermes enabled.

If I get Hermes source code and change those lines to:

`#build_apple_framework "iphoneos" "armv7;armv7s;arm64" "$ios_deployment_target"`
`build_apple_framework "iphonesimulator" "x86_64;arm64" "$ios_deployment_target"`
`#create_universal_framework "iphoneos" "iphonesimulator"`

and then run **./build-ios-framework.sh**, it compiles appropriate version of framework. Putting that hermes.framework in _Pods/hermes-engine/destroot/Library/Frameworks/iphoneos_ allows Xcode to build and run in iOS Simulator successfully.

Unfortunately, it isn't possible to compile universal framework containing arm64 architecture both for iphoneos and for iphonesimulator, because fat frameworks can't include the same architecture twice. The only workaround I have tested is creating xcframework with:

`xcodebuild -create-xcframework -framework ...path-to-iphoneos-hermes.framework -framework ...path-to-iphonesimulator-hermes.framework -output ~/hermes.xcframework`

The amount of Apple Silicon Mac users is growing rapidly, so I'm sure that this issue is rather important.

- [x] I have run `gradle clean` and confirmed this bug does not occur with JSC

Hermes version: 0.7.2
React Native version (if any): 0.64
OS version (if any): macOS 11.2.3 (Apple M1)
Platform (most likely one of arm64-v8a, armeabi-v7a, x86, x86_64): arm64

## Steps To Reproduce

1. Set `:hermes_enabled => true` in Podfile
2. Build React Native project for iOS Simulator
3. Build fails with: 

`building for iOS Simulator, but linking in dylib built for iOS, file '/Users/arkkeeper/git/projecto_mobile/ios/Pods/hermes-engine/destroot/Library/Frameworks/iphoneos/hermes.framework/hermes' for architecture arm64`

## The Expected Behavior

[...]

PR #475, opened 2021-03-20
arm64 simulators & catalyst support

### LINKED ISSUE #460: Support for Mac Catalyst
## Bug Description

React Native 0.64-rc has introduced Hermes for iOS. I'm trying to build for Mac Catalyst and there seems to be some problem with prebuilt binary shipped with Hermes for iOS.

```
ld: building for Mac Catalyst, but linking in dylib built for iOS Simulator, file '/Users/rying/myapp/ios/Pods/hermes-engine/destroot/Library/Frameworks/iphoneos/hermes.framework/hermes' for architecture x86_64
```

<del>- [ ] I have run `gradle clean` and confirmed this bug does not occur with JSC</del>

Hermes version: 0.7.2
React Native version (if any): 0.64.0-rc.3
OS version (if any): macOS 11.2.1
Platform (most likely one of arm64-v8a, armeabi-v7a, x86, x86_64): x86_64 (Catalyst)

## Steps To Reproduce

1. Init RN with 0.64.0-rc.3
2. Build for Mac Catalyst target

## The Expected Behavior

I see Hermes already supports x86-64 so it should play nice with Catalyst too. 

Thanks for the great work and I really hope it gets fixed soon!

Arkkeeper:
## Summary

Edits build scripts to allow merging architectures into .xcframework instead of using lipo for building fat framework.
Adds compatibility with arm64 iOS Simulators on Apple M1 Macs and allows using Hermes with RN apps built for Mac Catalyst.

Fixes #460 and #468

### DIFF
--- a/hermes-engine.podspec
+++ b/hermes-engine.podspec
@@ -20,7 +20,7 @@ Pod::Spec.new do |spec|
   # The podspec would be serialized to JSON and people will download prebuilt binaries instead of the source.
   # TODO(use the hash field as a validation mechanism when the process is stable)
   spec.source      = ENV['hermes-artifact-url'] ? { http: ENV['hermes-artifact-url'] } : { git: "https://github.com/facebook/hermes.git", tag: "v#{spec.version}" }
-  spec.platforms   = { :osx => "11.0", :ios => "10.0" }
+  spec.platforms   = { :osx => "10.13", :ios => "10.0" }
 
   spec.preserve_paths      = ["destroot/bin/*"].concat(HermesHelper::BUILD_TYPE == :debug ? ["**/*.{h,c,cpp}"] : [])
   spec.source_files        = "destroot/include/**/*.h"
--- a/hermes-engine.podspec
+++ b/hermes-engine.podspec
@@ -20,13 +20,13 @@ Pod::Spec.new do |spec|
   # The podspec would be serialized to JSON and people will download prebuilt binaries instead of the source.
   # TODO(use the hash field as a validation mechanism when the process is stable)
   spec.source      = ENV['hermes-artifact-url'] ? { http: ENV['hermes-artifact-url'] } : { git: "https://github.com/facebook/hermes.git", tag: "v#{spec.version}" }
-  spec.platforms   = { :osx => "10.13", :ios => "10.0" }
+  spec.platforms   = { :osx => "11.0", :ios => "10.0" }
 
   spec.preserve_paths      = ["destroot/bin/*"].concat(HermesHelper::BUILD_TYPE == :debug ? ["**/*.{h,c,cpp}"] : [])
   spec.source_files        = "destroot/include/**/*.h"
   spec.header_mappings_dir = "destroot/include"
 
-  spec.ios.vendored_frameworks = "destroot/Library/Frameworks/iphoneos/hermes.framework"
+  spec.ios.vendored_frameworks = "destroot/Library/Frameworks/iphoneos/hermes.xcframework"
   spec.osx.vendored_frameworks = "destroot/Library/Frameworks/macosx/hermes.framework"
 
   spec.xcconfig            = { "CLANG_CXX_LANGUAGE_STANDARD" => "c++14", "CLANG_CXX_LIBRARY" => "compiler-default", "GCC_PREPROCESSOR_DEFINITIONS" => "HERMES_ENABLE_DEBUGGER=1" }
--- a/utils/build-apple-framework.sh
+++ b/utils/build-apple-framework.sh
@@ -46,8 +46,6 @@ function build_host_hermesc {
 # Utility function to configure an Apple framework
 function configure_apple_framework {
   local build_cli_tools enable_bitcode
-  local catalyst="false"
-  local platform=$1
 
   if [[ $1 == iphoneos || $1 == catalyst ]]; then
     enable_bitcode="true"
@@ -59,16 +57,11 @@ function configure_apple_framework {
   else
     build_cli_tools="false"
   fi
-  if [[ $1 == catalyst ]]; then
-    catalyst="true"
-    platform="macosx"
-  fi
 
   local cmake_flags=" \
-    -DHERMES_APPLE_TARGET_PLATFORM:STRING=$platform \
+    -DHERMES_APPLE_TARGET_PLATFORM:STRING=$1 \
     -DCMAKE_OSX_ARCHITECTURES:STRING=$2 \
     -DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=$3 \
-    -DHERMES_APPLE_CATALYST:BOOLEAN=$catalyst \
     -DHERMES_ENABLE_DEBUGGER:BOOLEAN=true \
     -DHERMES_ENABLE_LIBFUZZER:BOOLEAN=false \
     -DHERMES_ENABLE_FUZZILLI:BOOLEAN=false \
@@ -100,18 +93,21 @@ function build_apple_framework {
   fi
 }
 
-# Merges compiled frameworks and combines all of
-# the architectures into the single xcframework
+# Accepts an array of frameworks and will place all of
+# the architectures into the first one in the list
 function create_universal_framework {
   cd ./destroot/Library/Frameworks || exit 1
 
-  echo "Creating universal framework for Apple Platforms"
+  local platforms=("$@")
+  local args=""
 
-  xcodebuild -create-xcframework -framework iphoneos/hermes.framework -framework iphonesimulator/hermes.framework -framework macosx/hermes.framework -output iphoneos/hermes.xcframework
+  echo "Creating universal framework for platforms: ${platforms[*]}"
 
-  rm -r iphonesimulator
-  rm -r macosx
-  rm -r iphoneos/hermes.framework
+  for i in "${!platforms[@]}"; do
+    args+="-framework ${platforms[$i]}/hermes.framework "
+  done
+
+  xcodebuild -create-xcframework $args -output "${platforms[0]}/hermes.xcframework"
 
   cd - || exit 1
 }
--- a/utils/build-apple-framework.sh
+++ b/utils/build-apple-framework.sh
@@ -109,5 +109,11 @@ function create_universal_framework {
 
   xcodebuild -create-xcframework $args -output "${platforms[0]}/hermes.xcframework"
 
+  # Once all was linked into a single framework, clean destroot
+  # from unused frameworks
+  for platform in "${@:2}"; do
+    rm -r "$platform"
+  done
+
   cd - || exit 1
 }
--- a/utils/build-apple-framework.sh
+++ b/utils/build-apple-framework.sh
@@ -46,7 +46,10 @@ function build_host_hermesc {
 # Utility function to configure an Apple framework
 function configure_apple_framework {
   local build_cli_tools enable_bitcode
-  if [[ $1 == iphoneos ]]; then
+  local catalyst="false"
+  local platform=$1
+
+  if [[ $1 == iphoneos || $1 == catalyst ]]; then
     enable_bitcode="true"
   else
     enable_bitcode="false"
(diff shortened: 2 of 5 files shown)


Huxpro:
@Arkkeeper Thanks for contributing! While cc-ing @grabbou to view this, I noticed that `test-apple-runtime` in CircleCI is failing. Could you also take care of that? The `test-apple-runtime` is setup at <https://github.com/facebook/hermes/blob/master/.circleci/config.yml#L202>

hermes-engine.podspec
```diff
@@ -20,13 +20,13 @@ Pod::Spec.new do |spec|
   # The podspec would be serialized to JSON and people will download prebuilt binaries instead of the source.
   # TODO(use the hash field as a validation mechanism when the process is stable)
   spec.source      = ENV['hermes-artifact-url'] ? { http: ENV['hermes-artifact-url'] } : { git: "https://github.com/facebook/hermes.git", tag: "v#{spec.version}" }
-  spec.platforms   = { :osx => "10.13", :ios => "10.0" }
+  spec.platforms   = { :osx => "11.0", :ios => "10.0" }
```
Huxpro, line 23: Maybe because of this version bump?

Arkkeeper: I reverted that bump, I don't think that it's essential for Catalyst support.
Now CircleCI fails on Android tests, but my commits have nothing to do with the android part of code.

Huxpro: Yeah do not worry about the Android tests. The android emulator image is somehow missing.

andreialecu:
Friendly ping @grabbou 

For context why this is extremely important:

RN 0.64 now ships with iOS Hermes support, but lack of support for iOS Simulator arm64 arch makes it impossible to work with it.

Additionally, `react-native-reanimated` v2 has been released, which uses JSI/turbo modules. 

RN apps using turbo modules can only be debugged with Flipper's Hermes Debugger.

andreialecu:
@Arkkeeper is there any way to use this PR before it's merged?

Arkkeeper:
You have to patch **react-native/scripts/react_native_pods.rb**.

First, add a variable to check whether it's ARM Mac, like:
**is_arm = `/usr/sbin/sysctl -n hw.optional.arm64 2>&1`.to_i**

Then modify definition of hermes-engine and refer to repo with:

**if is_arm == 1 then
pod 'hermes-engine', :git => 'https://github.com/Arkkeeper/hermes.git'
else
pod 'hermes-engine', '~> 0.7.2'
end**

Make sure that you have cmake and ninja installed with homebrew.

Then run yarn and prepare to wait about 15-20 minutes. Building hermes-engine from sources for several archs is rather tedious.

andreialecu:
Thank you. Some notes regarding your instructions, in case it helps others:

Ran into `fatal: repository 'Arkkeeper/hermes.git' does not exist` so I had to add it as:
`pod 'hermes-engine', :git => 'git@github.com:Arkkeeper/hermes.git'`.

Also, detecting arm64 didn't work in my case because I can only run cocoapods using `arch -x86_64`, due to other issues with it. So I just had to hardcode it without the check.

utils/build-apple-framework.sh
```diff
@@ -46,7 +46,10 @@ function build_host_hermesc {
 # Utility function to configure an Apple framework
 function configure_apple_framework {
   local build_cli_tools enable_bitcode
-  if [[ $1 == iphoneos ]]; then
+  local catalyst="false"
+  local platform=$1
+
+  if [[ $1 == iphoneos || $1 == catalyst ]]; then
```
alloy, line 52: Perhaps also replace further use of `$1` with the `$platform` alias here?

utils/build-apple-framework.sh
```diff
@@ -56,11 +59,16 @@ function configure_apple_framework {
   else
     build_cli_tools="false"
   fi
+  if [[ $1 == catalyst ]]; then
```
alloy, line 62: ditto

utils/build-apple-framework.sh
```diff
@@ -92,28 +100,18 @@ function build_apple_framework {
...
-
-  lipo -create -output "${platforms[0]}" "${platforms[@]}"
+  echo "Creating universal framework for Apple Platforms"
 
-  # Once all was linked into a single framework, clean destroot
-  # from unused frameworks
-  for platform in "${@:2}"; do
-    rm -r "$platform"
-  done
+  xcodebuild -create-xcframework -framework iphoneos/hermes.framework -framework iphonesimulator/hermes.framework -framework macosx/hermes.framework -output iphoneos/hermes.xcframework
```
alloy, line 110: The output should not assume the framework is just for iOS here, it should theoretically also be able to contain the macOS slices, no?

alloy (changes requested):
Awesome, I didn’t know somebody had already done this work! Most importantly we should make sure it still works for macOS too.

utils/build-apple-framework.sh
grabbou: Yes. Also, it should also get the platforms from the `platforms`, like the `lipo` did, without hardcoding them. I just think it's nicer and doesn't require manual `rm` of every folder, like it's currently done on line 112-114.

utils/build-apple-framework.sh
```diff
@@ -92,28 +100,18 @@ function build_apple_framework {
...
 function create_universal_framework {
   cd ./destroot/Library/Frameworks || exit 1
 
-  local platforms=("$@")
-
-  echo "Creating universal framework for platforms: ${platforms[*]}"
-
-  for i in "${!platforms[@]}"; do
-    platforms[$i]="${platforms[$i]}/hermes.framework/hermes"
-  done
```
grabbou, line 106: Anything in particular wrong with this code? Why not replace `"${platforms[$i]}/hermes.framework/hermes"` with `"${platforms[$i]}/hermes.framework"` and keep using it later?

That way, you could use `xcodebuild` in a similar fashion as I used `lipo`, `lipo -create -output "${platforms[0]}" "${platforms[@]}"`.

grabbou: I see the likely reason for replacing this and going with a hardcoded way was that we always build and include `macosx` platform now, with iOS's Catalyst project in mind.

However, there's one detail to keep in mind here that we may wanna take a closer look at. The caching. Right now, since both Catalyst and regular Mac builds share the same platform name "macosx", they will go into the same build folder. In other words, running iOS and Mac builds will produce different contents of "macosx" folder, potentially introducing conflicts and build issues when run together. It will also make it much harder to debug Hermes on both Mac (application) and iOS (Catalyst) at the same time, iOS build will invalidate Mac artifacts.

With that in mind, the cleanest solution would be to produce `build_macosx` and `build_catalyst` build folders as well as update `destroot` contents so that they don't collide each other. 

To do so, we would need to update `configure_apple_framework` (I have pasted a special comment on that few lines earlier) and CMake to handle that case.

utils/build-apple-framework.sh
```diff
@@ -56,11 +59,16 @@ function configure_apple_framework {
   else
     build_cli_tools="false"
   fi
+  if [[ $1 == catalyst ]]; then
+    catalyst="true"
+    platform="macosx"
+  fi
 
   local cmake_flags=" \
-    -DHERMES_APPLE_TARGET_PLATFORM:STRING=$1 \
```
grabbou, line 61: Please keep in mind that `$1` is still referred to on line 83. You probably want this to be replaced. If you don't do this, you are not building "build_macosx", but "build_catalyst" and effectively, always include Mac artifacts in your `xcframework`.

grabbou: **IMPORTANT**: However, if you read my comment on the caching/producing MacOS / Catalyst frameworks at the same time [here](https://github.com/facebook/hermes/pull/475/files/b63d2408861671d5e3eb4f7697647cd04d3501c7#r621143952), this could actually stay that way, since we want a different build folder for Mac and different build folder for Catalyst.

utils/build-apple-framework.sh
grabbou: As far as updating CMake configuration files, you need to go here https://github.com/facebook/hermes/blob/74595df507b2dd38a677c8cb8c7615ab6ccc61d0/API/hermes/CMakeLists.txt#L133 and change `FRAMEWORK DESTINATION Library/Frameworks/${HERMES_APPLE_TARGET_PLATFORM}` to be different when Catalyst is done. 

Since the names don't really matter for `xcodebuild`, I would suggest we produce `macosx` and `macosx_catalyst` folders inside `Library/Frameworks`, to go in line with the `build_macosx` and `build_macosx_catalyst`.

Then, the platform argument in the script would be `macosx` and `macosx_catalyst`. You would only need to tweak it inside `configure_apple_framework` function to set `DHERMES_APPLE_TARGET_PLATFORM` and `DHERMES_APPLE_CATALYST` in both scenarios. 

Once that is done, both `build_apple_framework` and `create_universal_framework` functions would need no special treatment for Catalyst, since "macosx_catalyst" would be a valid folder and would contain valid files.

grabbou: Once this change is done, we gonna have `catalyst` and `macosx` frameworks inside `destroot`. To keep this function universal, the aforementioned generic logic of not assuming any particular platform would need to be reverted.

utils/build-apple-framework.sh
grabbou: I would actually not use `$platform` here. Further explanation on line 62.

utils/build-apple-framework.sh
grabbou: Instead of using `platform`, I'd rephrase it to `HERMES_APPLE_TARGET_PLATFORM`, since that's what we are using this variable for. To change the value of `HERMES_APPLE_TARGET_PLATFORM` from `$1` when it's "catalyst", since that's not compatible with CMake.

However, using that as an argument and platform name is very clever - it gives us separate build folder and cache for both `macosx` and `catalyst`. If you look at the line `83` when we actually run `build` command, the build output is going to be `build_catalyst` or `build_macosx` since we're using `$1`.

If we replaced that to `$platform`, it would overwrite the build of another command.

grabbou:
Thanks for the PR! I left some comments to further optimise the scripts and the way we handle Catalyst, but I'd consider them minor tweaks, since overall, everything seems to be fine.

Arkkeeper:
Thank you for your suggestions! I've pushed some optimizations for the build scripts. I hope that my PR won't delay Hermes 0.8.1 and RN 0.65.

garrylachman:
> Building hermes-engine from sources for several archs is rather tedious.

Building third time in row, over 1 hour each time. not cool :(

umang-simform:
@Arkkeeper  Any timeline on how soon it can get merged?

kelset:
cc @grabbou remember to check this one ;)

utils/build-apple-framework.sh
```diff
@@ -98,22 +99,15 @@ function create_universal_framework {
...
   for i in "${!platforms[@]}"; do
-    platforms[$i]="${platforms[$i]}/hermes.framework/hermes"
-  done
-
-  lipo -create -output "${platforms[0]}" "${platforms[@]}"
-
-  # Once all was linked into a single framework, clean destroot
-  # from unused frameworks
-  for platform in "${@:2}"; do
-    rm -r "$platform"
```
grabbou, line 113: Any particular reason why this has been removed? This should be reverted. We need to clean up `destroot` to only contain the artifacts that are used in production.

This is because CircleCI will package the entire `destroot` folder into a tarball that is distributed along the `npm` package. If we don't clean it up, we are going to ship one `xcframework` and every platform separately as well.

utils/build-apple-framework.sh
```diff
@@ -98,22 +99,15 @@ function create_universal_framework {
...
+    args+="-framework ${platforms[$i]}/hermes.framework "
   done
 
-  lipo -info "${platforms[0]}"
```
grabbou, line 116: Any particular reason why this has been removed? I think this is a good add-on for debugging purposes on the CI.

grabbou (changes requested):
Thank you for addressing my previous comments. The PR looks really good! Rather than two tiny issues, it is good to go.

I will give it a shot and test it tomorrow, and if everything runs smoothly, I will merge it, reverting the deleted lines on top of your PR.

utils/build-apple-framework.sh
Arkkeeper: Because **lipo -info** fails on xcframework with a fatal error: 
can't map input file: iphoneos (Invalid argument)

utils/build-apple-framework.sh
Arkkeeper: OK, I've reverted that removal

utils/build-apple-framework.sh
grabbou: Ok

grabbou:
Code wise, it looks good. However, I just noticed that `test-apple-runtime` is currently failing. 

> Testing failed:
	Undefined symbol: facebook::hermes::makeHermesRuntime(hermes::vm::RuntimeConfig const&)
	Undefined symbol: facebook::hermes::HermesRuntime::debugJavaScript(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, facebook::hermes::HermesRuntime::DebugFlags const&)
	Testing cancelled because the build failed.

TheSavior:
I believe we should be closing this one in favor of https://github.com/facebook/hermes/pull/546. Thanks for the work so far!

PR #487, opened 2021-04-03
Automatically use ccache

neildhar:
Summary: If you have ccache installed, use it automatically.

Differential Revision: D27555687


davidAmiron:
I need my installed ccache to be used automatically!

kraenhansen:
@neildhar Any chance something like this could have a revival?

neildhar:
@kraenhansen It's pretty easy to enable ccache when configuring the build (by passing `-DCMAKE_CXX_COMPILER_LAUNCHER`), so this wasn't really necessary. And given that ccache changes how compilation is done, we wanted to keep it as something that needs to be explicitly enabled.

Do you have a particular use case in mind?

kraenhansen:
The use-case speeding up the build from source when it's driven by a React Native app, where passing the compiler launcher is not as straight forward right now - I might be able to propose that change in React Native instead.

Ccache is also opt-in there, either by passing a flag from the app Podfile or [setting the `USE_CCACHE` environment variable](https://github.com/facebook/react-native/blob/0a567a63cd65c6e1856d6fe5c3656fcaf6d24654/packages/react-native/scripts/react_native_pods.rb#L446) .. perhaps something similar could work for Hermes?

PR #490, opened 2021-04-07
Drop "on Android" in README

mrousavy:
## Summary

The README states that Hermes is an engine for Android, but it also works on iOS now. I believe Windows also supports Hermes, and you can run Hermes on macOS via Catalyst so I can also add those in the README as well.


neildhar:
Hi @mrousavy, thanks for submitting this! Hermes is now running on many platforms, so let's just drop the 'on Android' instead.

mrousavy:
facebook/react-native also states that react-native runs on "iOS, Android, [and other platforms.](https://reactnative.dev/docs/out-of-tree-platforms)". I think it makes more sense to tell people what platforms hermes can run on, rather than to omit it entirely. What do you think about adding the "and other platforms" link?

alloy:
I tend to agree with @neildhar on this. While optimised for React Native, the engine is usable for other purposes, so I don’t think it needs to follow 1:1 wording used for RN.

PR #494, opened 2021-04-10
Fix handing of minutes in time zone offset parsing

davelyon:
## Summary

When a time zone with an offset that includes negative minutes (e.g.
-2:30) is given the `new Date` or `Date.parse`, dates were parsed
incorrectly.

This resolves the issue by correctly negating the minutes offset, and
adds a test case that will catch any future regressions.

## Test Plan

Added a test case that used a -1:30 offset, and ran the test suite. It failed, showing a 1 hour difference in the unix timestamp.

Applied fix, and re-ran with all tests passing.

Before:

```
test/hermes/date-constructor.js:248:16: error: CHECK-NEXT: expected string not found in input
// CHECK-NEXT: 1618021800000
               ^
<stdin>:101:1: note: scanning from here
1618018200000
^
```

After:
```
[11/12] Running the Hermes regression tests
Testing Time: 31.64s
  Expected Passes    : 1353
  Unsupported Tests  : 55
```

### DIFF
--- a/lib/VM/JSLib/DateUtil.cpp
+++ b/lib/VM/JSLib/DateUtil.cpp
@@ -840,6 +840,7 @@ static double parseISODate(StringView u16str) {
       if (!scanInt(it, end, tzm)) {
         return nan;
       }
+      tzm *= sign;
     }
   }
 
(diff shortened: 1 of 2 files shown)


neildhar:
Nice catch, thanks for fixing this!

PR #500, opened 2021-04-23
chore: Docusaurus with webpack 5

slorber:
Use an unreleased version of Docusaurus to test the webpack5 support works for this site, detect potential problems, and perf improvements.

See also https://github.com/facebook/docusaurus/pull/4089


slorber:
ready for review, let me know if anything does not work

doc/CrossCompilation.md
```diff
@@ -39,4 +39,4 @@ This process is currently happened in different places for different platforms:
 
 1. For Android, this happened in `hermes/android/build.gradle`
 2. For Apple platforms, this happend in `hermes/utils/build-apple-framework.sh`
-3. For Emscripten, you can find an example from the `test-emscripte` job from `hermes/.circleci/config.yml`. Also see more details at [Building with Emscripten](../emscripten)
+3. For Emscripten, you can find an example from the `test-emscripte` job from `hermes/.circleci/config.yml`. Also see more details at [Building with Emscripten](./Emscripten.md)
```
Huxpro, line 42: I'm guessing this is better because it works on both github.com and hermes website (with Docusaurus replacing the link to the correct url)?

website/package.json
```diff
@@ -12,12 +12,12 @@
     "clear": "docusaurus clear"
   },
   "dependencies": {
-    "@docusaurus/core": "^2.0.0-alpha.69",
-    "@docusaurus/preset-classic": "^2.0.0-alpha.69",
+    "@docusaurus/core": "2.0.0-alpha.75",
```
Huxpro, line 15: This seems to be really strict. Is taking minor release w/ `^` bad?

Huxpro (commented):
Thanks for doing this! Left some questions so I can better understand these changes :)

doc/CrossCompilation.md
slorber: Docusaurus has a broken links checker now and this URL was reported as broken.

In general, I think it's a bad idea to link using relative URL paths, because it makes the link sensitive to your host provider adding an extra slash at the end of an URL or not (it is not something very std and lead to a ton of reported problems in all Jamstack frameworks)

website/package.json
slorber: Not bad, it's just that after my tests using canary `2.0.0-alpha.hash` somehow using ^ did not make yarn download alpha75 and the lockfile stayed on the canary version (probably because my version name is bad)

I can revert to ^ if you want but keep in mind we force users to have all core package versions aligned. 
We'll add a "docusaurus upgrade" cli anyway so it does not matter much, up to you

doc/CrossCompilation.md
slorber: (and yes, it also works on Github 👍 )

Huxpro: Oh thanks for providing that context. I certainly fought with trailing slash appended by github pages before: https://twitter.com/Huxpro/status/798816417097224193

website/package.json
Huxpro: Thanks. No need to revert. We can always change that if we really needed.

Huxpro (approved):

slorber:
👌 Thanks, happy to have one more site on webpack 5 🤗

PR #504, opened 2021-05-03
Add JSON fuzzer

shouc:
## Summary
The existing fuzzers could not explore individual functions efficiently. I added the fuzzer for JSON implementation in the project. If you would like more of these, I could also push the fuzzers for escape, hbc, etc. These fuzzers are compatible with google's oss-fuzz. 

## Test Plan
This is test.


tools/fuzzers/CMakeLists.txt
```diff
@@ -3,5 +3,17 @@
...
+# Check LPM installation
+if (HERMES_ENABLE_LPMFUZZERS)
+    if (EXISTS ${PROJECT_SOURCE_DIR}/tools/fuzzers/utils/LPM)
+        message(STATUS "LPM already installed")
+        set(FUZZING_UTILS_DIR ${PROJECT_SOURCE_DIR}/tools/fuzzers/utils)
+    else()
+        message(FATAL_ERROR "LPM is not available, run the
+            script in tools/fuzzers/utils/prepare-env.sh to install it")
+    endif()
+endif()
```
dulinriley, line 15: Can you instead use `find_package` from CMake to look for this? That's the best way to look for third-party dependencies. If not found you should just skip adding `fuzzer-json`

CMakeLists.txt
```diff
@@ -168,6 +168,9 @@ set(HERMES_ENABLE_CODE_COVERAGE OFF CACHE BOOL
 set(HERMES_ENABLE_LIBFUZZER OFF CACHE BOOL
   "Enable libfuzzer")
 
+set(HERMES_ENABLE_LPMFUZZERS OFF CACHE BOOL
+        "Enable LPM fuzzers based on libfuzzer")
```
dulinriley, line 172: In this help message can you use the unabbreviated "LPM" name so people can know what it means?
Also, if you use `find_package` like I suggest below you shouldn't need this new variable.

tools/fuzzers/fuzzer-json/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+# LICENSE file in the root directory of this source tree.
+
+if(HERMES_ENABLE_LIBFUZZER AND HERMES_ENABLE_LPMFUZZERS)
+    set(CMAKE_CXX_STANDARD 14)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    set(HERMES_ENABLE_EH ON)
+
+    if (NOT EXISTS ${PROJECT_SOURCE_DIR}/tools/fuzzers/fuzzer-json/json.pb.cc)
+        message(FATAL_ERROR "Proto file is not compiled, run tools/fuzzers/fuzzer-json/compile_proto.sh")
+    endif()
```
dulinriley, line 13: CMake can compile C++ files as part of the build. This shouldn't be necessary as a separate step.

tools/fuzzers/fuzzer-json/fuzzer-json.cpp
```diff
@@ -0,0 +1,102 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include <hermes/hermes.h>
+#include <jsi/jsi.h>
```
dulinriley, line 9: Use quotes here instead
```suggestion
#include "hermes/hermes.h"
#include "jsi/jsi.h"
```

tools/fuzzers/fuzzer-json/fuzzer-json.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "hermes/Parser/JSONParser.h"
+#include "libprotobuf-mutator/src/libfuzzer/libfuzzer_macro.h"
```
dulinriley, line 11: Place a newline between logical groups of includes.
Groups should be ordered as follows:

* Hermes headers
* Third party headers
* Standard library headers

Like so:
```
#include "hermes/hermes.h"
#include "jsi/jsi.h"

#include "libprotobuf-mutator/src/libfuzzer/libfuzzer_macro.h"
#include "json.pb.h"

#include <sstream>
#include <string>
```

dulinriley (changes requested):
To make this easier to maintain, it should use `find_package` in CMake.
This allows someone to download and install libprotobuf-mutator in any way they want (for example, via a package manager like homebrew or apt).

Also, the mutator file should be built as part of the normal CMake build for the Hermes tool. That way there's no need to run a two-step process for the build.

tools/fuzzers/fuzzer-json/CMakeLists.txt
shouc: This is to compile the protobuf files. Do you want me to make it directly into cmake file instead of asking users to run scripts?

dulinriley: Yes if possible we should compile this file as part of the build of the fuzzer tool.
It can be done like so:
```
add_executable(mytool ARGS...)
add_custom_command(
    OUTPUT  json.pb.cc
    COMMAND mytool ${CMAKE_CURRENT_BINARY_DIR}/json.pb.cc
    DEPENDS someInputFile.cpp.in
)
```
You can place the `add_executable` in the "external" directory so that we know it's building a third party tool for Hermes. It should also use `find_package` to avoid being built unless the LPM package is available.
If LPM is distributed with the generator tool available, you won't even need to build it at all

dulinriley: If LPM isn't distributed via `apt`, then I would follow the instructions here:
https://github.com/google/libprotobuf-mutator#quick-start-on-debianubuntu
And complete the "install" step which will put them into the `/usr/local/include` path to find the headers
and libraries.
You can include a link to that in a message if the package isn't found on the system

PR #535, opened 2021-06-17
Remove redundant cast

gengjiawen:
## Summary

Remove redundant cast

## Test Plan

build pass.


gengjiawen:
@avp Thanks for the review. But looks the import is stuck ?

PR #543, opened 2021-06-30
arm64 simulators & catalyst support (#475)

Huxpro:
Summary:
Edits build scripts to allow merging architectures into .xcframework instead of using lipo for building fat framework.
Adds compatibility with arm64 iOS Simulators on Apple M1 Macs and allows using Hermes with RN apps built for Mac Catalyst.

Fixes https://github.com/facebook/hermes/issues/460 and https://github.com/facebook/hermes/issues/468

Pull Request resolved: https://github.com/facebook/hermes/pull/475

Differential Revision: D29465045

Pulled By: Huxpro

### DIFF
--- a/utils/build-ios-framework.sh
+++ b/utils/build-ios-framework.sh
@@ -10,9 +10,10 @@ if [ ! -d destroot/Library/Frameworks/iphoneos/hermes.framework ]; then
     ios_deployment_target=$(get_ios_deployment_target)
 
     build_apple_framework "iphoneos" "armv7;armv7s;arm64" "$ios_deployment_target"
-    build_apple_framework "iphonesimulator" "x86_64;i386" "$ios_deployment_target"
+    build_apple_framework "iphonesimulator" "x86_64;arm64" "$ios_deployment_target"
+    build_apple_framework "catalyst" "x86_64;arm64" "$ios_deployment_target"
 
-    create_universal_framework "iphoneos" "iphonesimulator"
+    create_universal_framework "iphoneos" "iphonesimulator" "catalyst"
 else
     echo "Skipping; Clean \"destroot\" to rebuild".
 fi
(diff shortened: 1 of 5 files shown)


Huxpro:
This is a rebased version of the original PR that we can use to tell whether or not the CI breakages were false negatives due to outdating issue.

utils/build-ios-framework.sh
```diff
@@ -10,9 +10,10 @@ if [ ! -d destroot/Library/Frameworks/iphoneos/hermes.framework ]; then
     ios_deployment_target=$(get_ios_deployment_target)
 
     build_apple_framework "iphoneos" "armv7;armv7s;arm64" "$ios_deployment_target"
-    build_apple_framework "iphonesimulator" "x86_64;i386" "$ios_deployment_target"
+    build_apple_framework "iphonesimulator" "x86_64;arm64" "$ios_deployment_target"
```
grabbou, line 13: BTW, I do think we should brig `i386` back. We still support iOS 10.

Ref: https://github.com/facebook/hermes/pull/363#issuecomment-696626229

grabbou (changes requested):

grabbou:
There are two more things that I would do, before merging this PR, both relatively easy:

1/ Add a new target to our unit tests, that would be Mac Catalyst, and make CI run and test it (similar to what we do for iOS and Mac already)

2/ For the time being, we cannot do the same for Apple Silicon, since it's not supported. Is there someone with Apple M1 to do the proper testing?

grabbou:
Let's merge this one with a failing CI and @titozzz will send a follow up PR renaming the destination folder from `iphoneos` to `universal`.

Context: the build fails when a folder contains both `.framework` and `.xcframework`, which happens right now as we put `.xcframework` into `iphoneos` folder that already has a `.framework` with iPhone slices only. 

https://github.com/CocoaPods/CocoaPods/issues/10759

TheSavior:
I believe we should be closing this one in favor of https://github.com/facebook/hermes/pull/546

PR #563, opened 2021-07-30
Drop 32 bit build on iOS

Titozzz:
armv7 and armv7s are 32 bit and iOS 11 is 64 only.
Dropping to fix the build and as a bonus make it faster

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -271,7 +271,7 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: v3-repo-{{ .Environment.CIRCLE_SHA1 }}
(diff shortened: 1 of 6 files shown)


Huxpro:
Thanks @Titozzz for the awesome work to fix them!

Regarding aligning versions: historically we only bump the patch number on patch branch. It's misaligned because I mistakenly bumped the `hermes-engine.podspec` at https://github.com/facebook/hermes/commit/6f7b4c47e5cb98549ee3d734fc5c0498854dce22. I haven't formed an opinion on whether or not we should revert it to `0.8.0` or bumping everything to `0.8.1` though. What do you think?

Titozzz:
0.8.0 is perfect. Thanks 👍

PR #565, opened 2021-08-01
Use the CircleCI Android orb for the e2e tests

Huxpro:
Summary:
Try to migrate the E2E tests to use Android orb rather than
setting up everything from the scratch (resulting in flaky build).

Differential Revision: D30037087


Huxpro:
See https://github.com/facebook/hermes/pull/617/files

PR #571, opened 2021-08-13
[JSRangeErrorException - "Invalid timezone name!"] Fix Intl.DateTimeFormat bug with normalizing timeZone value

anton-patrushev:
## Info
I have tried to use `hermes-engine` v0.8.0 with `react-native` v0.65-rc4. The main reason was to test [`Intl`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl) API implementation since my RN project  uses Hermes engine (instead of JSC) for executing JS code on Android. I have also using [`date-fns`](https://github.com/date-fns/date-fns) & [`date-fns-tz`](https://github.com/marnusw/date-fns-tz) libraries. The second one relies on `Intl` API. That's the reason why I've tried to use new `hermes-engine` with `Intl` API support.
But when I started to testing it I had realized what something was going wrong. I had noticed what `Intl.DateTimeFormat` is broken. Below is an example which was broken by `Intl.DateTimeFormat` implementation.

That is piece of code which is formats date and applies provided timeZone.
```js
import { format } from "date-fns"
import { utcToZonedTime } from "date-fns-tz"

function formatTimezoneDate(date, timezone) {
  return format(utcToZonedTime(date, timezone), 'MMM do yyyy, h:mm:ss a');
}

const date = new Date("2021-08-03T22:04:46Z");
const timeZone = "America/New_York";

const formattedDate = formatTimezoneDate(date, timeZone);

// actualResult
console.log(formattedDate) // "Aug 3rd 2021, 10:04:46 pm"

// expected Result
// "Aug 3rd 2021, 6:04:46 pm"
```
[`utcToZonedTime`](https://github.com/marnusw/date-fns-tz/blob/9c495defad9b296984f0a213f7d94e81db1116a0/src/utcToZonedTime/index.js#L28) converts provided date into appropriate timeZone and returns newDate.

`utcToZonedTime` under the hoods uses [`tzParsetimeZone`](https://github.com/marnusw/date-fns-tz/blob/cec2ae63fbb7373a785b99e5388c6c1d84dfd7e9/src/_lib/tzParseTimezone/index.js#L14) function which accepts string `timeZone` value and returns `offsetInMilliseconds` number value.
This function call [`isValidTimezoneIANAString`](https://github.com/marnusw/date-fns-tz/blob/cec2ae63fbb7373a785b99e5388c6c1d84dfd7e9/src/_lib/tzParseTimezone/index.js#L128). Here `isValidTimezoneIANAString` function implementation.
```js
function isValidTimezoneIANAString(timeZoneString) {
  try {
    Intl.DateTimeFormat(undefined, {timeZone: timeZoneString});
    return true;
  } catch (error) {
    return false;
  }
}
```
I have added `console.log(error)` to catch statement and have seen what Intl.DateTimeFormat throws the next error:
```js
JSRangeErrorException("Invalid timezone name!")
```js

```
That is why `utcToZonedTime` function always converts date into UTC timezone (`tzParseTimeZone` always returns `0`. value)
I have found only one [place](https://github.com/facebook/hermes/blob/0b02a101f5d824becb73dcff2750e9384dcbc2ae/lib/Platform/Intl/java/com/facebook/hermes/intl/DateTimeFormat.java#L249) in Hermes source code where it can throw `JSRangeErrorException("Invalid timezone name!")`.
```java
    Object timeZone = JSObjects.Get(options, "timeZone");
    if (JSObjects.isUndefined(timeZone)) {
      timeZone = DefaultTimeZone();
    } else {
      String normalizedTimeZone = normalizeTimeZoneName(JSObjects.getJavaString(timeZone));
      if (!isValidTimeZoneName(normalizedTimeZone)) {
        throw new JSRangeErrorException("Invalid timezone name!");
      }
    }
    mTimeZone = timeZone;
```

[Here](https://github.com/facebook/hermes/blob/0b02a101f5d824becb73dcff2750e9384dcbc2ae/lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterAndroid.java#L207) is `isValidTimeZoneName` method implementation under the hood. It is simple as possible. Everything should be ok.
```java
import java.util.TimeZone;

// ...

  @Override
  public boolean isValidTimeZone(String timeZone) {
    return TimeZone.getTimeZone(timeZone).getID().equals(timeZone);
  }
```
But `isValidTimeZone` method receive already messed up timeZone string value. In my case instead of receiving `America/New_York` value it accepts `AMERICA/NEW_YORK`. This happens because of [`normalizeTimeZoneName`](https://github.com/facebook/hermes/blob/0b02a101f5d824becb73dcff2750e9384dcbc2ae/lib/Platform/Intl/java/com/facebook/hermes/intl/DateTimeFormat.java#L382) method. This method transforms timeZone string into uppercase version.
That's why `TimeZone.getTimeZone("AMERICA/NEW_YORK").getID().equals("AMERICA/NEW_YORK")` will always returns false. (`TimeZone.getTimeZone("AMERICA/NEW_YORK").getID()` always returns `"GMT:` string)

## Brief Changes
- fix `normalizeTimeZoneName` method to fit into JDK 1.8 `java.util.TimeZone` specs.
As far as I understand there are some differences between 1.7 & 1.8 `java.util.TimeZone` implementation. JDK 1.7 supply case insensitive TimeZone module, while JDK 1.8 provides case sensitive module for managing TimeZones. That's why calling the next piece of code will have different behaviors.
```java
// on JDK 1.7 it will be "America/New_York", but 1.8 version makes a fallback to "GMT" id.
TimeZone.getTimeZone("AMERICA/NEW_YORK").getID();
```
I have fixed `normalizeTimeZoneName` function to follow [ECMAScripts specs](https://tc39.es/ecma402/#sec-case-sensitivity-and-case-mapping) and make [compatibility](https://stackoverflow.com/a/33742252) with JDK 1.8 and higher versions.
```java
String id1 = normalizeTimeZoneName("America/New_York");
String id2 = normalizeTimeZoneName("AMERICA/NEW_YORK");
String id3 = normalizeTimeZoneName("america/new_york");
String id4 = normalizeTimeZoneName("UtC");
String id5 = normalizeTimeZoneName("gMt+07:00");

System.out.println(id1); // "America/New_York"
System.out.println(id2); // "America/New_York"
System.out.println(id3); // "America/New_York"
System.out.println(id4); // "UTC"
System.out.println(id5); // "GMT+07:00"
        
System.out.println(isValidTimeZoneName(id1)); // true
System.out.println(isValidTimeZoneName(id2)); // true
System.out.println(isValidTimeZoneName(id3)); // true
System.out.println(isValidTimeZoneName(id4)); // true
System.out.println(isValidTimeZoneName(id5)); // true
```
That's how it behaves now. Tested on my End. Sorry for not providing unit tests for it. I wasn't able to find out how your test system is designed.


TheLonelyAstronaut (approved):

mhorowitz:
Thank you for the detailed bug report and the fix!

CircleCI seems to be having issues on Android, but our internal CI system should run tests on this.  If you want to run t hem yourself, take a look at https://github.com/facebook/hermes/blob/main/android/intltest/build.gradle which is where the tests are run from.  There are instructions at the top.  The actual test cases come from https://github.com/tc39/test262

anton-patrushev:
@mhorowitz So what are next steps you propose?
I have tried to run Intl tests for Android. I think I wasn't successful since almost all of them was failed due to some unknown reason. I have implemented one test, which will ensure what `Intl.DateTimeFormat` API works correctly, but I wasn't able to check, is this test works correctly or not. I have committed that test. But anyway, can we merge this PR (maybe by applying some requested changes or whatever)? Or maybe you could run this test on your internal CI?
It will be great if someone could run this test using this fixed version of Hermes (my fork) and version of Hermes from the main repo and compare the result. It will indicates that my fix is really working (actually for me it works in the real app).

anton-patrushev:
@mhorowitz Any updates?

mhorowitz:
Unfortunately, the regression tests didn't behave as I expected internally, either.  They run now, and the regression tests pass.  The test you added in https://github.com/facebook/hermes/pull/571/commits/23616f71bf7a8d61a773efd763c2eea88860c77e did not, though.  I had to tweak it a bit, now I have this:

      rt.evaluateJavaScript(
          new StringBuilder()
              .append("var date = new Date('2021-08-13T14:00:00Z');\n")
              .append("var formattedDate = Intl.DateTimeFormat('en-US', {\n")
              .append("timeZone: 'America/New_York',\n")
              .append("month: 'numeric',\n")
              .append("day: 'numeric',\n")
              .append("hour: 'numeric',\n")
              .append("minute: 'numeric'\n")
              .append("}).format(date);\n")
              .toString());

This passes.  However, I realized this doesn't (I think) test the underlying fundamental problem of case folding the time zone.  If I modify the test to mangle the case:

              .append("timeZone: 'AMERICA/new_YORK',\n")

The output is GMT, not US/Eastern:

    com.facebook.hermes.test.HermesInstrumentationTest > testDateTimeFormat[Pixel_3_API_29(AVD) - 10] FAILED
            org.junit.ComparisonFailure: expected:<'8/14, [10:00 A]M'> but was:<'8/14, [2:00 P]M'>

It's possible I'm misunderstanding the change you've made, but I expected the unusual casing would succeed based on the description in the PR.  What am I missing?

anton-patrushev:
@mhorowitz Did you run these tests on my branch (with modified Java code) or on other branch, which has the same Java code as the main branch (before my fix)?
This important since I believe `AMERICA/new_YORK` should work with my fix too. (Because I have implemented function, which is reliable to transform this random case strings into Java TZ compatible format to get the right TimeZone ID).

I will be back with answer (I have an empty project with manually patched `hermes-engine` executable, which I made locally (thanks to your docs 🤗), to make some kind of manual E2E QA test).
I'm sure it works with `America/New_York`, but don't remember have I tested it with random case or not 🤔.

mhorowitz:
@anton-patrushev I did my testing on 692ef467 which appears to to be the top revision in your PR.  Is there some other revision I should be testing?

anton-patrushev:
@mhorowitz Could I ask you to run this [test case](https://github.com/facebook/hermes/blob/23616f71bf7a8d61a773efd763c2eea88860c77e/android/intltest/java/com/facebook/hermes/test/HermesInstrumentationTest.java#L93) (which you mentioned and tweaked) on the main hermes branch to ensure it failed. It should failed since Android Intl API doesn't work for me from the main branch.
In that case we can assume my fix works indeed, otherwise I should think about other issue root-case.
I will be back with my manual tests results.
Thanks in advance!

anton-patrushev:
@mhorowitz FYI
I have created [repo](https://github.com/anton-patrushev/Hermes-Intl-Example) with Intl API usage example.
And I have provided some [screenshots](https://github.com/anton-patrushev/Hermes-Intl-Example/tree/main/screenshots).
This [one](https://github.com/anton-patrushev/Hermes-Intl-Example/blob/main/screenshots/Android-Hermes-Local-Fix.png?raw=true) uses the build from current branch with my fix.
Android-Hermes-Local-Fix

So you can compare it with the stable `hermes-engine` package. It doesn't work correctly for this example.
Android-Hermes-Stable

But my fix still doesn't support case insensitivity :(
I will try to fix it and will be back with updates.

mhorowitz:
@anton-patrushev Thanks for the screen shots!  This is consistent with what I see.  Your changes prevent Intl from crashing, but do not fix the issues with timezone case sensitivity.  I'll hold off merging this for now while you investigate the fix for that.

andreialecu:
We're also running into this problem which unfortunately makes Hermes Intl unusable for us.

@anton-patrushev unless I'm reading it wrong, I believe the remaining problem is here:

https://github.com/anton-patrushev/hermes/blob/c31fe892e1a40e37276ad7c29142eecb4168586f/lib/Platform/Intl/java/com/facebook/hermes/intl/DateTimeFormat.java#L246-L252

`mTimeZone` should be set to the result of the normalization, otherwise it will remain mangled.

anton-patrushev:
> We're also running into this problem which unfortunately makes Hermes Intl unusable for us.
> 
> @anton-patrushev unless I'm reading it wrong, I believe the remaining problem is here:
> 
> https://github.com/anton-patrushev/hermes/blob/c31fe892e1a40e37276ad7c29142eecb4168586f/lib/Platform/Intl/java/com/facebook/hermes/intl/DateTimeFormat.java#L246-L252
> 
> `mTimeZone` should be set to the result of the normalization, otherwise it will remain mangled.

Thanks for highlighting it @andreialecu!
I will try to investigate it anytime soon. Seems like your assumption is right!
I will be back with answer!

But just to be on the same page.
@mhorowitz My fix not only prevent Intl from crashing with random case timezone name, but it make Intl works correctly with already normalized timezone name (e.g. `America/New_York`). Since as for me, it  throws an error too, even when I use `America/New_York` in the ideal case 🤷‍♂️.

I will be back with final fix to support case insensitivity too!

andreialecu:
I'd like to point out something regarding the implementation in this PR. There are a bunch of timezones which may not pass through the capitalization function properly. Take a look at: https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

Examples: `Etc/GMT+1` or `US/Alaska`

However, I think it's possible to normalize the timezone in a different way. `java.util.TimeZone.getAvailableIDs()` does return an array with available time zone names. So if the same upper-casing function was applied to the items in that array and then compared to the input, there should be a match. 

The original casing of the match would then be the normalized time zone name.

Hopefully this makes sense. :)

anton-patrushev:
Finally, I'm back :)
So, as @andreialecu proposed, I have reimplemented the whole approach for normalizing and validating timeZone, which comes from the JS side into Android Java internals.

If you would have any questions regarding my approach do not hesitate to ask me.
You can check my [repo](https://github.com/anton-patrushev/Hermes-Intl-Example) with the Hermes Intl usage example.
I have updated [screenshots](https://github.com/anton-patrushev/Hermes-Intl-Example/tree/main/screenshots). This one with my final fix on Android.
Image

@mhorowitz Please take a look.

anton-patrushev:
@mhorowitz
Is there any chance it would be merged?

anton-patrushev:
@mhorowitz any updates?

anton-patrushev:
Can someone provide any info on this?
I found it quite critical for Hermes usage in the Android RN apps.
Sorry for tagging @mhorowitz, but you are the only person who was assigned to this PR (and related issue) from the Facebook team.

Can I do something on my end to make it merged?

andreialecu:
Hey @Huxpro, hope you don't mind the ping. This issue is preventing Intl from working in most (I think) real-life apps on Android.

Mind taking a look at the PR, please?

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterAndroid.java
```diff
@@ -204,8 +210,21 @@ else if (needTime)
...
 
   @Override
-  public boolean isValidTimeZone(String timeZone) {
-    return TimeZone.getTimeZone(timeZone).getID().equals(timeZone);
+  public String normalizeValidTimeZone(final String timeZone) throws JSRangeErrorException {
+    Optional<String> normalizedValidTimeZone = allAvailableTimeZones.stream()
+      .filter(new Predicate<String>() {
+        @Override
+        public boolean test(String tz) {
+          return tz.equalsIgnoreCase(timeZone);
```
mhorowitz, line 218: I believe this is incorrect.  https://tc39.es/ecma402/#sec-case-sensitivity-and-case-mapping only allows mapping/folding ascii a-z <-> A-Z.  equalsIgnoreCase is probably applying additional locale-specific comparisons.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterICU.java
```diff
@@ -269,8 +276,21 @@ public void configure(
...
   @RequiresApi(api = Build.VERSION_CODES.N)
   @Override
-  public boolean isValidTimeZone(String timeZone) {
-    return TimeZone.getTimeZone(timeZone).getID().equals(timeZone);
+  public String normalizeValidTimeZone(final String timeZone) throws JSRangeErrorException {
+    Optional<String> normalizedValidTimeZone = allAvailableTimeZones.stream()
+      .filter(new Predicate<String>() {
+        @Override
+        public boolean test(String tz) {
+          return tz.equalsIgnoreCase(timeZone);
```
mhorowitz, line 284: same here

lib/Platform/Intl/java/com/facebook/hermes/intl/DateTimeFormat.java
```diff
@@ -244,9 +244,10 @@ private void initializeDateTimeFormat(List<String> locales, Map<String, Object>
     if (JSObjects.isUndefined(timeZone)) {
       timeZone = DefaultTimeZone();
     } else {
-      String normalizedTimeZone = normalizeTimeZoneName(JSObjects.getJavaString(timeZone));
-      if (!isValidTimeZoneName(normalizedTimeZone)) {
-        throw new JSRangeErrorException("Invalid timezone name!");
+      try {
+        timeZone = mPlatformDateTimeFormatter.normalizeValidTimeZone(timeZone.toString());
+      } catch (JSRangeErrorException error) {
+        throw error;
```
mhorowitz, line 250: you can just remove this try/catch, it doesn't do anything.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterAndroid.java
andreialecu: You just need to bring this back: 

https://github.com/facebook/hermes/pull/571/files#diff-bd6622e75c91b32a4ac2f2a7f9ad2dc1e6c9ffa72c71e78070f3fc57c8d4f4e3L382-L397

Instead of using the built-in case insensitive comparison methods.

anton-patrushev:
Closed that, since #627 had been [merged](https://github.com/facebook/hermes/commit/9238467f1cceead5426bca8ef5c794137d504429) into main.
Thanks to @calebmackdavenport !!!

PR #594, opened 2021-09-11
[Security]Fix ReDoS

ready-research:
## Summary

Fix potential ReDoS
Reported in https://www.huntr.dev/bounties/124b85d7-896a-43bf-9ff6-011f4aec51df

### DIFF
--- a/tools/node-hermes/nodelib/internal/util/inspect.js
+++ b/tools/node-hermes/nodelib/internal/util/inspect.js
@@ -226,7 +226,7 @@ var meta = ['\\x00', '\\x01', '\\x02', '\\x03', '\\x04', '\\x05', '\\x06', '\\x0
 // License: MIT, authors: @sindresorhus, Qix-, arjunmehta and LitoMore
 // Matches all ansi escape code sequences in a string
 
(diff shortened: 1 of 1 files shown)


ljharb:
in node, `util.inspect` is not meant to be used in production code, and thus there's no vulnerability in it - that seems to be the case here too?

tmikov:
This is technically NodeJS code, just slightly adapted to work with Hermes. Here is the NodeJS version: https://github.com/nodejs/node/blob/dccf64ee9d3683c282a53acb0186ce8c97d7ced9/lib/internal/util/inspect.js#L224

I don't really understand the vulnerability nor the fix and the fix itself does appear to be a valid regular expression, it results in the following error:
```
Invalid RegExp: Parenthesized expression not closed
```

ljharb:
@tmikov sure, but the use case of "inspect" is still meant to be for debugging, not in prod, right?

tmikov:
@ljharb I don't remember, but you are probably right. However `node-hermes` itself is not used in production, it is an experiment. Anyway, the question is moot since the fix appears to be incorrect.

PR #610, opened 2021-10-14
Get canonical locales

SarahCoded:
## Summary

 Implements the getCanonicalLocales function on iOS.

## Test Plan

Running test262/test/intl402/Intl passes on 14 of the tests. The error log for the fails are below. Some of these have failed because BCP 47 tag validation is yet to be implemented (no errors are thrown on some invalid locales).
``` EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/Intl/getCanonicalLocales/error-cases.js
	Uncaught Test262Error: Expected a RangeError to be thrown but no exception was thrown at all
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/Intl/getCanonicalLocales/canonicalized-unicode-ext-seq.js
	Uncaught Test262Error: Expected true but got false
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/Intl/getCanonicalLocales/main.js
	Uncaught Test262Error: Expected [th-TH-u-nu] and [th-TH-u-nu-thai] to have the same contents. th-TH-u-nu-thai
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/Intl/getCanonicalLocales/has-property.js
	Uncaught Test262Error: Expected a Test262Error to be thrown but no exception was thrown at all
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/Intl/getCanonicalLocales/Locale-object.js
	Uncaught TypeError: Cannot read property 'prototype' of undefined
	    at global (/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/Locale-object-2d_po3dq.js:290:18)
```
Adding these failed tests to the skiplist now means that running /test262/test/intl402/Intl --test-intl results in all tests passing.
```-----------------------------------
| Total                |       64 |
| Pass                 |       14 |
| Permanently Skipped  |       31 |
```
Custom tests have also been written in get-canonical-locales.js. It tests an empty string, invalid unicode, and valid locales. The valid locales pass apart from (["aam", "zyb", "art-lojban"]) currently, which gives 'aam' instead of 'aas'.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -553,75 +553,29 @@ jobs:
             cmake --build ./embuild --target emhermesc
             EMHERMESC="$PWD/embuild/bin/emhermesc.js" node ./hermes/tools/emhermesc/test.js
   test-e2e:
-    # For now we run E2E tests on MacOS as running an Android emulator on Linux is
-    # not supported with Circle CI:
-    # https://support.circleci.com/hc/en-us/articles/360000028928-Testing-with-Android-emulator-on-CircleCI-2-0
-    # Windows is another option, but I came across a convenient Mac OS example first:
-    # https://gist.github.com/DoguD/58b4b86a5d892130af84074078581b87
-    macos:
-      xcode: 12.4.0
-    environment:
-      - TERM: dumb
-      # Homebrew currently breaks while updating:
-      # https://discuss.circleci.com/t/brew-install-fails-while-updating/32992
-      - HOMEBREW_NO_AUTO_UPDATE: 1
-      - JVM_OPTS: -Xmx3200m
-      - ANDROID_HOME: /usr/local/share/android-sdk
-      - ANDROID_SDK_HOME: /usr/local/share/android-sdk
-      - ANDROID_SDK_ROOT: /usr/local/share/android-sdk
-      - ANDROID_NDK_HOME: /usr/local/share/android-ndk
-      - QEMU_AUDIO_DRV: none
-      - JAVA_HOME: $(/usr/libexec/java_home)
+    executor:
+      name: android/android-machine
+    working_directory: /home/circleci
     steps:
-      - run:
-          name: Setup dependencies
-          command: |
-            mkdir -p /tmp/hermes/output
-            # Node some version greater than 11 is needed by one of the app dependencies
-            curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.1/install.sh | bash
-            NVM_DIR="$HOME/.nvm"
-            . "$NVM_DIR/nvm.sh"
-            echo 'export NVM_DIR="$HOME/.nvm"' >> $BASH_ENV
-            echo '. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
-            nvm install 14
-            nvm use 14
-            # Reinstall the latest version of Yarn.
-            npm install --global yarn
-            brew tap homebrew/cask
-            brew install --cask android-sdk
-            brew install --cask android-ndk
-            echo 'export PATH="/usr/local/share/android-sdk/platform-tools:/usr/local/share/android-sdk/tools/bin:$PATH"' >> $BASH_ENV
-            # At some point, adb ended up in platform-tools/platform-tools. Work around that.
-            echo 'export PATH="/usr/local/share/android-sdk/platform-tools/platform-tools:$PATH"' >> $BASH_ENV
-      - run:
-          name: Print versions
-          command: |
-            node --version
-            yarn --version
-            echo $(/usr/libexec/java_home)
-      - run:
-          name: Install emulator dependencies
-          command: (yes | sdkmanager "platform-tools" "platforms;android-26" "extras;intel;Hardware_Accelerated_Execution_Manager" "build-tools;26.0.0" "system-images;android-26;google_apis;x86" "emulator" --verbose) || true
-      - run:
-          name: Create emulator definition
-          command: |
-            avdmanager create avd -n Pixel_2_API_26 -k "system-images;android-26;google_apis;x86" -g google_apis -d "Nexus 5"
-      - run:
-          name: Run emulator in background
-          command: /usr/local/share/android-sdk/emulator/emulator @Pixel_2_API_26 -memory 2048 -noaudio
-          background: true
-      - checkout
       - attach_workspace:
-          at: /tmp/hermes/input
+          at: /tmp/input
       - run:
-          name: Run E2E test with main Hermes and Release React Native
+          name: Prepare RNTester
           command: |
-            cd test/circleci-e2e
-            ./run.sh --hermes-npm-package /tmp/hermes/input/hermes-engine-v*.tgz
-          # The react-native setup tool takes AGES "Installing CocoaPods dependencies"
-          no_output_timeout: 45m
-      - store_artifacts:
-          path: /tmp/hermes/output/
+            npm install --global yarn
+            git clone https://github.com/facebook/react-native
+            cd react-native
+            yarn install
+            npm install /tmp/input/hermes-engine-v*.tgz
+            echo "console.log('Using Hermes: ' + (global.HermesInternal != null));" >> packages/rn-tester/js/RNTesterApp.android.js
+      - android/start-emulator-and-run-tests:
+          post-emulator-launch-assemble-command: ''
+          run-tests-working-directory: react-native
+          run-logcat: true
+          test-command: >-
+            ./gradlew :packages:rn-tester:android:app:installHermesRelease
+            && adb shell am start com.facebook.react.uiapp/.RNTesterActivity
+            && timeout 30s adb logcat -e "Using Hermes: true" -m 1
 
   test-e2e-intl:
     executor:


lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,155 +4,208 @@
...
  * LICENSE file in the root directory of this source tree.
  */
-
 #include "hermes/Platform/Intl/PlatformIntl.h"
 
 #import <Foundation/Foundation.h>
 
+NSString *u16StringToNSString(std::u16string src) {
+  auto size = src.size();
+  if (size == 0) {
```
neildhar, line 13: Why is it preferable to return nil as opposed to an empty string?

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,155 +4,208 @@
...
+    return nil;
+  }
+  auto cString = (const unichar*)src.c_str();
+  if (cString == nullptr) {
```
neildhar, line 17: `c_str` cannot return nullptr, so this check can be eliminated

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,155 +4,208 @@
...
+  }
+  return [NSString stringWithCharacters:cString length:size];
+}
+
+std::u16string nsStringToU16String(NSString *src) {
+  auto result = std::u16string();
+  auto constexpr divisor = sizeof(char16_t) / sizeof(char);
+  if ([src canBeConvertedToEncoding:NSUTF16StringEncoding]) {
+    auto size = [src lengthOfBytesUsingEncoding:NSUTF16StringEncoding] + 1; // for NULL terminator;
+    auto buffer = (char *)malloc(size);
```
neildhar, line 28: Instead of setting up a temporary buffer, we can resize the string upfront and use that as the buffer directly.

neildhar (commented):
Just some thoughts on the conversion functions, we can discuss them in greater detail

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,79 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  */
-
```
neildhar, line 7: There are a few of these deleted newlines in the PR, please remove them to avoid adding noise to the PR

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,79 @@
...
 #include "hermes/Platform/Intl/PlatformIntl.h"
 
 #import <Foundation/Foundation.h>
 
+NSString *u16StringToNSString(std::u16string src) {
+  auto size = src.size();
+  auto cString = (const unichar *)src.c_str();
```
neildhar, line 13: style nit

```suggestion
  const auto *cString = (const unichar *)src.c_str();
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,79 @@
...
     vm::Runtime *runtime,
     const std::vector<std::u16string> &locales) {
-  return std::vector<std::u16string>{u"fr-FR", u"es-ES"};
-}
-
+  // 1. If locales is undefined, then
+  // Return a new empty List.
+  if (locales.size() == 0) {
+    return std::vector<std::u16string>();
+  }
```
neildhar, line 36: We don't actually need this check, note that the below code will generate an empty vector if `locales` is empty.
```suggestion
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,79 @@
...
+  // 6. Let k be 0.
+  // 7. Repeat, while k < len
+  for (std::u16string locale : locales) {
+    // We don't have steps for 7a. 7b. 7c. i-iv  .. as we only allow string
+    // arrays here.. Smoke validation. Throw RangeError if input locale string
+    // is (1) empty (2) non-ASCII string.
+    if (locale.empty()) {
+      return runtime->raiseRangeError("Incorrect locale information provided");
+    }
+    if (std::find(locale.begin(), locale.end(), '_') != locale.end()) {
```
neildhar, line 57: As discussed, let's remove this for now. `NSLocale` seems to be very forgiving of non-compliant tags, so we'll need to implement a more complete tag parser later (underscores are just one part of that).

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,79 @@
...
+  // https://tc39.es/ecma402/#locale-objects As of now, 'locales' can only be a
+  // string list/array. 'O' is not a string array of locales
+  // 5. Let len be ? ToLength(? Get(O, "length")).
+  // 6. Let k be 0.
+  // 7. Repeat, while k < len
+  for (std::u16string locale : locales) {
+    // We don't have steps for 7a. 7b. 7c. i-iv  .. as we only allow string
+    // arrays here.. Smoke validation. Throw RangeError if input locale string
+    // is (1) empty (2) non-ASCII string.
+    if (locale.empty()) {
```
neildhar, line 54: Let's remove this as well, and add a comment about BCP 47 tag validation being a TODO item.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,79 @@
...
+      return runtime->raiseRangeError("Incorrect locale information provided");
+    }
+    if (std::find(locale.begin(), locale.end(), '_') != locale.end()) {
+      return runtime->raiseRangeError("Incorrect locale information provided");
+    }
+    // 7.c.v & 7.c.vi
```
neildhar, line 60: Include the text of what these steps require us to do

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,79 @@
...
+    auto tempLocale = u16StringToNSString(locale);
+    NSString *tempCanonicalizedTag =
+        [NSLocale canonicalLocaleIdentifierFromString:tempLocale];
+    auto canonicalizedTag = nsStringToU16String(tempCanonicalizedTag);
+    // 7.c.vii
+    if (!canonicalizedTag.empty() &&
```
neildhar, line 66: Similarly, we can leave checking for empty tags to the tag validation code we will add.

```suggestion
    if (
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,79 @@
...
+        std::find(seen.begin(), seen.end(), canonicalizedTag) == seen.end()) {
+      seen.push_back(canonicalizedTag);
```
neildhar, line 68: Since we don't need `canonicalizedTag` any more, we can use `std::move`. You can read more about it [here](https://en.cppreference.com/w/cpp/utility/move).

```suggestion
      seen.push_back(std::move(canonicalizedTag));
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,79 @@
...
+    // arrays here.. Smoke validation. Throw RangeError if input locale string
+    // is (1) empty (2) non-ASCII string.
+    if (locale.empty()) {
+      return runtime->raiseRangeError("Incorrect locale information provided");
+    }
+    if (std::find(locale.begin(), locale.end(), '_') != locale.end()) {
+      return runtime->raiseRangeError("Incorrect locale information provided");
+    }
+    // 7.c.v & 7.c.vi
+    auto tempLocale = u16StringToNSString(locale);
```
neildhar, line 61: Let's give this a more descriptive name instead of tempLocale. Perhaps something like `localeNSString`. Since this is a pointer, we should also follow the convention of using `auto *`

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,79 @@
...
+    NSString *tempCanonicalizedTag =
```
neildhar, line 62: Similarly, we should give this a more descriptive name, something like `canonicalizedTagNSString`.

test/hermes/intl/get-canonical-locales.js
```diff
@@ -0,0 +1,53 @@
...
+print(Intl.getCanonicalLocales(["aam", "zyb", "art-lojban"]));
+// CHECK-NEXT: aas,za,jbo
+
+try {
+  Intl.getCanonicalLocales(["EN_US"]);
+  print("Succeeded");
+} catch (e) {
+  print("Caught", e.name, e.message);
+}
+//CHECK-NEXT: Caught{{.*}}
```
neildhar, line 28: We'll need to omit this test since we're not handling invalid identifiers yet.

test/hermes/intl/get-canonical-locales.js
```diff
@@ -0,0 +1,53 @@
...
+
+try {
+  Intl.getCanonicalLocales([]);
```
neildhar, line 31: If you're expecting the test to succeed, there is no need to wrap it in a try/catch block. Just print the result and validate it like the first few tests written above.

If it does throw an exception, the test will fail anyway.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,84 @@
...
+// for more discussion of locales and CanonicalizeLocaleList.
 vm::CallResult<std::vector<std::u16string>> getCanonicalLocales(
     vm::Runtime *runtime,
     const std::vector<std::u16string> &locales) {
-  return std::vector<std::u16string>{u"fr-FR", u"es-ES"};
+  return canonicalizeLocaleList(runtime, locales);
 }
 
+// 1. Let O be RequireObjectCoercible(this value).
+// 2. Let S be ? ToString(O).
```
neildhar, line 84:
```suggestion
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,84 @@
...
+  // Note:: Some other major input validation occurs closer to VM in
+  // 'normalizeLocales' in JSLib/Intl.cpp
+  // 2. Let seen be a new empty List.
+  std::vector<std::u16string> seen;
+  // 3. If Type(locales) is String or Type(locales) is Object and locales has an
+  // [[InitializedLocale]] internal slot, then
+  // 4. Else
+  // We don't yet support Locale object -
+  // https://tc39.es/ecma402/#locale-objects As of now, 'locales' can only be a
+  // string list/array. 'O' is not a string array of locales
```
neildhar, line 43: Let's also note the point about validation here.
```suggestion
  // string list/array. Validation occurs in normalizeLocaleList, so this function just takes a vector of strings.
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,84 @@
...
+    }
+  }
+  return seen;
+}
+
+// Implementer note: This method corresponds roughly to
+// https://tc39.es/ecma402/#sec-canonicalizelocalelist
+//
+// Also see the implementer notes on DateTimeFormat#DateTimeFormat()
+// for more discussion of locales and CanonicalizeLocaleList.
```
neildhar, line 76: We should now reference `getCanonicalLocales` instead. 

```suggestion
// https://tc39.es/ecma402/#sec-canonicalizelocalelist
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,84 @@
...
+std::u16string nsStringToU16String(NSString *src) {
+  auto size = src.length;
+  std::u16string result;
+  result.resize(size);
+  [src getCharacters:(unichar *)&result[0] range:NSMakeRange(0, size)];
+  return result;
+}
+
 namespace hermes {
 namespace platform_intl {
```
neildhar, line 27: We should wrap these helper functions in an anonymous namespace. You can read more about that [here](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnon-exported-entities).

You can see how helpers are wrapped in `PlatformIntlAndroid.cpp` as an example ([link](https://github.com/facebook/hermes/blob/5be5dcd5799e2a3f7bb507d9233c8548fb9f21e8/lib/Platform/Intl/PlatformIntlAndroid.cpp#L24)).

```suggestion
namespace hermes {
namespace platform_intl {

namespace {
NSString *u16StringToNSString(std::u16string src) {
  auto size = src.size();
  const auto *cString = (const unichar *)src.c_str();
  return [NSString stringWithCharacters:cString length:size];
}

std::u16string nsStringToU16String(NSString *src) {
  auto size = src.length;
  std::u16string result;
  result.resize(size);
  [src getCharacters:(unichar *)&result[0] range:NSMakeRange(0, size)];
  return result;
}
}

```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -4,20 +4,84 @@
...
+    }
+    // 7.c.vi. Let canonicalizedTag be CanonicalizeUnicodeLocaleId(tag).
+    auto *localeNSString = u16StringToNSString(locale);
+    NSString *canonicalizedTagNSString =
+        [NSLocale canonicalLocaleIdentifierFromString:localeNSString];
+    auto canonicalizedTag = nsStringToU16String(canonicalizedTagNSString);
+    // Throw if invalid unicode was previously present in locale
+    if (canonicalizedTag.empty()) {
+      return runtime->raiseRangeError("Incorrect locale information provided");
+    }
```
neildhar, line 62: As discussed, we can do this validation later
```suggestion
```

test/hermes/intl/get-canonical-locales.js
```diff
@@ -0,0 +1,40 @@
...
+//CHECK-NEXT: Caught{{.*}}
+
+// Invalid unicode
+try {
+  Intl.getCanonicalLocales(["\ud800bc"]);
+  print("Succeeded");
+} catch (e) {
+  print("Caught", e.name, e.message);
+}
+//CHECK-NEXT: Caught{{.*}}
```
neildhar, line 40: These should no longer throw an exception right? So the test would fail.

SarahCoded: Yes, both tests now fail, here is their output.
``` 
>> Intl.getCanonicalLocales(["\ud800bc"]);
[ "", [length]: 1 ]
>> Intl.getCanonicalLocales([""]);
[ "", [length]: 1 ]
```

neildhar: In that case, they should be removed, because they will fail when you run `check-hermes`

test/hermes/intl/get-canonical-locales.js
```diff
@@ -0,0 +1,40 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// RUN: %hermes -O -target=HBC %s | %FileCheck --match-full-lines %s
```
neildhar, line 8: Similar to `intl.js` we want to only run this test when Intl is enabled.

```suggestion
// RUN: %hermes -O -target=HBC %s | %FileCheck --match-full-lines %s
// REQUIRES: intl
```

PR #611, opened 2021-10-15
Add test262 Intl skiplist

silva-nick:
## Summary

Creates a skiplist for the test262 Intl tests and a flag to enable/disable them. Also creates a test specifically for getCanonincalLocales of Intl.

## Test Plan

The getCanonincalLocales tests have been run against v8 to ensure conformity.

## Todo

Run on a more stable environment to guarantee testing correctness.


utils/testsuite/testsuite.py
```diff
@@ -920,6 +925,13 @@ def get_arg_parser():
         action="store_true",
         help="Force lazy evaluation",
     )
+    parser.add_argument(
+        "--test-intl",
+        dest="test_intl",
+        default=False,
+        action="store_true",
+        help="Also test skiplisted INTL tests",
```
neildhar, line 933: I think this help message could be a little confusing. 

```suggestion
        help="Run supported Intl tests.",
```

utils/testsuite/testsuite_skiplist.py
```diff
@@ -1900,6 +1900,8 @@
     "cross-realm",
 ]
 
+INTL_SKIP_LIST = []
```
neildhar, line 1903: Let's actually rename this to something more descriptive. Perhaps something like `SUPPORTED_INTL_TESTS`. Calling it a skip list can cause confusion as to why the Intl tests are split across two different skip lists.

utils/testsuite/testsuite_skiplist.py
```diff
@@ -1900,6 +1900,9 @@
     "cross-realm",
 ]
 
+# Tests for Intl features supported by Hermes
+SUPPORTED_INTL_TESTS = []
```
neildhar, line 1904: Sorry for the back and forth. @avp noted that the way these lists are set up, we'll end up maintaining a set of individual Intl tests in the main skip list, and then the complement of that set in the `SUPPORTED_INTL_TESTS` list. That adds more stuff to keep track of.

Instead, we can make this list just a list of Intl tests as follows:

```suggestion
# Tests for Intl features
INTL_TESTS = ["test262/test/intl402"]
```

Then whenever Intl is disabled, all Intl tests are disabled, and whenever Intl is enabled, only the Intl tests not in the main skip list are enabled. As we move through the project, we'll only need to update the main skip list.

silva-nick: Got it. This definitely makes sense, thanks.

PR #616, opened 2021-10-21
Add locale case functions

silva-nick:
## Summary

Implements Intl `toLocaleLowerCase` and `toLocaleUpperCase`. 

## Test Plan

The current implementation passes 1/4 test262 tests for toLowerCaseLocale, and 2/3 tests for toUpperCaseLocale. The tests used were found in test262/test/intl402/String/prototype/.

Due to the NSLocale ...caseStringWithLocale implementation, the final Unicode output for some cases does not match, even though the casing has been altered for specific locale data. Therefore, we have skipped three of the test262 files and added the test262 cases which do pass to the Hermes test suite. Below is a log of the tests that have failed and will not be included in testing. 

```bash
PASS 8/10 on special_casing_Azeri.js
print('I\u0323\u0307'.toLocaleLowerCase(‘az’) === ‘i\u0323’); \u0131\u0323\u0307

print(‘I\uD800\uDDFD\u0307’.toLocaleLowerCase(‘az’) === ‘i\uD800\uDDFD’); \u0131\U000101fd\u0307
```
PASS 8/10 on special_casing_Turkish.js
```bash
print('I\u0323\u0307'.toLocaleLowerCase('tr') === 'i\u0323'); \u0131\u0323\u0307

print('I\uD800\uDDFD\u0307'.toLocaleLowerCase('tr') === 'i\uD800\uDDFD');  \u0131\U000101fd\u0307
```
PASS 15/27 on special_casing_Lithuanian.js
```bash
print('I\uD834\uDD85'.toLocaleLowerCase('lt') === 'i\u0307\uD834\uDD85'); i\U0001d185

print('J\uD834\uDD85'.toLocaleLowerCase('lt') === 'j\u0307\uD834\uDD85'); j\U0001d185

print('\u012E\uD834\uDD85'.toLocaleLowerCase('lt') === '\u012F\u0307\uD834\uDD85'); \u012F\U0001d185

print('I\uD800\uDDFD\u0300'.toLocaleLowerCase('lt') === 'i\u0307\uD800\uDDFD\u0300'); i\U000101fd\u0300

print('J\uD800\uDDFD\u0300'.toLocaleLowerCase('lt') === 'j\u0307\uD800\uDDFD\u0300'); j\U000101fd\u0300

print('\u012E\uD800\uDDFD\u0300'.toLocaleLowerCase('lt') === '\u012F\u0307\uD800\uDDFD\u0300'); \u012F\U000101fd\u0300

print('I\u0325\uD834\uDD85'.toLocaleLowerCase('lt') === 'i\u0307\u0325\uD834\uDD85'); i\u0325\U0001d185

print('J\u0325\uD834\uDD85'.toLocaleLowerCase('lt') === 'j\u0307\u0325\uD834\uDD85'); j\u0325\U0001d185

print('\u012E\u0325\uD834\uDD85'.toLocaleLowerCase('lt') === '\u012F\u0307\u0325\uD834\uDD85'); \u012F\u0325\U0001d185

print('I\uD800\uDDFD\uD834\uDD85'.toLocaleLowerCase('lt') === 'i\u0307\uD800\uDDFD\uD834\uDD85'); i\U000101fd\U0001d185

print('J\uD800\uDDFD\uD834\uDD85'.toLocaleLowerCase('lt') === 'j\u0307\uD800\uDDFD\uD834\uDD85'); j\U000101fd\U0001d185

print('\u012E\uD800\uDDFD\uD834\uDD85'.toLocaleLowerCase('lt') === '\u012F\u0307\uD800\uDDFD\uD834\uDD85'); \u012F\U000101fd\U0001d185
```

**Test summary:**
```bash
| Results              |   FAIL   |
| Total                |        4 |
| Pass                 |        1 |
| Fail                 |        3 |
| Skipped              |        0 |
| Permanently Skipped  |        0 |
| Pass Rate            |   25.00% |
| Execute fail         |        3 |
```

This PR also includes new test coverage for the functions of which it passes 100%.


lib/Platform/Intl/LocaleMatcher.h
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+@interface LocaleMatcher : NSObject
+
++ (NSString *)bestAvailableLocale:(NSString *)locale:(NSArray<NSString *> *)availableLocales;
++ (NSString *)getBestAvailableLocale:(std::vector<std::u16string>)requestedLocales;
```
neildhar, line 11: I think as far as possible, we should implement this in C++ and use C++ `u16string`s. Especially for utility functions like this, there is a good chance we can use a C++ implementation to also replace parts of our Java implementation, which would reduce code duplication across the platforms.

test/hermes/intl/to-locale-lowercase.js
```diff
@@ -0,0 +1,27 @@
...
+// CHECK-LABEL: get locale lowercase test
+
+print(`${'TEST'.toLocaleLowerCase()}`);
+// CHECK-NEXT: test
+
+print(`${'İstanbul'.toLocaleLowerCase('en-US')}`);
+// CHECK-NEXT: i̇stanbul
+
+print(`${'CAFÉ'.toLocaleLowerCase('en-NZ')}`);
+// CHECK-NEXT: café
```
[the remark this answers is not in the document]
neildhar: That's a good idea, perhaps we should reuse the same string with a few different locales to show that providing a different locale actually makes a difference to the output.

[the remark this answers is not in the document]
SarahCoded: I’ve updated the test file to include a cafeteria string that gives 3 different outputs depending on the locale (which all pass with Hermes). Generally, there aren't many locales where the casing will change anything besides Turkish and Lithuanian.

test/hermes/intl/to-locale-lowercase.js
```diff
@@ -0,0 +1,36 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// RUN: %hermes -O -target=HBC %s | %FileCheck --match-full-lines %s
+// REQUIRES: intl
+
+print("get locale lowercase test");
+// CHECK-LABEL: get locale lowercase test
+
+print(`${'TEST'.toLocaleLowerCase()}`);
```
neildhar, line 14: These don't need to be in template literals (should apply to all of them)
```suggestion
print('TEST'.toLocaleLowerCase());
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -68,12 +208,69 @@
...
 vm::CallResult<std::u16string> toLocaleLowerCase(
     vm::Runtime *runtime,
     const std::vector<std::u16string> &locales,
     const std::u16string &str) {
-  return std::u16string(u"lowered");
+  NSString *nsStr = u16StringToNSString(str);
+
+  // 3. Let requestedLocales be ? CanonicalizeLocaleList(locales).
+  vm::CallResult<std::vector<std::u16string>> requestedLocales =
+      getCanonicalLocales(runtime, locales);
```
neildhar, line 221:
```suggestion
      canonicalizeLocaleList(runtime, locales);
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -68,12 +208,69 @@
...
+  // if they support case mapping for additional locales.
+  // 8. Let locale be BestAvailableLocale(availableLocales, noExtensionsLocale).
+  // Convert to C++ array for bestAvailableLocale function
+  static std::vector<std::u16string> availableLocalesVector = getAvailableLocalesVector();
+  llvh::Optional<std::u16string> locale =
+      bestAvailableLocale(availableLocalesVector, noExtensionsLocale);
+  // 9. If locale is undefined, let locale be "und".
+  if (!locale.hasValue()) {
+    locale = u"und";
+  }
```
neildhar, line 254: All of this looks like it will be reusable, at the very least in `toLocaleUpperCase`. Let's factor it out into a function that takes a locale list and returns the locale name as a `u16string`. This does mean that we'll need to extract the default locale as a string and then reconstruct the `NSLocale`, but it keeps the option to potentially reuse this code on other platforms.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -68,12 +208,69 @@
...
+  // es2022, 6.1.4) of the code points of cuList.
+  NSString *L = u16StringToNSString(locale.getValue());
+  // 13. Return L.
+  return nsStringToU16String([nsStr
+      lowercaseStringWithLocale:[[NSLocale alloc] initWithLocaleIdentifier:L]]);
 }
+
+// Implementer note: This method corresponds roughly to
+// https://tc39.es/ecma402/#sup-string.prototype.tolocaleuppercase
 vm::CallResult<std::u16string> toLocaleUpperCase(
```
neildhar, line 274: Is the plan to implement this in this PR as well?

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -24,6 +24,146 @@
...
+        result.extension = result.locale.substr(noExtensionsLocale.length(), locale.length());
+      }
+      // iii. Return result.
+      return result;
+    }
+  }
+  // 3. Let defLocale be DefaultLocale().
+  NSString *defLocale = [[NSLocale currentLocale] localeIdentifier];
+  // 4. Set result.[[locale]] to defLocale.
+  result.locale = nsStringToU16String(defLocale);
```
neildhar, line 138: Let's also factor out the code to get the current default locale identifier as a `u16string`, then there is nothing platform specific about this function.

lib/Platform/Intl/PlatformIntlApple.mm
SarahCoded: Once I moved a portion of toLocaleLowerCase up to a new function, localeListToLocaleString, I wasn’t able to call canonicalizeLocaleList, but I could still call getCanonicalLocales and I’m not sure why. Also, I wondered why it is better to call one over the other, as they do the same thing.

test/hermes/intl/to-locale-lowercase.js
```diff
@@ -0,0 +1,132 @@
...
+print('CAFÉTÉRIA İĮ̀'.toLocaleLowerCase('tr'));
+// CHECK-NEXT: cafétérıa iį̀
+
+print('CAFÉTÉRIA İĮ̀'.toLocaleLowerCase('lt'));
+// CHECK-NEXT: cafétéria i̇į̇̀
+
+print('\U0001f3eb,̆IA\u0307'.toLocaleLowerCase('cs-CZ'));
+// CHECK-NEXT: u0001f3eb,̆iȧ
+
+print(''.toLocaleLowerCase('en-US').length);
```
neildhar, line 35: Let's actually put the tests from test262 in a separate file, and add an explicit comment stating that they are a working subset of the test262 tests.

test/hermes/intl/to-locale-lowercase.js
```diff
@@ -0,0 +1,132 @@
...
+print('\u0130'.toLocaleLowerCase('tr'));
+// CHECK-NEXT: i
+
+print('I\u0307'.toLocaleLowerCase('tr'));
+// CHECK-NEXT: i
+
+print('IA\u0307'.toLocaleLowerCase('tr') === '\u0131a\u0307');
+// CHECK-NEXT: true
+
+print('I\u0300\u0307'.toLocaleLowerCase('tr') === '\u0131\u0300\u0307');
```
neildhar, line 74: We should omit the failing tests entirely (instead of modifying them to work), since they won't pass on other platforms.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -24,6 +24,190 @@
...
 }
+std::vector<std::u16string> getAvailableLocalesVector() {
+    NSArray<NSString *> *availableLocales = [NSLocale availableLocaleIdentifiers];
+    std::vector<std::u16string> availableLocalesVector;
+      for (id object in availableLocales) {
+        std::u16string u16StringFromVector = nsStringToU16String(object);
+        availableLocalesVector.push_back(u16StringFromVector);
+      }
+    return availableLocalesVector;
+}
```
neildhar, line 35: Now that I think about it, it might be cleaner to put the static list of available locales in this function instead, so that any caller of the function will just get a reference to the pre-computed version. We can initialise it using a lambda as follows:

```suggestion
const std::vector<std::u16string>& getAvailableLocalesVector() {
    static const std::vector<std::u16string> *availableLocalesVector = []{
      NSArray<NSString *> *availableLocales = [NSLocale availableLocaleIdentifiers];
      // Intentionally leaked to avoid destruction order problems.
      auto *vec = new std::vector<std::u16string>();
      for (id str in availableLocales)
        vec.push_back(nsStringToU16String(str));
      return vec;
      }();
    return *availableLocalesVector;
```

The above also eliminates the intermediate variable `u16StringFromVector`, which is more efficient because we get a move instead of a copy of the string. It also means that this function can now be used by multiple other functions, without each of them needing to store their own copy. Finally, it avoids issues with static destruction order by just leaking the vector.

lib/Platform/Intl/PlatformIntlApple.mm
neildhar: In C++, the declaration order of functions matters. So you can't call a function that is declared after the current function. In this case, you would need to move `localeListToLocaleString` below `canonicalizeLocaleList `. You can still call `getCanonicalLocales` though, because it is forward declared in `PlatformIntl.h`.

As for why one is preferred, it is just to match the language in the spec.

test/hermes/intl/to-locale-lowercase.js
SarahCoded: I didn't modify any of the tests, as they were copied directly from the test262 file and then ran against Hermes (and removed/added to the test plan above if they failed). All the ones in the file are genuine passes.

neildhar (commented):
This looks good overall, except for some cosmetic/performance concerns. We can discuss more when we meet tomorrow.

Once these comments are addressed, I think this is good to go, barring concerns from @dmitryrykun.

utils/testsuite/testsuite_skiplist.py
```diff
@@ -1812,7 +1812,11 @@
     "test262/test/intl402/NumberFormat/",
     "test262/test/intl402/PluralRules/",
     "test262/test/intl402/Segmenter/",
-    "test262/test/intl402/String/",
+    # "test262/test/intl402/String/",
+    "test262/test/intl402/String/prototype/toLocaleLowerCase/special_casing_Lithuanian.js",
+    "test262/test/intl402/String/prototype/toLocaleLowerCase/special_casing_Azeri.js",
+    "test262/test/intl402/String/prototype/toLocaleLowerCase/special_casing_Turkish.js",
+    "test262/test/intl402/String/prototype/toLocaleUpperCase/special_casing_Lithuanian.js",
```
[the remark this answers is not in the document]
SarahCoded: Not a duplicate, the first tests lowerCase special_casing_Lithuanian.js and the second tests upperCase special_casing_Lithuanian.js. Kind of confusing as they both have the same end file name.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -68,17 +212,96 @@
   return canonicalizeLocaleList(runtime, locales);
 }
 
+llvh::Optional<std::u16string> localeListToLocaleString(vm::Runtime *runtime,
```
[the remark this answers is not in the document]
SarahCoded: So I’ve changed localeListToLocaleString to return just a string, but as a consequence I also had to change bestAvailableLocale to a return string. But I think I made a cancellation, as by setting locale to “und” in bestAvailableLocale, it does not need to be rechecked in localeListToLocaleString.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -68,17 +212,96 @@
...
+  const std::vector<std::u16string> &locales) {
+  // 3. Let requestedLocales be ? CanonicalizeLocaleList(locales).
+  vm::CallResult<std::vector<std::u16string>> requestedLocales =
+  canonicalizeLocaleList(runtime, locales);
+  // getcano doesnt throw so should this be removed?
+  // How should an exception be thrown here if return state is string?
+//  if (LLVM_UNLIKELY(requestedLocales == llvh::ExecutionStatus::EXCEPTION)) {
+//    return llvh::ExecutionStatus::EXCEPTION;
+//  }
```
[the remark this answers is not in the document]
SarahCoded: Xcode says ‘No viable conversion from returned value of type 'hermes::vm::ExecutionStatus' to function return type 'std::u16string' (aka 'basic_string<char16_t>’)’. It also pushes to change ExecutionStatus to llvh::ExecutionStatus, but it still gives the same error of no viable conversion.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -24,6 +24,149 @@
...
+    subtags.emplace_back(s, tagEnd);
+    if (tagEnd == e)
+      break;
+    s = tagEnd + 1;
+  }
+  std::u16string result;
+  size_t size = subtags.size();
+  for (size_t s = 0; s < size;) {
+    result.append(subtags[s]);
+    s++;
```
silva-nick, line 95:
```suggestion
    s++;
    if (s < size) {
      result.append(u"-");
    }
```

test/hermes/intl/test262-to-locale-lowercase.js
```diff
@@ -0,0 +1,109 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
```
neildhar, line 7: Since this is adapted from test262, we should reproduce their copyright notice here.

```suggestion
/* This code is adapted from Test262, which has the following copyright notice:
*
* The << Software identified by reference to the Ecma Standard* ("Software)">>  is protected by copyright and is being 
* made available under the  "BSD License", included below. This Software may be subject to third party rights (rights 
* from parties other than Ecma International), including patent rights, and no licenses under such third party rights 
* are granted under this license even if the third party concerned is a member of Ecma International.  SEE THE ECMA 
* CODE OF CONDUCT IN PATENT MATTERS AVAILABLE AT http://www.ecma-international.org/memento/codeofconduct.htm FOR 
*INFORMATION REGARDING THE LICENSING OF PATENT CLAIMS THAT ARE REQUIRED TO IMPLEMENT ECMA INTERNATIONAL STANDARDS*.
*
* Copyright (C) 2012-2013 Ecma International
* All rights reserved.
*/

```

lib/Platform/Intl/PlatformIntlApple.mm
neildhar: I think we should move this to the start of the loop, and instead test if `s > 0`. As written, this will not correctly handle the case where there is a trailing unicode extension right? I'm thinking that something like `tr-TR-u-banana` will become `tr-TR-`.

test/hermes/string-locale.js
```diff
@@ -6,9 +6,9 @@
  */
 
 /*
-RUN: LC_ALL=en_US.UTF-8 _HERMES_TEST_LOCALE=en_US.UTF-8 %hermes -O -target=HBC %s \
+RUN: LC_ALL=en_US.UTF-8 _HERMES_TEST_LOCALE=en_US %hermes -O -target=HBC %s \
 RUN:            | %FileCheck --match-full-lines -check-prefix EN %s
-RUN: LC_ALL=tr_TR.UTF-8 _HERMES_TEST_LOCALE=tr_TR.UTF-8 %hermes -O -target=HBC %s \
+RUN: LC_ALL=tr_TR.UTF-8 _HERMES_TEST_LOCALE=tr_TR %hermes -O -target=HBC %s \
 RUN:            | %FileCheck --match-full-lines -check-prefix TR %s
 RUN: LC_ALL=UTF-8 _HERMES_TEST_LOCALE=UTF-8 %hermes -O -target=HBC %s \
```
neildhar, line 13: For consistency, since we know this won't do anything, perhaps we should remove the `_HERMES_TEST_LOCALE` from this one.

```suggestion
RUN: LC_ALL=UTF-8 %hermes -O -target=HBC %s \
```

SarahCoded: Done. Looking at it again, I've changed the tags under _HERMES_TEST_LOCALE from _ to -, as when the parser is introduced, it should error throw on underscores.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -68,17 +221,94 @@
...
+  // Convert to C++ array for bestAvailableLocale function
+  const std::vector<std::u16string> availableLocalesVector =
+      getAvailableLocalesVector();
+  llvh::Optional<std::u16string> locale =
+      bestAvailableLocale(availableLocalesVector, noExtensionsLocale);
+  // 9. If locale is undefined, let locale be "und".
+  if (!locale) {
+    locale = u"und";
+  }
+  return locale;
```
neildhar, line 260:
```suggestion
  llvh::Optional<std::u16string> locale =
      bestAvailableLocale(availableLocalesVector, noExtensionsLocale);
  // 9. If locale is undefined, let locale be "und".
  return locale.getValueOr(u"und");
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -68,17 +221,94 @@
   return canonicalizeLocaleList(runtime, locales);
 }
 
+llvh::Optional<std::u16string> localeListToLocaleString(
```
neildhar, line 224: This should now return a CallResult

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -68,17 +221,94 @@
...
+    vm::Runtime *runtime,
+    const std::vector<std::u16string> &locales) {
+  // 3. Let requestedLocales be ? CanonicalizeLocaleList(locales).
+  vm::CallResult<std::vector<std::u16string>> requestedLocales =
+      canonicalizeLocaleList(runtime, locales);
+  if (LLVM_UNLIKELY(requestedLocales == llvh::ExecutionStatus::EXCEPTION)) {
+    //return llvh::ExecutionStatus::EXCEPTION;
```
neildhar, line 231:
```suggestion
  if (LLVM_UNLIKELY(requestedLocales == vm::ExecutionStatus::EXCEPTION)) {
    return vm::ExecutionStatus::EXCEPTION;
```

SarahCoded: It's still inisisting that No viable conversion from returned value of type 'hermes::vm::ExecutionStatus' to function return type 'llvh::Optional<std::u16string>' (aka 'Optional<basic_string<char16_t>>')?

neildhar: See my other comment about changing the return type to `CallResult`, that will make the return type line up correctly.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -68,17 +221,91 @@
...
+  std::u16string noExtensionsLocale =
+      toNoUnicodeExtensionsLocale(requestedLocale);
+
+  // 7. Let availableLocales be a List with language tags that includes the
+  // languages for which the Unicode Character Database contains language
+  // sensitive case mappings. Implementations may add additional language tags
+  // if they support case mapping for additional locales.
+  // 8. Let locale be BestAvailableLocale(availableLocales, noExtensionsLocale).
+  // Convert to C++ array for bestAvailableLocale function
+  const std::vector<std::u16string> availableLocalesVector =
```
RaisinTen, line 252:
```suggestion
  const std::vector<std::u16string>& availableLocalesVector =
```
`getAvailableLocalesVector()` returns the local static vector by const-ref, so why are we creating a new copy here?

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -24,6 +24,159 @@
...
+    // any Unicode locale extension sequences removed.
+    std::u16string noExtensionsLocale = toNoUnicodeExtensionsLocale(locale);
+    // b. Let availableLocale be BestAvailableLocale(availableLocales,
+    // noExtensionsLocale).
+    llvh::Optional<std::u16string> availableLocale =
+        bestAvailableLocale(availableLocales, noExtensionsLocale);
+    // c. If availableLocale is not undefined, then
+    if (availableLocale) {
+      // i. Set result.[[locale]] to availableLocale.
+      result.locale = std::move(*availableLocale);
```
RaisinTen, line 136:
```suggestion
      result.locale = *availableLocale;
```
`*availableLocale` is an rvalue already. Isn't this cast unnecessary?

lib/Platform/Intl/PlatformIntlApple.mm
neildhar: Good catch! Also marking [this](https://github.com/facebook/hermes/pull/616#discussion_r740584433) as unresolved.

lib/Platform/Intl/PlatformIntlApple.mm
neildhar: `*availableLocale` is an lvalue reference, you can see the implementation of `operator*` [here](https://github.com/facebook/hermes/blob/2219f4c5a0c55480088392deab47c000393b0181/external/llvh/include/llvh/ADT/Optional.h#L187). I think you're referring to [this one](https://github.com/facebook/hermes/blob/2219f4c5a0c55480088392deab47c000393b0181/external/llvh/include/llvh/ADT/Optional.h#L196), which returns an rvalue reference if the operator is invoked on an rvalue reference.

Concretely, this allows the below statements to be interchangeable:
```
std::move(*availableLocale)
*std::move(availableLocale)
```
In the first case, we get an lvalue reference from `operator*` on an lvalue and then `std::move` that. In the second case, we get an rvalue reference from `*` because it is invoked on an rvalue (and therefore triggers the rvalue version you're referring to).

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -24,6 +24,159 @@
   [src getCharacters:(unichar *)&result[0] range:NSMakeRange(0, size)];
   return result;
 }
+const std::vector<std::u16string> &getAvailableLocalesVector() {
+  static const std::vector<std::u16string> *availableLocalesVector = [] {
+    NSArray<NSString *> *availableLocales =
```
dmitryrykun, line 29: You can use `auto` keyword for Objective-C types as well.

PR #622, opened 2021-10-29
Add date time format

silva-nick:
## Summary
We aim to implement the ECMA 402 Datetime Format API on IOS.

## Test Plan
We implemented custom test cases, all of which currently pass in the date-time-format.js file. In using the NSDateFormatter API, there are occasional discrepancies we’ve noticed so far listed below.
- NumberingSystem and DayPeriod implementation has not been found
- AM/PM is not multilingual, de-DE should be "03 Uhr" not "3 AM”
- TimeZoneNames shortGeneric and longGeneric return the same values as short and long respectively
- Fractional second digits can only be displayed to 3 milliseconds	
- Currently, the NS canonicalLocaleIdentifierFromString feature has been removed in order to pass ignore-invalid-unicode-ext-values.js and resolved-locale-with-hc-unicode.js from test262. Once the BCP 47 parser has been added to this PR, there will no longer be a need for canonicalLocaleIdentifierFromString, or for replacing ‘_’ with ‘-‘ in parts of the code.
	
```
| Results              |   FAIL   |
| Total                |      164 |
| Pass                 |       64 |
| Fail                 |       17 |
| Skipped              |       82 |
| Pass Rate            |   79.01% |
| Compile fail         |        1 |
| Execute fail         |       16 |
```
The current implementation adds the following Test262 passes
- constructor-options-timeZoneName-valid.js
- constructor-options-timeZoneName-invalid.js
- timezone-invalid.js
- timezone-canonicalized.js
- default-options-object-prototype.js
- supportedLocalesOf/basic.js
- resolvedOptions/ignore-invalid-unicode-ext-values.js
- resolvedOptions/resolved-locale-with-hc-unicode.js
- resolvedOptions/hourCycle-default.js

The current failures are listed below

Formatted year doesn't contain expected year – expected 1970.000000, got January 1970.
- proleptic-gregorian-calendar.js

Throws a typeError instead of a rangeError
- /prototype/formatRangeToParts/date-is-nan-throws.js
- /prototype/formatRangeToParts/date-x-greater-than-y-throws.js
- /prototype/formatRangeToParts/date-is-infinity-throws.js
- /prototype/formatRange/date-is-nan-throws.js
- /prototype/formatRange/date-x-greater-than-y-throws.js
- /prototype/formatRange/date-is-infinity-throws.js

Intl not declared in global, compile fail
- subclassing.js

Uncaught TypeError: Cannot convert undefined value to object
- /prototype/formatRangeToParts/length.js
- /prototype/formatRangeToParts/name.js
- /prototype/formatRange/length.js
- /prototype/formatRange/name.js

Expected ‘function’ not ‘undefined’
- /prototype/formatRangeToParts/prop-desc.js
- /prototype/formatRange/prop-desc.js

yearName has not been implemented, i.e. setting calendar as Japanese might set yearName as ‘xin-chou’
- /prototype/formatToParts/related-year.js

Not in JSON.stringify format
- /prototype/toStringTag/toString.js

Uncaught Test262Error: descriptor value should be Intl.DateTimeFormat
- /prototype/toStringTag/toStringTag.js

The error output of Test262 is below

```
COMPILE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/subclassing.js
	/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/subclassing-792vh05w.js:288:35: warning: the variable "Intl" was not declared in function "global"
	var referenceDateTimeFormat = new Intl.DateTimeFormat(locales);
	                                  ^~~~
	/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/subclassing-792vh05w.js:291:1: error: invalid statement encountered.
	class MyDateTimeFormat extends Intl.DateTimeFormat {
	^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
	/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/subclassing-792vh05w.js:299:18: warning: the variable "MyDateTimeFormat" was not declared in function "global"
	var format = new MyDateTimeFormat(locales);
	                 ^~~~~~~~~~~~~~~~
	Emitted 1 errors. exiting.
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRangeToParts/date-is-nan-throws.js
	Uncaught Test262Error: NaN/date Expected a RangeError but got a TypeError
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRangeToParts/length.js
	Uncaught TypeError: Cannot convert undefined value to object
	    at getOwnPropertyDescriptor (native)
	    at verifyProperty (/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/length-j5v0k0ox.js:181:53)
	    at global (/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/length-j5v0k0ox.js:465:15)
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRangeToParts/name.js
	Uncaught TypeError: Cannot convert undefined value to object
	    at getOwnPropertyDescriptor (native)
	    at verifyProperty (/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/name-t4wzhgup.js:181:53)
	    at global (/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/name-t4wzhgup.js:465:15)
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRangeToParts/prop-desc.js
	Uncaught Test262Error: `typeof Intl.DateTimeFormat.prototype.formatRangeToParts` is `function` Expected SameValue(«undefined», «function») to be true
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRangeToParts/date-x-greater-than-y-throws.js
	Uncaught Test262Error: x > y Expected a RangeError but got a TypeError
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRangeToParts/date-is-infinity-throws.js
	Uncaught Test262Error: +Infinity/date Expected a RangeError but got a TypeError
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatToParts/related-year.js
	Uncaught Test262Error: Expected SameValue(«false», «true») to be true
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/format/proleptic-gregorian-calendar.js
	Uncaught Test262Error: Formatted year doesn't contain expected year – expected 1970.000000, got January 1970. Expected SameValue(«-1», «-1») to be false
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRange/date-is-nan-throws.js
	Uncaught Test262Error: NaN/date Expected a RangeError but got a TypeError
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRange/length.js
	Uncaught TypeError: Cannot convert undefined value to object
	    at getOwnPropertyDescriptor (native)
	    at verifyProperty (/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/length-3c_4rp_7.js:181:53)
	    at global (/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/length-3c_4rp_7.js:465:15)
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRange/name.js
	Uncaught TypeError: Cannot convert undefined value to object
	    at getOwnPropertyDescriptor (native)
	    at verifyProperty (/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/name-burszpp8.js:181:53)
	    at global (/var/folders/cg/lnpgdmcj7nsg2nfwmx52t97r0000gp/T/name-burszpp8.js:465:15)
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRange/prop-desc.js
	Uncaught Test262Error: `typeof Intl.DateTimeFormat.prototype.formatRange` is `function` Expected SameValue(«undefined», «function») to be true
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRange/date-x-greater-than-y-throws.js
	Uncaught Test262Error: x > y Expected a RangeError but got a TypeError
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/formatRange/date-is-infinity-throws.js
	Uncaught Test262Error: +Infinity/date Expected a RangeError but got a TypeError
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/toStringTag/toString.js
	Uncaught Test262Error: Expected SameValue(«[object Object]», «[object Intl.DateTimeFormat]») to be true
EXECUTE_FAILED /Users/mlhfellow/Documents/t/test262/test/intl402/DateTimeFormat/prototype/toStringTag/toStringTag.js
	Uncaught Test262Error: descriptor value should be Intl.DateTimeFormat
```


lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -116,26 +364,489 @@
...
+  static std::vector<std::u16string> minuteVector = {u"Undefined", u"TwoDigit", u"Numeric"};
+  enum class Second : size_t {Undefined = 0, TwoDigit, Numeric};
+  static std::vector<std::u16string> secondVector = {u"Undefined", u"TwoDigit", u"Numeric"};
+  enum class TimeZoneName : size_t {Undefined = 0, Short, Long};
+  static std::vector<std::u16string> timeZoneNameVector = {u"Undefined", u"Short", u"Long"};
+  enum class DateTimeStyle : size_t {Undefined = 0, Full, Long, Medium, Short};
+  static std::vector<std::u16string> dateTimeStyleVector = {u"Undefined", u"Full", u"Long", u"Medium", u"Short"};
+
+  static std::u16string formatMatcherString(const FormatMatcher &formatMatcher) {
+    return formatMatcherVector[static_cast<size_t>(formatMatcher)];
```
neildhar, line 400: The static lookup tables can be moved into the scope of the functions that use them. You should also make them `static` and `constexpr`. For example:
```suggestion
    static constexpr char16_t* formatMatchers[] = {u"Undefined", u"Bestfit", u"Basic"};
    return formatMatchers[static_cast<size_t>(formatMatcher)];
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -81,6 +214,121 @@
   return std::u16string(u"uppered");
 }
 
+// Implementation of
+// https://tc39.es/ecma402/#sec-getoption
+// Funcion split into two depending on which data type is needed
+vm::CallResult<Option> getOptionBool(
```
neildhar, line 220: Since these are split, we can simplify this function:
1. It can return a `CallResult<bool>`.
2. The `type` parameter can be omitted, and we don't need to check it below.
3. The `fallback` parameter should just be a `bool`.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -81,6 +214,121 @@
...
+  // 5. If type is "boolean", then
+  if (type == u"boolean") {
+    if(!value->second.isBool()) {
+      // TODO: in what layer do we call into runtime to throw exceptions
+      //return runtime->raiseRangeError(u"Boolean option expected but not found");
+      return vm::ExecutionStatus::EXCEPTION;
+    }
+  }
+  // 7. If values is not undefined and values does not contain an element equal to value, throw a RangeError exception.
+  if (!values.empty() && llvh::find(values, value->second.getString()) == values.end()) {
```
neildhar, line 243: `value->second` is a bool, you can't call `getString` on it.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -81,6 +214,121 @@
...
+  if (value == options.end()) {
+    return fallback;
+  }
+  // 4. Assert: type is "boolean" or "string".
+  // 5. If type is "boolean", then
+  if (type == u"boolean") {
+    if(!value->second.isBool()) {
+      // TODO: in what layer do we call into runtime to throw exceptions
+      //return runtime->raiseRangeError(u"Boolean option expected but not found");
+      return vm::ExecutionStatus::EXCEPTION;
```
neildhar, line 239: If the spec says you should throw at a certain point, then in general, that's where you should call into runtime. Let's avoid diverging from the spec wherever possible.

neildhar (commented):
Some comments on the first part

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -950,10 +950,14 @@ bool isTimeZoneValid(
...
-  Option timeZone = options.find(u"timeZone")->second.getString();
+  std::u16string timeZone;
   std::u16string timeZoneValue;
+// Check find first to avoid error
+  if (options.find(u"timeZone") != options.end()) {
+    timeZone = options.find(u"timeZone")->second.getString();
+  }
 //  25. If timeZone is undefined, then
-  if (timeZone.getString() == u"") {
+  if (timeZone == u"") {
```
neildhar, line 960: You can merge these two checks.

```suggestion
  auto timeZoneIter = options.find(u"timeZone");
  if (timeZoneIter == options.end()) {
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -227,48 +379,1214 @@ LocaleMatch lookupMatcher(
...
-  return std::vector<std::u16string>{u"en-CA", u"de-DE"};
+  // 1. Let availableLocales be %DateTimeFormat%.[[AvailableLocales]].
+  NSArray<NSString *> *nsAvailableLocales =
+      [NSLocale availableLocaleIdentifiers];
+
+  // 2. Let requestedLocales be ? CanonicalizeLocaleList(locales).
+  vm::CallResult<std::vector<std::u16string>> requestedLocales =
+      getCanonicalLocales(runtime, locales);
+  std::vector<std::u16string> availableLocales =
+      nsStringArrayToU16StringArray(nsAvailableLocales);
```
neildhar, line 667: Can we now use the utility function to get the available locales?

neildhar (commented):
It looks like there are still some open questions on parts of this PR, let's set up some time early next week to talk through this and the BCP47 PR.

lib/Platform/Intl/PlatformIntlApple.mm
SarahCoded: Unfortunately, getAvailableLocalesVector() wasn’t included in the utilities function, it’s still in the PR add locale case functions on line 27.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -227,32 +356,737 @@ LocaleMatch lookupMatcher(
...
+// https://tc39.es/ecma402/#datetimeformat-objects
 struct DateTimeFormat::Impl {
-  std::u16string locale;
+  // Enum declaration for DateTimeFormat internal slots
+  // https://tc39.es/ecma402/#datetimeformat-objects
+  enum class FormatMatcher : size_t {Undefined = 0, Bestfit, Basic};
+  static constexpr const char16_t* formatMatchers[] = {u"Undefined", u"Bestfit", u"Basic"};
+
+  enum class HourCycle : size_t {Undefined = 0, H11, H12, H23, H24};
+  static constexpr const char16_t*  hourCycleVector[] = {u"Undefined", u"H11", u"H12", u"H23", u"H24"};
```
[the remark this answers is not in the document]
SarahCoded: I decided against using enums in the impl struct in the end, as calling ->getString() on Option from getOption() worked fine. Is this OK, or should enums be used really?

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -301,6 +320,138 @@ LocaleMatch lookupMatcher(
...
+    vm::Runtime *runtime) {
+  //  1. Assert: Type(options) is Object.
+  //  2. Let value be ? Get(options, property).
+  auto value = options.find(property);
+  //  3. If value is undefined, return fallback.
+  if (value == options.end()) {
+    return fallback;
+  }
+  if (!value->second.isBool()) {
+    return runtime->raiseTypeError(vm::TwineChar16("Option for property ") + vm::TwineChar16(property.c_str()) + vm::TwineChar16(" is not a bool"));
```
dmitryrykun, line 372: These checks and error messages for getOptionString and getOptionBool must be identical except for the type checked.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -301,6 +320,138 @@ LocaleMatch lookupMatcher(
...
+  auto value = options.find(property);
+  if (value == options.end()) {
+    return fallback;
+  }
+  //  3. Return ? DefaultNumberOption(value, minimum, maximum, fallback).
+  auto defaultNumber = defaultNumberOption(
+      value->second.getNumber(),
+      minimum,
+      maximum,
+      llvh::Optional<uint8_t>(fallback),
```
dmitryrykun, line 417: Isn't `fallback` already Optional?

lib/Platform/Intl/PlatformIntlApple.mm
SarahCoded: Thanks, I’ve amended it so that both error messages are now the same. I don’t think the llvh::find check needs to be used for getOptionBool, as a values vector isn’t passed in?

dmitryrykun: Yeah, you right about llvh::find, I didn't see that values vector is not there.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -337,48 +493,1338 @@ LocaleMatch lookupMatcher(
...
+  // https://github.com/unicode-org/cldr/blob/master/common/bcp47/calendar.xml
+  static std::vector<std::u16string> s_calendarAliasMappings = {
+      u"gregorian", u"gregory"};
+  if (std::find(
+          s_calendarAliasMappings.begin(),
+          s_calendarAliasMappings.end(),
+          value) == s_calendarAliasMappings.end()) {
+    return value;
+  } else {
+    int pos = find(
```
dmitryrykun, line 550: 1. What is the difference in behaviour between `find` and `std::find`?
2. What happens if value is not in s_calendarAliasMappings? Will app crash trying to access `s_calendarAliasMappings.end()`?
3. I suggest extracting lookup procedure into a separate function since it is the same for all three `resolve*Alias` functions.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -337,48 +493,1338 @@ LocaleMatch lookupMatcher(
...
+      u"islamic",
+      u"islamic-umalqura",
+      u"islamic-tbla",
+      u"islamic-civil",
+      u"islamic-rgsa",
+      u"iso8601",
+      u"japanese",
+      u"persian",
+      u"roc"};
+  uint8_t pos = iter - keys.begin();
```
dmitryrykun, line 694: Please use `auto` here.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -113,14 +131,15 @@
 struct LocaleMatch {
   std::u16string locale;
   std::u16string extension;
+  std::u16string matchedLocale;
 };
 LocaleMatch lookupMatcher(
-    const std::vector<std::u16string> &requestedLocales,
+    const vm::CallResult<std::vector<std::u16string>> &requestedLocales,
```
dmitryrykun, line 137: Does it have to be `CallResult`? Feels a little odd that `CallResult` is passed as an input to a function.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -337,48 +493,1338 @@ LocaleMatch lookupMatcher(
...
+// Implementation of
+// https://tc39.es/ecma402/#sec-intl.datetimeformat.supportedlocalesof
 vm::CallResult<std::vector<std::u16string>> DateTimeFormat::supportedLocalesOf(
     vm::Runtime *runtime,
     const std::vector<std::u16string> &locales,
     const Options &options) noexcept {
-  return std::vector<std::u16string>{u"en-CA", u"de-DE"};
+  // 1. Let availableLocales be %DateTimeFormat%.[[AvailableLocales]].
+  // 2. Let requestedLocales be ? CanonicalizeLocaleList(locales).
+  vm::CallResult<std::vector<std::u16string>> requestedLocales =
```
dmitryrykun, line 793: Can use `auto` here as well. And in general it's a good thing to use type inference as much as possible.

lib/Platform/Intl/PlatformIntlApple.mm
SarahCoded: 1. Sorry, that looks like a typo they should both be the same. 
2. I tried to emulate https://github.com/facebook/hermes/blob/c65a5ebb12621ae866501051923cbe397dfaa3f8/lib/Platform/Intl/java/com/facebook/hermes/intl/UnicodeExtensionKeys.java#L110 , but I’m not entirely sure what these AliasMappings functions are trying to achieve as it just returns value, and yes it does feel odd that there isn’t any error checking in the function.

dmitryrykun: Thank you sharing this link.
`s_calendarAliasMappings`  in this Java implementation is an actual map. Its type is `Map<String, String>`. This means that it represents mapping from `u"gregorian"` to `u"gregory"`. `u"gregorian"` is the key, and `u"gregory"` is the value.
Their lookup procedure checks if there is a key in the map, if yes - it returns corresponding value, if not - it returns the key itself.
In order to be able to implement this functionality in C++, you should be using `std::map` instead of `std::vector`.

dmitryrykun: It is also true for `resolveNumberSystemAlias` and `resolveCollationAlias`.

SarahCoded: Ah right thank you that makes sense now. So no error checking is needed as when the value isn’t found in the map, it will just return the value.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -337,48 +493,1297 @@ LocaleMatch lookupMatcher(
...
+    const std::vector<std::u16string> &relevantExtensionKeys
+    //  In line with Java, haven't included LocaleData
+) {
+  //  Skip 1/2/3, as we don't need to have independant implementations for best
+  //  fit
+  auto localeMatchResult = lookupMatcher(requestedLocales, availableLocales);
+  //  5. Let result be a new Record.
+  ResolveLocale result;
+  std::u16string value;
+  std::vector<std::u16string> supportedExtensionAdditionKeys;
```
dmitryrykun, line 686: Also is there a reason why there are no **7.** and **8.** ?

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -337,48 +493,1297 @@ LocaleMatch lookupMatcher(
...
+  //  9. For each element key of relevantExtensionKeys, do
+  for (std::u16string key : relevantExtensionKeys) {
+    result.value = u"";
+    if (localeMatchResult.extension != u"") { // 9.h.
+      if (localeMatchResult.extension.find(key)) { // 9.h.i.
+        std::u16string requestedValue = localeMatchResult.extension;
+        if (requestedValue != u"") {
```
dmitryrykun, line 693: requestedValue must be `keywords[key]`, and `keywords` must be defined in **7.**, which is missing. With current implementation this condition breaks. This check is essentially the one that we had in this line    ` if (localeMatchResult.extension != u"") { // 9.h.`, which means that `else` section is unreachable.
Please double check the correctness of this entire function.

lib/Platform/Intl/PlatformIntlApple.mm
SarahCoded: Sorry, I hadn’t yet tested any of the functionality in resolveLocale as LocaleMatcher wasn’t returning any extensions even when there were some. I had trouble writing this, as I tried to follow the structure of https://github.com/facebook/hermes/blob/1ece7cc6e59183055295829d1ba1c364cca1a805/lib/Platform/Intl/java/com/facebook/hermes/intl/LocaleResolver.java#L20 but it got complicated as the variable naming didn’t match the spec and there weren’t many comments.

lib/Platform/Intl/PlatformIntlApple.mm
SarahCoded: For JSC I noticed they used something like localeMatchResult.extension.split('-', keywords); to set the keywords instead of UnicodeExtensionComponents. Is there a built in function like split that could be used instead of writing out a loop for this?

lib/Platform/Intl/PlatformIntlApple.mm
dmitryrykun: Okay, so the problem is that Java implementation does not follow the specification precisely.
https://tc39.es/ecma402/#sec-resolvelocale 7.a and 7.b say that there have to be components that are populated by this function https://tc39.es/ecma402/#sec-unicode-extension-components. And keywords is a member of components.
Java people decided to do this work in LocaleMatcher https://github.com/facebook/hermes/blob/1ece7cc6e59183055295829d1ba1c364cca1a805/lib/Platform/Intl/java/com/facebook/hermes/intl/LocaleMatcher.java#L59 , and assigned components directly to LocaleMatchResult.extensions

dmitryrykun: I would suggest to implement https://tc39.es/ecma402/#sec-unicode-extension-components, implement **7.**, and stick to spec as much as possible.

lib/Platform/Intl/PlatformIntlApple.mm
dmitryrykun: Unfortunately I can't think of any built-in C++ function that does that.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -337,48 +490,1307 @@ LocaleMatch lookupMatcher(
...
+      hourCycle->getValue() = u"";
+      // 15. Set opt.[[hc]] to hourCycle.
+      opt.hc = hourCycle->getValue();
+    }
+  }
+
+  // 16 - 23
+  static const std::vector<std::u16string> relevantExtensionKeys = {
+      u"ca", u"nu", u"hc"};
+  auto r = resolveLocale(
```
dmitryrykun, line 998: As I understand, according to https://tc39.es/ecma402/#sec-initializedatetimeformat **17.**, `opt` should be used here instead of `options`.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -337,48 +490,1307 @@ LocaleMatch lookupMatcher(
...
+      eraOption,
+      yearOption,
+      monthOption,
+      dayOption,
+      hourOption,
+      minuteOption,
+      secondOption,
+      timeZoneNameOption,
+      dayPeriodOption};
+  // 36. If dateStyle is not undefined or timeStyle is not undefined, then
```
dmitryrykun, line 1168: What about **35.**?

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -337,48 +490,1307 @@ LocaleMatch lookupMatcher(
...
+  // 4. Let matcher be ? GetOption(options, "localeMatcher", "string", «
+  //   "lookup", "best fit" », "best fit").
+  auto matcher = getOptionString(
+      optionsToUse,
+      u"localeMatcher",
+      {u"lookup", u"best fit"},
+      llvh::Optional<std::u16string>(u"best fit"),
+      runtime);
+  // 5. Set opt.[[localeMatcher]] to matcher.
+  opt.localeMatcher = matcher->getValue();
```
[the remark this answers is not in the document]
SarahCoded: I didn't check for hasValue on this one because the fallback is "best fit" (normally it's blank), so in theory it has to have a value. But is it safer to just use hasValue() anyway?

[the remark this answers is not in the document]
dmitryrykun: Sorry, I missed that there is a fallback value, thank yo for the clarification.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -337,48 +490,1338 @@ LocaleMatch lookupMatcher(
...
+      case 3: {
+        [customFormattedDate appendString:@"SSS"];
+      }
+    }
+  }
+  // Not supported - dayPeriod (at night/in the morning)
+  // Set the custom date from all the concatonated NSStrings (locale will
+  // automatically seperate the order) Only set a template format if it isn't
+  // empty
+  if (![customFormattedDate isEqual:@""]) {
```
dmitryrykun, line 1697: It is more common to check string length in Objective-C. So this condition would read like `customFormattedDate.length > 0`

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -337,48 +490,1338 @@ LocaleMatch lookupMatcher(
...
+    return u"second";
+  }
+  if (formatter == 'S') {
+    return u"fractionalSecond";
+  }
+  return u"literal";
+}
+std::u16string removeDuplicatesFromString(const std::u16string &s) {
+  std::u16string arr;
+  std::unordered_map<char, char> exists;
```
dmitryrykun, line 1769: I would be cleaner to use `unordered_set` here.

SarahCoded: Unfortunately it says no template named 'unordered_set' in namespace 'std';.

neildhar: You have to `#include <unordered_set>`

PR #627, opened 2021-10-31
Fix Intl.DateTimeFormat TimeZone

calebmackdavenport:
## Summary

Please feel free to edit anything as necessary.
Picking up where https://github.com/facebook/hermes/pull/571 left off

Error:
intl_error
Cause: https://github.com/facebook/hermes/pull/627#discussion_r739885220

_Coauthored with @anton-patrushev_ 🎉 
_A special thanks to @hcwesson_ 🙏 

## Test Plan

Followed testing instructions provided by @mhorowitz https://github.com/facebook/hermes/pull/571#pullrequestreview-773381009

image


lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterAndroid.java
```diff
@@ -204,8 +211,21 @@ else if (needTime)
   }
 
   @Override
-  public boolean isValidTimeZone(String timeZone) {
-    return TimeZone.getTimeZone(timeZone).getID().equals(timeZone);
```
calebmackdavenport, line 208: Previously, `normalizeTimeZoneName` would convert `"America/New_York"` to `"AMERICA/NEW_YORK"`
However, the problem is now `TimeZone.getTimeZone("AMERICA/NEW_YORK").getID()` => `"GMT"`

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterICU.java
```diff
@@ -269,8 +277,21 @@ public void configure(
...
   @RequiresApi(api = Build.VERSION_CODES.N)
   @Override
-  public boolean isValidTimeZone(String timeZone) {
-    return TimeZone.getTimeZone(timeZone).getID().equals(timeZone);
+  public String normalizeValidTimeZone(final String timeZone) throws JSRangeErrorException {
+    Optional<String> normalizedValidTimeZone = allAvailableTimeZones.stream()
+    .filter(new Predicate<String>() {
+      @Override
+      public boolean test(String tz) {
+        return tz.compareToIgnoreCase(timeZone) == 0;
```
calebmackdavenport, line 285: Use [compareToIgnoreCase](https://docs.oracle.com/javase/7/docs/api/java/lang/String.html#compareToIgnoreCase(java.lang.String)) instead of equalsIgnoreCase to resolve issue described here: https://github.com/facebook/hermes/pull/571#discussion_r723799518

anton-patrushev:
@calebmackdavenport thanks man!

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterAndroid.java
```diff
@@ -204,8 +211,21 @@ else if (needTime)
...
+        return tz.compareToIgnoreCase(timeZone) == 0;
+      }
+    }).findAny();
+
+    if(!normalizedValidTimeZone.isPresent()) {
+      String errorMessage = timeZone + " is invalid timeZone";
+      throw new JSRangeErrorException(errorMessage);
+    }
+
+    return normalizedValidTimeZone.get();
```
neildhar, line 228: I think it would be easier to read and understand this code if it was just a for loop that returns as soon as it finds a match, but perhaps that's a matter of personal preference.

```suggestion
    for (String id : TimeZone.getAvailableIDs()){
      if(id.compareToIgnoreCase(timeZone) == 0){
        return id;
      }
    }
    throw new JSRangeErrorException(timeZone + " is an invalid timeZone");
```

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterAndroid.java
```diff
@@ -204,8 +211,21 @@ else if (needTime)
   }
 
   @Override
-  public boolean isValidTimeZone(String timeZone) {
-    return TimeZone.getTimeZone(timeZone).getID().equals(timeZone);
+  public String normalizeValidTimeZone(final String timeZone) throws JSRangeErrorException {
```
neildhar, line 214: Let's keep this function in `DateTimeFormat.java` so that we don't need to maintain two copies of it.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterAndroid.java
```diff
@@ -15,10 +15,17 @@
...
+import java.util.Optional;
+import java.util.List;
+import java.util.Arrays;
+import java.util.function.Predicate;
+import java.util.regex.Pattern;
 
 public class PlatformDateTimeFormatterAndroid implements IPlatformDateTimeFormatter {
   private DateFormat mDateFormat = null;
 
+  private List<String> allAvailableTimeZones = Arrays.asList(TimeZone.getAvailableIDs());
```
neildhar, line 27: We probably don't need to maintain this list, since we can just call `getAvailableIDs` whenever we need the IDs.

lib/Platform/Intl/java/com/facebook/hermes/intl/DateTimeFormat.java
```diff
@@ -244,9 +245,13 @@ private void initializeDateTimeFormat(List<String> locales, Map<String, Object>
     if (JSObjects.isUndefined(timeZone)) {
       timeZone = DefaultTimeZone();
     } else {
-      String normalizedTimeZone = normalizeTimeZoneName(JSObjects.getJavaString(timeZone));
-      if (!isValidTimeZoneName(normalizedTimeZone)) {
-        throw new JSRangeErrorException("Invalid timezone name!");
+      try {
```
neildhar, line 248: This try-catch block seems unnecessary, since it just rethrows any exception thrown inside it.

lib/Platform/Intl/java/com/facebook/hermes/intl/DateTimeFormat.java
```diff
@@ -244,9 +245,13 @@ private void initializeDateTimeFormat(List<String> locales, Map<String, Object>
...
+        if (!isValidTimeZoneName(timeZone.toString())) {
```
neildhar, line 249: It seems like we don't need to perform this validation any more, since `normalizeTimeZoneName` will already throw an exception for any timezone that is invalid.

neildhar (changes requested):
Thanks for working on this! It seems like the validation is now unnecessary, since the normalisation function effectively also performs validation of the input.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterAndroid.java
calebmackdavenport: This is much cleaner, thank you 👏

lib/Platform/Intl/java/com/facebook/hermes/intl/DateTimeFormat.java
```diff
@@ -244,10 +255,10 @@ private void initializeDateTimeFormat(List<String> locales, Map<String, Object>
     if (JSObjects.isUndefined(timeZone)) {
       timeZone = DefaultTimeZone();
     } else {
-      String normalizedTimeZone = normalizeTimeZoneName(JSObjects.getJavaString(timeZone));
-      if (!isValidTimeZoneName(normalizedTimeZone)) {
+      if (!isValidTimeZoneName(timeZone.toString())) {
         throw new JSRangeErrorException("Invalid timezone name!");
       }
```
neildhar, line 260: As mentioned, we don't need to perform this check any more, since the normalising function will automatically reject anything that isn't a valid timezone anyway.

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterAndroid.java
neildhar: Remember to delete this

lib/Platform/Intl/java/com/facebook/hermes/intl/PlatformDateTimeFormatterICU.java
```diff
@@ -17,10 +17,16 @@
...
+import java.util.Optional;
+import java.util.List;
+import java.util.Arrays;
 
 public class PlatformDateTimeFormatterICU implements IPlatformDateTimeFormatter {
   private DateFormat mDateFormat = null;
 
+  @RequiresApi(api = Build.VERSION_CODES.N)
+  private List<String> allAvailableTimeZones = Arrays.asList(TimeZone.getAvailableIDs());
+
```
neildhar, line 29: Remember to delete this

lib/Platform/Intl/java/com/facebook/hermes/intl/DateTimeFormat.java
```diff
@@ -379,25 +390,8 @@ private void initializeDateTimeFormat(List<String> locales, Map<String, Object>
...
-      }
-    }
-
-    return normalized.toString();
-  }
-
   private boolean isValidTimeZoneName(String timeZone) {
-    return mPlatformDateTimeFormatter.isValidTimeZone(timeZone);
+    return Pattern.compile("^[\\x00-\\x7F]+[a-zA-Z]+$").matcher(timeZone).find();
   }
```
neildhar, line 395: We should be able to eliminate this function altogether now.

lib/Platform/Intl/java/com/facebook/hermes/intl/DateTimeFormat.java
```diff
@@ -138,6 +140,15 @@ private Object ToDateTimeOptions(Object options, String required, String default
     return options;
   }
 
+  public String normalizeValidTimeZone(final String timeZone) throws JSRangeErrorException {
```
neildhar, line 143: Since this now rejects invalid input with an exception, we can just call it `normalizeTimeZone`.

```suggestion
  public String normalizeTimeZone(final String timeZone) throws JSRangeErrorException {
```

lib/Platform/Intl/java/com/facebook/hermes/intl/DateTimeFormat.java
neildhar: Interesting, in the case of comparing timezones, it doesn't explicitly forbid non-ASCII characters, it just says that we have to convert to uppercase before comparing, and when when we compare to uppercase, we should only convert lowercase ASCII characters. That's what the current implementation (before this PR) of `normalizeTimeZoneName` does. So let's follow exactly what the spec says here. Instead of doing validation with a regex, we should convert both the argument and the timezone we are comparing it to to uppercase (only for ASCII characters), and then compare those to determine if there is a match.

Does that sound reasonable?

mhorowitz: Instead, should we just write our own string comparator which only operates on a-z, as described in https://tc39.es/ecma402/#sec-case-sensitivity-and-case-mapping? That seems simpler than having a separate step involving an ad hoc regexp check.

PR #634, opened 2021-11-08
Add utility functions

SarahCoded:
## Summary

In order to complete resolvedLocales in date time format, lookupMatcher is required. As a consequence, lookupMatcher requires toNoUnicodeExtensionsLocale(), bestAvailableLocale() and getDefaultLocale(). 

These four functions have therefore been added to this PR to help break dependancies between functions. They were previously implemented in the add-locale-case-functions PR.
## Test Plan
Three of the functions, toNoUnicodeExtensionsLocale(), bestAvailableLocale() and getDefaultLocale() have been tested within the toLocaleLowerCase/toLocaewUpperCase functions. They have passed the relevant hermes test suite tests - test262-to-locale-lowercase.js, to-locale-lowercase.js and to-locale-uppercase.js.


neildhar (changes requested):

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -24,6 +24,122 @@
...
+      return candidate;
+    }
+
+    // b. Let pos be the character index of the last occurrence of "-" (U+002D)
+    // within candidate.
+    size_t pos = candidate.rfind(u'-');
+
+    // ...If that character does not occur, return undefined.
+    if (pos == std::string::npos) {
+      return u"und";
```
neildhar, line 59: The advantage of using optional here is that we can clearly distinguish `undefined`.
```suggestion
      return llvh::None;
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -24,6 +24,122 @@
...
+    s = tagEnd + 1;
+  }
+  std::u16string result;
+  size_t size = subtags.size();
+  for (size_t s = 0; s < size;) {
+    s++;
+    if (s > 0) {
+      result.append(u"-");
+    }
+    result.append(subtags[s]);
```
neildhar, line 92: We should increment after the element is processed
```suggestion
    if (s > 0) {
      result.append(u"-");
    }
    result.append(subtags[s]);
    s++;
```

neildhar (changes requested):

PR #641, opened 2021-11-19
Fix: Fix fuzzilli compilation error

sdt27:
If you are using the recent hermes source code, you need to comment out the following code in order to compile successfully.
external/llvh/cmake/config-ix.cmake:323
```

# Define LLVM_HAS_ATOMICS if gcc or MSVC atomic builtins are supported.
# include(CheckAtomic)
```


neildhar:
Hi @sdt27, could you share a little more about your setup and the error you're running into? I'm able to successfully build the fuzzilli target on trunk.

PR #643, opened 2021-11-23
Improve the compilation command to avoid the lack of some libraries.

P1umer:
## Summary
Compile with hermes/utils/build/configure.py can avoid the problem of missing libraries.


neildhar:
Hi @P1umer, could you share a little more about what error you're running into when trying to build, and how this change addresses the build failures? In theory, switching from directly invoking `cmake` to using the configuration script should really make a difference.

PR #645, opened 2021-11-24
Add BCP 47 Parser

silva-nick:
## Summary

This PR adds a BCP-47 compliant Unicode language tag parser to Intl. Adding the parser gives the IOS version of Intl correct behavior for ECMA402 [isStructurallyValidLanguageTag](https://tc39.es/ecma402/#sec-isstructurallyvalidlanguagetag) and [canonicalizeUnicodeLocaleId](https://tc39.es/ecma402/#sec-canonicalizeunicodelocaleid).

It also provides a platform independent language tag parser such that the Java implementation (LocaleIdentifier.java) may be removed in the future.

## Test Plan

- This change includes custom unit tests for each of the parsers use cases abstractly based off [test262](https://github.com/tc39/test262/blob/main/test/intl402/language-tags-valid.js)
- It also passes all test 262 getCanonicalLocales tests except for those using the locale object
```bash
| Total                |       38 |
| Pass                 |       17 |
| Permanently Skipped  |        2 |
```
- It also includes a few additional getCanonicalLocales tests in `test/hermes/intl/get-canonical-locales.js`


neildhar (commented):
Some initial feedback, let's set up some time next week to discuss the design of the PR in detial.

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -73,43 +73,50 @@ bool isUnicodeLanguageSubtag(const std::u16string &str, int start, int end) {
   // https://unicode.org/reports/tr35/#Unicode_Language_and_Locale_Identifiers
   // = alpha{2,3} | alpha{5,8};
   // root case?
-  return isCharType(str, start, end, 2, 3, &isASCIILetter) ||
-    isCharType(str, start, end, 5, 8, &isASCIILetter);
+  const llvh::ArrayRef<char16_t> ref(str.data() + start, end - start + 1);
+  return isCharType(ref, 2, 3, &isASCIILetter) ||
+    isCharType(ref, 5, 8, &isASCIILetter);
 }
 bool isUnicodeScriptSubtag(const std::u16string &str, int start, int end) {
```
neildhar, line 80: All of these functions should now take `ArrayRef`, instead of needing to take the underlying string and a range.

lib/Platform/Intl/CMakeLists.txt
```diff
@@ -4,16 +4,16 @@
 # LICENSE file in the root directory of this source tree.
 
 if(HERMES_ENABLE_INTL)
+  add_hermes_library(hermesBCP47Parser STATIC BCP47Parser.cpp)
+  hermes_link_icu(hermesBCP47Parser)
```
neildhar, line 8: No need to link ICU, since the parser doesn't use it

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -26,6 +26,168 @@
...
+  if (!parsedId.transformedLanguageIdentifier.regionSubtag.empty()){
+    transformedExtension += u"-" + parsedId.transformedLanguageIdentifier.regionSubtag;
+  }
+  for (const auto &subtag : parsedId.transformedLanguageIdentifier.variantSubtagList) {
+    transformedExtension += u"-" + subtag;
+  }
+  
+  // Sort tfields by the alphabetical order of its keys
+  std::vector<std::u16string> tFields;
+  for (auto it = parsedId.transformedExtensionFields.begin(); it != parsedId.transformedExtensionFields.end(); it++) {
```
neildhar, line 112: range based for loop

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -26,6 +26,168 @@
...
+  // Any extensions are in alphabetical order by their singleton (eg, en-t-xxx-u-yyy, not en-u-yyy-t-xxx).
+
+  // Sort other extensions by singleton
+  std::vector<std::u16string> otherExtensions;
+  for (auto it = parsedId.otherExtensionMap.begin(); it != parsedId.otherExtensionMap.end(); it++) {
+    std::u16string oExt;
+    oExt += it->first;
+    for (const auto &ext : it->second) {
+      oExt += u"-" + ext;
+    }
```
neildhar, line 77:
```suggestion
  for (const auto &p: parsedId.otherExtensionMap) {
    std::u16string oExt = p.first;
    for (const auto &ext : p.second) {
      oExt += u"-" + ext;
    }
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -26,6 +26,168 @@
 }
 }
 
+// https://unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers
+// https://tc39.es/ecma402/#sec-canonicalizeunicodelocaleid
+vm::CallResult<std::u16string> canonicalizeLocaleId(const std::u16string &localeId, vm::Runtime *runtime) {
+  llvh::Optional<bcp47_parser::ParsedLocaleIdentifier> parserOpt = bcp47_parser::parseLocaleId(localeId);
```
neildhar, line 32: Long type names like this are a good candidate for `auto`, especially when the type is clear based on the right hand side.

```suggestion
  auto parserOpt = bcp47_parser::parseLocaleId(localeId);
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -26,6 +26,168 @@
...
+  if (!parserOpt.hasValue()) {
+    // Not structurally valid language tag
+    return runtime->raiseRangeError(vm::TwineChar16("Unicode locale id ") + vm::TwineChar16(localeId.c_str()) +vm::TwineChar16(" is not structurally valid."));
+  }
+
+  std::u16string canoLocaleId;
+  canoLocaleId.reserve(localeId.length());
+
+  // Casing, variant, and attribute sorting are handled in parsing
+  bcp47_parser::ParsedLocaleIdentifier parsedId = parserOpt.getValue();
```
neildhar, line 43: Move is more efficient here

```suggestion
  bcp47_parser::ParsedLocaleIdentifier parsedId = std::move(*parserOpt);
```

It would be even more efficient to never make a new object at all:
```suggestion
  const auto &parsedId = *parserOpt;
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -26,6 +26,168 @@
...
+
+  // Sort other extensions by singleton
+  std::vector<std::u16string> otherExtensions;
+  for (auto it = parsedId.otherExtensionMap.begin(); it != parsedId.otherExtensionMap.end(); it++) {
+    std::u16string oExt;
+    oExt += it->first;
+    for (const auto &ext : it->second) {
+      oExt += u"-" + ext;
+    }
+    otherExtensions.push_back(u"-" + oExt);
```
neildhar, line 78: For consistency with the other places, perhaps the "-" should be added later, when appending to canoLocaleId?

neildhar (commented):
Some initial stylistic comments

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,560 @@
...
+      }
+    }
+    
+    if (!nextSubtag()) {
+      return true;
+    }
+  }
+
+  
+  return false;
```
neildhar, line 254: This code is not reachable

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,560 @@
...
+    if (!nextSubtag()) {
+      return true;
+    }
+  }
+  
+  while (true) {
+    if (!isUnicodeVariantSubtag(getCurrentSubtagRef())) {
+      return true;
+    } else {
+      if (!addVariantSubtag(transformedExtensionId)) {
```
neildhar, line 243:
```suggestion
    } else if (!addVariantSubtag(transformedExtensionId)) {
```

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,560 @@
...
+    languageId->regionSubtag = getCurrentSubtag();
+    if (!transformedExtensionId) {
+      toASCIIUpperCase(languageId->regionSubtag);
+    }
+    if (!nextSubtag()) {
+      return true;
+    }
+  }
+  
+  while (true) {
```
neildhar, line 239: You may want to consider making this a do..while loop. Loosely, it could then be restructured as:
```
do { 
  // body
} while (nextSubTag);

return true;
```

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,560 @@
...
+  bool nextSubtag();
+  
+  ParsedLocaleIdentifier parsedLocaleIdentifier;
+  std::u16string localeId_;
+  std::u16string originalLocaleId_;
+  int subtagStart_;
+  int subtagEnd_;
+};
+
+LanguageTagParser::LanguageTagParser(const std::u16string &localeId) {
```
neildhar, line 171: We can use something called a "member initialiser list" to initialise the properties in the constructor. You can read more about that [here](https://en.cppreference.com/w/cpp/language/constructor).

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,539 @@
...
+  // = alphanum{3,8};
+  return isCharType(subtagRef, 3, 8, &isASCIILetterOrDigit);
+}
+bool isUnicodeExtensionKey(const llvh::ArrayRef<char16_t> subtagRef) {
+  // https://unicode.org/reports/tr35/#Unicode_Language_and_Locale_Identifiers
+  // = alphanum alpha;
+  if (subtagRef.size() != 2) {
+    return false;
+  }
+  return isASCIILetterOrDigit(subtagRef.front()) && isASCIILetter(subtagRef.back());
```
neildhar, line 106: consistency nit (same below)

```suggestion
  return subtagRef.size() == 2 && isASCIILetterOrDigit(subtagRef.front()) && isASCIILetter(subtagRef.back());
```

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,539 @@
...
+class LanguageTagParser {
+ public:
+  LanguageTagParser(const std::u16string &localeId);
+  LanguageTagParser();
+  ~LanguageTagParser();
+
+  // public function declaration
+  bool parseUnicodeLocaleId();
+  ParsedLocaleIdentifier getParsedLocaleId();
+  std::u16string toString();
```
neildhar, line 147: Is this ever used?

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,539 @@
...
+  std::u16string localeId_;
+  const std::u16string &originalLocaleId_;
+  int subtagStart_;
+  int subtagEnd_;
+};
+
+LanguageTagParser::LanguageTagParser(const std::u16string &localeId) : localeId_{localeId}, originalLocaleId_(localeId), subtagStart_(0), subtagEnd_(-1) {
+  toASCIILowerCase(localeId_);
+}
+LanguageTagParser::~LanguageTagParser() = default;
```
neildhar, line 174:
```suggestion
```

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,539 @@
...
+  // = (sep alphanum{2,8})+;
+  return isCharType(subtagRef, 2, 8, &isASCIILetterOrDigit);
+}
+}
+
+class LanguageTagParser {
+ public:
+  LanguageTagParser(const std::u16string &localeId);
+  LanguageTagParser();
+  ~LanguageTagParser();
```
neildhar, line 142:
```suggestion
```

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,539 @@
...
+  std::u16string getCurrentSubtag();
+  llvh::ArrayRef<char16_t> getCurrentSubtagRef();
+  bool hasMoreSubtags();
+  bool nextSubtag();
+  
+  ParsedLocaleIdentifier parsedLocaleIdentifier;
+  std::u16string localeId_;
+  const std::u16string &originalLocaleId_;
+  int subtagStart_;
+  int subtagEnd_;
```
neildhar, line 168: Use unsigned types, `size_t` is appropriate here.

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,539 @@
...
+    if (!nextSubtag()) {
+      return true;
+    }
+  }
+}
+
+bool LanguageTagParser::addVariantSubtag(bool transformedExtensionId) {
+  ParsedLanguageIdentifier &languageId = transformedExtensionId ? parsedLocaleIdentifier.transformedLanguageIdentifier : parsedLocaleIdentifier.languageIdentifier;
+
+  // Insert in alphabetical order
```
neildhar, line 245: We should either sort them later, or use a `std::set`. If we can fit sorting in a logical way, then I have a slight preference for that, otherwise either is fine.

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,539 @@
...
+  if (!parseExtensions()) {
+    return false;
+  }
+  
+  return !hasMoreSubtags();
+}
+
+bool LanguageTagParser::parseUnicodeLanguageId(bool transformedExtensionId) {
+  if (!transformedExtensionId) {
+    if (!nextSubtag() || !isUnicodeLanguageSubtag(getCurrentSubtagRef())) {
```
neildhar, line 196: For every other function, we advance to the subtag we want it to process before invoking it. So it feels a little weird that this function has to initialise the subtag on its own. Perhaps it would make sense to hoist the call to `nextSubtag` out into `parseUnicodeLocaleId`?

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,539 @@
...
+
+bool LanguageTagParser::hasMoreSubtags() {
+  return localeId_.length() > 0 && subtagEnd_ < (int)localeId_.length() - 1;
+}
+bool LanguageTagParser::nextSubtag() {
+  if (!hasMoreSubtags()) {
+    return false;
+  }
+  
+  int length = localeId_.length();
```
neildhar, line 497:
```suggestion
  size_t length = localeId_.length();
```

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,539 @@
...
+  
+  if (subtagEnd_ >= subtagStart_) {
+    if (!isSubtagSeparator(localeId_[subtagEnd_+1])) {
+      return false;
+    }
```
neildhar, line 502: Is this ever possible?

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,539 @@
...
+  }
+  
+  return true;
+}
+
+bool LanguageTagParser::hasMoreSubtags() {
+  return localeId_.length() > 0 && subtagEnd_ < (int)localeId_.length() - 1;
+}
+bool LanguageTagParser::nextSubtag() {
```
neildhar, line 492: I think much of this code would become simpler if `subtagEnd_` was instead one past the end of a subtag.

neildhar (changes requested):
This is looking good! I've reviewed through `parseUnicodeLanguageId`. I'll take a look at extension parsing tomorrow.

lib/Platform/Intl/BCP47Parser.cpp
silva-nick: No, I used to use it for testing, but I don't think it is necessary.

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,516 @@
...
+          return false;
+        }
+        break;
+      }
+      case 'x': {
+        // private use extension
+        if (!parsePUExtension()) {
+          return false;
+        }
+        return true;
```
neildhar, line 279: Add a comment here that the PU extension must be the last extension.

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,516 @@
...
+    languageId.variantSubtagList.insert(position, subtag);
+  }
+  return true;
+}
+
+bool LanguageTagParser::parseExtensions() {
+  while (true) {
+    if (subtagEnd_ != subtagStart_) {
+      return true;
+    }
```
neildhar, line 247: We should merge these conditions for readability.

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,516 @@
...
+        // other extension
+        if (!parseOtherExtension(singleton)) {
+          return false;
+        }
+        break;
+      }
+    }
+  }
+
+  return false;
```
neildhar, line 291: This is currently unreachable, once we merge the conditions above, it should return true.

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,516 @@
...
+}
+
+// unicode_locale_extensions = sep [uU]
```
neildhar, line 294: It would be good to have these comments at each function, they're very useful for understanding the code.

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,516 @@
...
+// |(sep attribute)+ (sep keyword)*) ;
+//
+// keyword = = key (sep type)? ;
+//
+// key = = alphanum alpha ;
+//
+// type = = alphanum{3,8}
+//  (sep alphanum{3,8})* ;
+//
+// attribute = alphanum{3,8} ;
```
neildhar, line 305: These are implemented above, so the comments should just be in the place where the implementation is.

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,516 @@
...
+bool LanguageTagParser::parseUnicodeExtension() {
+  if (!parsedLocaleIdentifier.unicodeExtensionAttributes.empty() ||
+      !parsedLocaleIdentifier.unicodeExtensionKeywords.empty()) {
+    return false;
+  }
+  
+  while (isUnicodeExtensionAttribute(getCurrentSubtagRef())) {
+    // Insert in sorted order
+    if (parsedLocaleIdentifier.unicodeExtensionAttributes.empty()) {
```
neildhar, line 314: Similarly here, it's probably better to sort this at the end, or once it is needed. If sorting is explicitly only needed for canonicalisation, then we can also do it there.

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,516 @@
...
+
+    if (!nextSubtag()) {
+      return true;
+    }
+  }
+  
+  while (isUnicodeExtensionKey(getCurrentSubtagRef())) {
+    std::u16string key = getCurrentSubtag();
+
+    parsedLocaleIdentifier.unicodeExtensionKeywords.insert({key, {}});
```
neildhar, line 332: Use the iterator returned by `insert` for the next line

lib/Platform/Intl/BCP47Parser.cpp
```diff
@@ -0,0 +1,516 @@
...
+      while (isUnicodeExtensionKeyTypeItem(getCurrentSubtagRef())) {
+        extensionKeyTypes->push_back(getCurrentSubtag());
+        if (!nextSubtag()) {
+          return true;
+        }
+      }
+    }
+  }
+  
+  return true;
```
neildhar, line 347: From my reading of the spec, we must have at least one keyword or attribute, however, it is possible to get to this point without any.

PR #652, opened 2021-12-03
Add utility functions

SarahCoded:
## Summary
This PR contains four utility functions - getOptionString, getOptionBool, getNumberOption and defaultNumberOption. They will be useful for implementing future Intl functions such as DateTimeFormat, NumberFormat, and Collator. 

getOption has been split into getOptionString and getOptionBool, diverging from the ecma402 spec, which defines getOption as one function. This has been done to improve readability throughout the codebase, and it means that getOptionBool won’t have to pass through empty vectors. 

getOptionString reviews a property, such as hour, and sees if there is a match with the valid properties pertaining to hour (numeric, 2-digit). If the property is empty, a fallback option will be returned. If there is a value, getOption will see if there is a match within the values vector, in which case the matching value will be returned, else if there is not a match, a RangeError will be raised. Likewise getOptionBool returns a bool value or empty fallback, but without the need for a vector.

getOptionNumber, instead of checking a string value match, checks whether the number option falls within an accepted range of minimum and maximum before returning through the defaultNumberFunction.

## Test Plan
These four functions have been used within Intl.DateTimeFormat() successfully to validate the options that the user has given. They have passed test262 tests whereby the spec defines that a RangeError should be raised. Error messages display which option has caused an exception where possible. A RangeError will be thrown if a number is passed as an option property in getOptionString(), and vice versa if a string is passed through getOptionNumber().


neildhar (commented):
It looks like merging the `getOption` functions may still hurt readability.

I propose having two functions, with the bool version having a simpler implementation since it does not need to take a `values` argument. They could have the following signatures:
```
vm::CallResult<Option> getOptionBool(const Options &options, const std::u16string &property, const Option &fallback, vm::Runtime *runtime)
```
```
vm::CallResult<Option> getOptionString(const Options &options, const std::u16string &property, llvh::ArrayRef<std::u16string> values, const Option &fallback, vm::Runtime *runtime)
```

What do you think about how that would fit in with the existing DateTimeFormat and NumberFormat code you have written?

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -191,7 +191,83 @@ LocaleMatch lookupMatcher(
...
+  if (value->second.isString()) {
+    if (!values.empty() &&
+        llvh::find(values, value->second.getString()) == values.end()) {
+      return runtime->raiseRangeError(vm::TwineChar16("Value ") + vm::TwineChar16(value->second.getString().c_str()) + vm::TwineChar16(" out of range for Intl.DateTimeFormat options property " + vm::TwineChar16(property.c_str())));
+    }
+  } else {
+    if (!values.empty() &&
+        (value->second.getBool() != true && value->second.getBool() != false)) {
+      return runtime->raiseRangeError(vm::TwineChar16("Value ") + vm::TwineChar16(value->second.getString().c_str()) + vm::TwineChar16("  out of range for Intl.DateTimeFormat options property " + vm::TwineChar16(property.c_str())));
+    }
```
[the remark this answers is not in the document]
SarahCoded: I’m having trouble trying to get an error throw here if there is something that isn’t entered in as a bool i.e. hour12: “something”. Options is expecting a bool, so I think it just turns the value into True when using getBool/isBool because its kind is bool?

SarahCoded:
> It looks like merging the `getOption` functions may still hurt readability.
> 
> I propose having two functions, with the bool version having a simpler implementation since it does not need to take a `values` argument. They could have the following signatures:
> 
> ```
> vm::CallResult<Option> getOptionBool(const Options &options, const std::u16string &property, const Option &fallback, vm::Runtime *runtime)
> ```
> 
> ```
> vm::CallResult<Option> getOptionString(const Options &options, const std::u16string &property, llvh::ArrayRef<std::u16string> values, const Option &fallback, vm::Runtime *runtime)
> ```
> 
> What do you think about how that would fit in with the existing DateTimeFormat and NumberFormat code you have written?

Thanks, I’ve managed to get it all to work by splitting up the options. I actually ended up also separating getOptionNumber out to DefaultNumberOption(as the spec suggests). 
I had to change uint8_t to a double in places, as that was the only number value type used to create an Option to return as far as I could tell, and uint8_t wasn’t working with the NAN check.
The only other thing that might be amiss is that for passing the undefined fallback option. I have been using Option(nullptr). It works, but produces the the following warning when building 
```
warning: implicit conversion of nullptr constant to 'bool' [-Wnull-conversion]
      Option(nullptr),
      ~~~~~~ ^~~~~~~
             false
```
Would there be a better way to represent undefined instead of an empty string?

neildhar:
As discussed previously, we can use `llvh::Optional` to represent `undefined`. With these split functions, we can then write them with the following signatures:
```
vm::CallResult<llvh::Optional<bool>> getOptionBool(const Options &options, const std::u16string &property, llvh::Optional<bool> fallback, vm::Runtime *runtime)

vm::CallResult<llvh::Optional<std::u16string>> getOptionString(const Options &options, const std::u16string &property, llvh::ArrayRef<std::u16string> values, llvh::Optional<std::u16string> fallback, vm::Runtime *runtime)
```

You'll need to check for both exceptions and undefined in the return value, so be careful with that.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -192,6 +192,104 @@ LocaleMatch lookupMatcher(
...
+    const Option &fallback,
+    vm::Runtime *runtime) {
+  //  1. Assert: Type(options) is Object.
+  //  2. Let value be ? Get(options, property).
+  auto value = options.find(property);
+  //  3. If value is undefined, return fallback.
+  if (value == options.end()) {
+    return Option(fallback);
+  }
+  //  8. Return value.
```
neildhar, line 242: For now, it is okay to do what the Java implementation does, and throw an exception if this is not a bool (by checking using `isBool`)

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -192,6 +192,104 @@ LocaleMatch lookupMatcher(
...
+    return Option(fallback);
+  }
+  // 4. Assert: type is "boolean" or "string".
+  // 5. If type is "boolean", then
+  // a. Set value to ! ToBoolean(value).
+  // 6. If type is "string", then
+  // a. Set value to ? ToString(value).
+  // 7. If values is not undefined and values does not contain an element equal
+  // to value, throw a RangeError exception.
+  if (!values.empty() &&
```
neildhar, line 218: Similarly, here, we should throw if it is not a string

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -192,6 +192,113 @@ LocaleMatch lookupMatcher(
...
+       llvh::find(values, value->second.getString()) == values.end())) {
+    return runtime->raiseRangeError(
+        vm::TwineChar16("Value ") +
+        vm::TwineChar16(value->second.getString().c_str()) +
+        vm::TwineChar16(
+            " out of range for Intl.DateTimeFormat options property ") +
+            vm::TwineChar16(property.c_str()));
+  }
+  // 8. Return value.
+  return llvh::Optional<std::u16string>(value->second.getString());
```
neildhar, line 229: Do we need to explicitly call this constructor here?

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -192,6 +192,113 @@ LocaleMatch lookupMatcher(
...
+// https://tc39.es/ecma402/#sec-defaultnumberoption
+vm::CallResult<llvh::Optional<uint8_t>> defaultNumberOption(
+    const double value,
+    const std::uint8_t minimum,
+    const std::uint8_t maximum,
+    llvh::Optional<uint8_t> fallback,
+    vm::Runtime *runtime) {
+  //  1. If value is undefined, return fallback.
+  if (!value && value != 0) {
+    return fallback;
```
neildhar, line 263: If value may be undefined, then it should be passed in as an optional. Zero and undefined are separate.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -192,6 +192,113 @@ LocaleMatch lookupMatcher(
...
+        vm::TwineChar16(
+            " out of range for Intl.DateTimeFormat options property ") +
+            vm::TwineChar16(property.c_str()));
+  }
+  //  8. Return value.
+  return llvh::Optional<bool>(value->second.getBool());
+}
+// https://tc39.es/ecma402/#sec-defaultnumberoption
+vm::CallResult<llvh::Optional<uint8_t>> defaultNumberOption(
+    const double value,
```
neildhar, line 256: This argument should be an `llvh::Optional<Option>`. If the option does not contain a number, we should throw an exception, as we do in other places.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -192,6 +192,113 @@ LocaleMatch lookupMatcher(
...
+    const std::uint8_t minimum,
+    const std::uint8_t maximum,
+    llvh::Optional<uint8_t> fallback,
+    vm::Runtime *runtime) {
+  //  1. Assert: Type(options) is Object.
+  //  2. Let value be ? Get(options, property).
+  auto value = options.find(property);
+  if (value == options.end()) {
+    return fallback;
+  }
```
neildhar, line 288: This check is now performed in DefaultNumberOption right? So we can delete this.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -192,6 +192,113 @@ LocaleMatch lookupMatcher(
...
+  //  3. Return ? DefaultNumberOption(value, minimum, maximum, fallback).
+  auto defaultNumber = defaultNumberOption(
+      value->second.getNumber(),
```
neildhar, line 291: We don't know that this property is a number, since we haven't checked it. The check will be performed in `defaultNumberOption`.

neildhar (changes requested):

lib/Platform/Intl/PlatformIntlApple.mm
SarahCoded: Do you mean to replace it with ```return value->second.getString()```? It says that there is no viable conversion from std::u16string to function return type 'vm::CallResult<llvh::Optional<std::u16string>> .

lib/Platform/Intl/PlatformIntlApple.mm
SarahCoded: I tried using llvh::Optional<Option> value, but this meant I needed to pass in value->second, which causes a fault when empty, and the idea was to check for the need for a fallback in defaultNumberOption? So I changed it to an unordered map iterator. Although this also meant that I had to pass options to defaultNumberOptions to check if it is empty. Sorry, I think I might have misunderstood what you meant.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -192,6 +192,112 @@ LocaleMatch lookupMatcher(
...
+            " out of range for Intl.DateTimeFormat options property ") +
+            vm::TwineChar16(property.c_str()));
+  }
+  //  8. Return value.
+  return llvh::Optional<bool>(value->second.getBool());
+}
+// https://tc39.es/ecma402/#sec-defaultnumberoption
+vm::CallResult<llvh::Optional<uint8_t>> defaultNumberOption(
+    const Options &options,
+    const std::unordered_map<std::u16string, Option>::const_iterator &value,
```
neildhar, line 257:
```suggestion
    llvh::Optional<Option> value,
```

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -192,6 +192,112 @@ LocaleMatch lookupMatcher(
...
+vm::CallResult<llvh::Optional<uint8_t>> getNumberOption(
+    const Options &options,
+    const std::u16string &property,
+    const std::uint8_t minimum,
+    const std::uint8_t maximum,
+    llvh::Optional<uint8_t> fallback,
+    vm::Runtime *runtime) {
+  //  1. Assert: Type(options) is Object.
+  //  2. Let value be ? Get(options, property).
+  auto value = options.find(property);
```
neildhar, line 286:
```suggestion
  llvh::Optional<Option> value;
  auto iter = options.find(property);
  if (iter != options.end()){
    value = Option(iter->second);
  }
```

neildhar (commented):
Passing an iterator here will make it harder to reuse these functions later (e.g. in `NumberFormat`). We should pass in an `llvh::Optional<Option>`.

lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -283,17 +283,18 @@ LocaleMatch lookupMatcher(
...
-      maximum,
-      fallback,
-      runtime);
+  auto defaultNumber =
+      defaultNumberOption(options, value, minimum, maximum, fallback, runtime);
   if (defaultNumber == vm::ExecutionStatus::EXCEPTION) {
-    return runtime->raiseRangeError(vm::TwineChar16(property.c_str()) + vm::TwineChar16(" value is out of range"));
+    return runtime->raiseRangeError(
+        vm::TwineChar16(property.c_str()) +
+        vm::TwineChar16(" value is out of range"));
```
neildhar, line 297: The exception should be thrown in `defaultNumberOption`

PR #654, opened 2021-12-13
Remove float cast flag in 0.10.0 branch

neildhar:


mrousavy:
Don't have permission to approve this PR - but my build now got accepted by Apple with this PR! 👍

kelset:
has this changed also been replicated into main, btw?

neildhar:
@kelset Yes, it is in main as 20aa6fab478f2cd6ded5f6980fc4ad7ea82f3ee1

kelset:
cool - so it's either a matter of making a 0.10.1 or a 0.11, if I'm understanding the current status of releases for Hermes.

Considering that for 0.67 we ended up staying on 0.9, maybe at this point might be worth just coordinating the branch cut for 0.68 with a 0.11 for hermes?
(cc @lunaleaps)

neildhar:
I was actually wondering whether we should just replace the artefact for the 0.10.0 CocoaPod with the patched version, given that this is a minor but necessary change for it to work. But I am currently away and haven't had a chance to think through and discuss any potential pitfalls of that approach.

billnbell:
??

PR #747, opened 2022-05-10
Fixing Hermes Android builds on Windows

mganandraj:
Hermes Android builds fail on Windows while trying run a bash script which tried to guess the target architecture.
This patch avoid the bash script when running on Windows.

## Summary

Hermes Android builds fail on Windows while trying run a bash script which tried to guess the target architecture.
This patch avoid the bash script when running on Windows.

## Test Plan

Verified that the android build succeeds on Windows host


neildhar:
Hi @mganandraj, thanks for looking into this. I'm wondering whether there's a more general way to handle this. In particular, it doesn't seem right to assume that we are building for an Android target (and depend on variables set by the NDK) whenever we are building on a Windows host with a non-MSVC compiler. 

I'm wondering if we could either:
1. Print a warning when on a Windows host and just set the value to "unknown"
2. Skip the call to `get_host_triple` when `LLVM_HOST_TRIPLE ` has already been set by the caller, and then make sure we manually set that somewhere higher up (possibly even in Gradle).

If the value doesn't actually matter, 1. is probably simpler.

mganandraj:
> 

Fully agreed ! It is confusing to try to psedo-guess the value which is not used...

PR #790, opened 2022-07-29
Remove the debugger in the Windows and Linux cli

fbmal7:


tmikov:
Why remove the debugger?

fbmal7:
> Why remove the debugger?

@tmikov This is just removing the debugger from the `hermes` CLI executable itself. It's currently unusable already, so this will strictly be a size/perf win. We also are obviously keeping the debugger in `hdb` for people who actually need to use the debugger.

PR #806, opened 2022-09-01
Fix compilation warnings in BigIntSupport.h

vmoroz:
## Summary

Fix compilation warnings in `BigIntSupport.h`.
The related changes in `BigIntSupport.h`:
- `size_t` constants are changed to `uint32_t` to avoid unnecessary conversion between 32 and 64 integers.
- `unsigned` type changed to `uint32_t` to be explict about the type size.
- `static` keyword for standalone variables and functions is changed to `inline`. The `static` keyword in header files forces creation of copies of the variables and functions in each compilation unit that use the header file and unnecessary increases size of binary code.
- Fixed a couple of typos in comments. 

## Details

Currently the code in `BigIntSupport.h` produces the following warnings in MSVC:

> [build] (...clipped...)\include\hermes/Support/BigIntSupport.h(111,55): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?) [(...clipped...)\build\lib\VM\hermesVMRuntime.vcxproj]

After the fix we do not see these warnings anymore.

The new code also produces more efficient machine code.
See below x64 debug assembly before and after the change.

Before the change:
Before_Fix

After the change:
After_Fix

The new code is shorter by 5 bytes.

Though, the code is probably completely optimized out for the Release bits and the only real benefit is that we do not see the warnings anymore.

## Test Plan

All unit tests are passing. No changes in program logic.

### DIFF
--- a/include/hermes/Support/BigIntSupport.h
+++ b/include/hermes/Support/BigIntSupport.h
@@ -68,15 +68,17 @@ std::optional<std::string> getNumericValueDigits(
 using SignedBigIntDigitType = int64_t;
 using BigIntDigitType = uint64_t;
 
-static constexpr size_t BigIntDigitSizeInBytes = sizeof(BigIntDigitType);
-static constexpr size_t BigIntDigitSizeInBits = BigIntDigitSizeInBytes * 8;
+inline constexpr uint32_t BigIntDigitSizeInBytes =
+    static_cast<uint32_t>(sizeof(BigIntDigitType));
+inline constexpr uint32_t BigIntDigitSizeInBits = BigIntDigitSizeInBytes * 8;
 
 /// Arbitrary upper limit on number of Digits a bigint may have.
-static constexpr size_t BigIntMaxSizeInDigits = 0x400; // 1k digits == 8k bytes
+inline constexpr uint32_t BigIntMaxSizeInDigits =
+    0x400; // 1k digits == 8k bytes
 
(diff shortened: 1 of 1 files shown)


include/hermes/Support/BigIntSupport.h
```diff
@@ -107,7 +107,8 @@ inline unsigned constexpr maxCharsPerDigitInRadix(uint8_t radix) {
   // With minNumBitsPerChar being the lower bound estimate of how many bits each
   // char can represent, the upper bound of how many chars "fit" in a bigint
   // digit is ceil(sizeofInBits(bigint digit) / minNumBitsPerChar).
-  unsigned numCharsPerDigits = BigIntDigitSizeInBits / (1 << minNumBitsPerChar);
+  unsigned numCharsPerDigits =
+      static_cast<unsigned>(BigIntDigitSizeInBits) / (1 << minNumBitsPerChar);
```
jpporto, line 111: Instead of casting, can you make BigIntDisitSizeInBits a uint32_t? It doesn't need to be 64-bits.

vmoroz: I have changed the constants to `unit32_t`.

vmoroz: I have also updated the comments.

vmoroz:
@jpporto please let me know what the linter issues are - I cannot see them. I will fix them.

PR #807, opened 2022-09-01
Fix creation of std::unique_ptr<const Buffer>

vmoroz:
## Summary

The `hbc::BCProviderFromBuffer::createBCProviderFromBuffer` method expects `std::unique_ptr<const Buffer>`, but we currently provide `std::unique_ptr<Buffer>` (no `const`). It is more optimal to provide value of the same type to avoid any unnecessary value conversion.
The code is changed to call `std::make_unique<const Buffer>` instead of `std::make_unique<Buffer>`.

## Test Plan

All unit tests are passing. There are no changes in program flow.

### DIFF
--- a/lib/BCGen/HBC/BytecodeFormConverter.cpp
+++ b/lib/BCGen/HBC/BytecodeFormConverter.cpp
@@ -165,7 +165,7 @@ class BytecodeFormConverter {
       BytecodeForm sourceForm)
       : bytes_(bytes), fields_(fields) {
     auto res = hbc::BCProviderFromBuffer::createBCProviderFromBuffer(
-        std::make_unique<Buffer>(bytes.data(), bytes.size()), sourceForm);
+        std::make_unique<const Buffer>(bytes.data(), bytes.size()), sourceForm);
     if (!res.first) {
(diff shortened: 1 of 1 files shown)


vmoroz:
BTW, I still see in debug bits
```
Exception has occurred: W32/0xC0000409
Unhandled exception at 0x00007FF8231D2240 (ntdll.dll) in HermesBCGenTests.exe: Indirect call guard check detected invalid control transfer.
``` 
While this change is still good on its own, it seems that it did not completely fix the issue it targets yet.

vmoroz:
I have discovered the root cause of the issue that I was hoping to address with this PR - it was an incorrect `/guard:cf` configuration that we have in `microsoft/hermes-windows`.
Thus, I am renaming this PR to be more as a 'nice-to-have' change than something that fixes a serious issue.

neildhar:
Hey @vmoroz, given that this doesn't fix a problem as originally suspected, I'd favour omitting this change, to minimise blame and for consistency with other similar calls.

vmoroz:
Closing as low priority change that does not affect any important functionality.

PR #810, opened 2022-09-01
Fix TypedArray functions

vmoroz:
## Summary

Fix TypedArray `byteLength()` and `length()` functions when the buffer is detached.
Before the change they return non-zero value.
The PR is fixing it and they return zero in such case.

## Test Plan

The new code passes Node-API tests in hermes-windows repo which are otherwise fail.
It seems that hermes does not have any unit tests for the detached buffer.
I am looking for some advice on how I must add the new test.


vmoroz:
Sorry, I forgot to format the code - it caused the linter failure.

PR #813, opened 2022-09-15
Use multiple cores when building with CMake

cipolleschi:
Summary:
When installing RNTester in React Native we build hermes from source. Currently, we are not asking our users to install ninja, so the build fallback uses CMake. We realized that we have very long builds and that is because we are not using multiple cores when building Hermes.

This diff tries to address that problem, using the same code we are using in other CI jobs

Differential Revision: D39538731


Kudo:
just a note that there's another tunable `-j ${NUM_CORES}` when building hermesc.
reference: https://github.com/facebook/react-native/blob/9cb716ff76e45583999b040be8a99b00f0e022e1/sdks/hermes-engine/utils/build-apple-framework.sh#L33

PR #825, opened 2022-10-06
Extract Keywords into a separate file

avp:
Summary:
For easier reusability, move the class into another file.
Eventually we can move it from the `sem` namespace into `sema`,
but for now keep everything working as it is.

Differential Revision: D39985176


avp:
Superseded by #834

PR #829, opened 2022-10-06
Add new SemanticResolver

avp:
Summary:
Implement a new SemanticResolver which extracts all the resolution
information into `SemContext`.

This is based directly on Juno's implementation with the same
structure except for the fact that we use RAII classes instead
of the callback structure. There's nothing preventing the callback
structure from being used here, and helper methods for using callbacks
can be easily added in the future, it's just that RAII seemed fairly
straightforward and simple for this implementation.

With the exception of the Unresolver pass,
everything in the Juno implementation is contained in this diff.

The Hermes implementation also does some manipulation of the AST,
including the try/catch transform, which will be in future diffs.

Differential Revision: D39985175

### DIFF
--- a/include/hermes/ADT/ScopedHashTable.h
+++ b/include/hermes/ADT/ScopedHashTable.h
@@ -162,6 +162,15 @@ class ScopedHashTable {
     return result->second->value_;
   }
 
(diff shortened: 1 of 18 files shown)


avp:
Superseded by #834

PR #830, opened 2022-10-06
Add new SemContext struct

avp:
Summary:
Place the new SemContext struct in the `sema` namespace.
It contains storage for information on functions, scopes, and decls.

The `SemContext` struct does NOT include the following mappings:
* identifier -> Decl
* AST node -> LexicalScope
These two mappings will be stored directly on the AST via the decoration
mechanism, which was not available in Juno.

Differential Revision: D39991926


avp:
Superseded by #834

PR #831, opened 2022-10-06
Add ScopeDecorationBase

avp:
Summary:
In order to associate scopes with the AST nodes that create them
without using a side table, add a `ScopeDecorationBase` that only
has one field: the scope.

Differential Revision: D39992368


avp:
Superseded by #834

PR #834, opened 2022-10-06
[SH][PR] Semantic Resolution

avp:
Summary:
Support marking identifiers which can be affected by `eval` and `with`
as "unresolvable", much like we do in Juno.

We don't yet support the `canRename` flag, because static Hermes
is as yet unconcerned with outputting JS source,
but there is a comment in the code that explains where to populate such
a flag if it exists.

Differential Revision: D40068199


tmikov:
rejected!

PR #854, opened 2022-11-21
Object::instanceOf() should be const-qualified

RedBeard0531:
The virtual function on jsi::Runtime already takes a const Object.

## Summary

It is annoying and potentially expensive to have to copy the `Object` just to check if it is an instance of some class.

## Test Plan

If it compiles, it should be fine.


tmikov:
I don't have an opinion on this, but I wonder whether we should consider the case where the object is a proxy, meaning the `getPrototypeOf()` handler can mutate the object?

neildhar:
This was done across the board in b81666598672cb5f8b365fe6548d3273f216322e.

PR #890, opened 2023-01-13
Parse class property modifiers in TypeScript

seonggun-kim:
## Summary

Parse class member accessibility (i.e., `private`, `protected`, and `public`) and `readonly` modifiers in TypeScript.

In TypeScript, modifiers can also be used as identifier when they appear in a place where an identifier is expected. To handle such cases easily, let the lexer to convert the current reserved word token into an identifier so the parser can continue without any specialization.

The modifiers are isolated in an optional `TSModifiers` node so JS/Flow ASTs are not affected by this change and more TypeScript modifiers can be added in the future without having to change other AST nodes.

In the course, it also fixes a bug that `private`-declared property (e.g., `private foo`) was treated as private property (e.g., `#foo`). Please refer to https://www.typescriptlang.org/docs/handbook/2/classes.html#caveats for the differences.

## Test Plan

By building check-hermes target. New test cases have been added.


seonggun-kim:
test262 failed because `readonly` is reserved even for normal JS. I will update the PR after fixing it.
```
COMPILE_FAILED test262/test/language/reserved-words/unreserved-words.js
	/var/folders/bl/wbxjgtzx7j5_mjsmfr3ynlc00000gp/T/tmpdojkd95n/language/reserved-words/unreserved-words.strict.js:250:5: error: 'identifier' expected in declaration
	var readonly = 1;
	~~~~
```

lib/Parser/JSParserImpl.cpp
```diff
@@ -20,6 +20,27 @@ namespace hermes {
 namespace parser {
 namespace detail {
 
+namespace {
+
+bool canNextTokenFollowModifier(OptValue<TokenKind> optNext) {
```
avp, line 25: Can we have a doc-comment on this function? It needs a description of what it does, return value, etc. Also, please put a `TS` in this function's name to indicate it's only for typescript.

lib/Parser/JSParserImpl.cpp
```diff
@@ -4579,15 +4624,20 @@ Optional<ESTree::ClassBodyNode *> JSParserImpl::parseClassBody(SMLoc startLoc) {
         break;
 
       case TokenKind::rw_static:
-        // static MethodDefinition
-        // static FieldDefinition
-        isStatic = true;
-        advance();
+        if (context_.getParseTS() && isStatic) {
+          // when `static` is already seen, treat the current one as identifier.
+          lexer_.unreserveCurrentToken();
```
avp, line 4629: I think you've uncovered a more wide-reaching problem in the Hermes parser  relating to the use of `static` as a field name. Before class props were introduced, there was a simple and limited set of tokens after `static` that would indicate it was the method name instead of a modifier but now it is valid to have `static` be the name of the property as well.

This will require changes to work for Flow, JS, and TS. I will put up a fix and link you to it when it lands - that should allow you to simplify logic like this because everything regarding the reswords should be able to be properly handled in the call below.

I need to think about this more, but I'm not sure yet whether we'll end up needing `unreserveCurrentToken` or something like it - it's likely we won't because we do have logic in `parsePropertyName` to handle reswords and pull the identifiers out of them (no mutation of the `Token` itself needed).

avp (commented):
Thanks for updating this. I'll look at fixing the existing bug relating to `static` (see comment) and get back to you when it's up.

lib/Parser/JSParserImpl.cpp
seonggun-kim: That's fair. Then perhaps I might move it to `class JSParserImpl` as a TS-specific method so we can have the implementation in `JSParserImpl-ts.cpp` file. Does that make sense to you?

Similarly, do you think `getAccessibilityIdent()` that I added should also be TS-specific? If so, I will do the same.

avp: > getAccessibilityIdent()

It seems like instead of storing a `TokenKind` you can just store `NodeLabel accessibility` and do `accessibility = tok_->getResWordIdent()` at the assignment points, to avoid needing that kind of function any more. Is there something that prevents using `getResWordIdent` and `getResWordOrIdent` for problems like these in this context?

seonggun-kim: No, nothing particular (other than my ignorance of the API 😀). I think that would be better. Thank you for the suggestion!

lib/Parser/JSParserImpl.cpp
avp: I've added https://github.com/facebook/hermes/commit/0d012688c8546bc927f3a011c02e6394c0f88fbf to make `static` parse properly, along with some tests.

seonggun-kim: I've updated this PR based on the commit.
* Removed `unreserveCurrentToken` and `getAccessibilityIdent`.
* Moved `canNextTokenFollowModifier` to `JSParserImpl:: canFollowModifierTS`, along with the doc-comment.

avp (commented):
One minor comment.

lib/Parser/JSParserImpl.h
```diff
@@ -1284,6 +1285,30 @@ class JSParserImpl {
...
       ESTree::IdentifierNode *ident);
+
+  /// Check if the given token kind may come right after a modifier.
+  /// It is often called with a lookahead token to decide whether the current
+  /// token is a modifier or not based on the context.
+  ///
+  /// \param optTokenKind an optional TokenKind value.
+  /// \return true if the given token kind is one of the modifier or identifier
+  /// kinds. false otherwise.
+  bool canFollowModifierTS(OptValue<TokenKind> optTokenKind) {
```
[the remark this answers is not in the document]
seonggun-kim: Sure, done!

PR #903, opened 2023-01-24
Use modern Podfile syntax to avoid polluting application build settings

yhkaplan:
## Summary

Fixes conflicts in pod build settings like the one below:

Example warning
> Can't merge user_target_xcconfig for pod targets: ["RNReanimated", "hermes-engine"]. Singular build setting CLANG_CXX_LANGUAGE_STANDARD has different values.

Background:

> The former attribute xcconfig is deprecated and will cause a linter error when pushing new versions to trunk. The new attributes are available as pod_target_xcconfig and user_target_xcconfig, which makes their effects more clear. The latter attribute (user_target_xcconfig) should be used with great care, because well designed Pods should be self-contained and make as few assumptions about their environment as possible. Furthermore, this attribute can cause conflicts when different values are specified by two Pods for a build setting which doesn't allow multiple values and so cannot be merged.
- https://blog.cocoapods.org/CocoaPods-0.38/

## Test Plan

Build in ApplePlatformsIntegrationMacTests

### DIFF
--- a/hermes-engine.podspec
+++ b/hermes-engine.podspec
@@ -33,7 +33,7 @@ Pod::Spec.new do |spec|
   spec.ios.vendored_frameworks = "destroot/Library/Frameworks/universal/hermes.xcframework"
   spec.osx.vendored_frameworks = "destroot/Library/Frameworks/macosx/hermes.framework"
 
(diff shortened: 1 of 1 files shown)


yhkaplan:
Hmm, probably need HERMES_ENABLE_DEBUGGER in test app 

```
Testing failed:
	No type named 'DebugFlags' in 'facebook::hermes::HermesRuntime'
	No member named 'debugJavaScript' in 'facebook::hermes::HermesRuntime'
	Testing cancelled because the build failed.

** TEST FAILED **
```

PR #905, opened 2023-01-27
Run DCE after SimplifyCFG to remove unused local functions.

seonggun-kim:
## Summary

SimplifyCFG may result in unused local functions by optimising out all their callsites. TypeInference may not be able to finish analysing such functions because their captured variables will remain NoType.

Here is an example excerpted and abbreviated from https://github.com/facebook/react-native/blob/0.70-stable/Libraries/StyleSheet/processTransform.js

```
function moduleFactory(module) {
  function processTransform(list) {
    // To be simplified by SimplifyCFG.
    // Any code following this if statement is unreachable.
    if (true === true) {
      return list;
    }

    // Although `list.forEach()` gets removed,
    // the anonymous function itself still remains.
    list.forEach(function (item) {
      switch (item.key) {
        case "foo":
          foo(item.value); // TypeInference cannot find where foo comes from
          break;
        case "bar":
          bar(item.value); // The same for bar
          break;
        default:
          throw new Error("Invalid key: " + key);
      }
    });
  }

  function foo(value) {
    console.log("This is foo(): " + value);
  }

  function bar(value) {
    console.log("This is bar(): " + value);
  }

  module.exports = processTransform;
}
```

TypeInference will keep re-trying to infer `LoadFrameInst` for `foo` and `bar`.

```
Inferring LoadFrameInst
Missing type for operand 0 of LoadFrameInst(Variable)
Inferring LoadPropertyInst
Inferring CallInst
Missing type for operand 0 of CallInst(LoadFrameInst)
```

We might want to look into TypeInference so it can handle such cases, but running DCE after SimplifyCFG is anyway beneficial, hence this PR.

## Test Plan

Tested with the above example, along with the unittests.

### DIFF
--- a/lib/Optimizer/PassManager/Pipeline.cpp
+++ b/lib/Optimizer/PassManager/Pipeline.cpp
@@ -52,6 +52,9 @@ void hermes::runFullOptimizationPasses(Module &M) {
   PM.addDCE();
 
   PM.addSimplifyCFG();
+  // SimplifyCFG may remove all callsites of a particular local function.
+  // Run DCE after SimplifyCFG to remove such functions if any.
+  PM.addDCE();
   PM.addStackPromotion();
   PM.addMem2Reg();
   PM.addStackPromotion();


tmikov:
This is a great find, thank you! We discussed this at length internally and we have identified two separate bugs revolving around dead functions and the IR invariant that loads must correspond to stores.

The first bug is in SimplifyCFG for code like this:
```js
function foo() {
    if (false) {
        let x = 10;
        let inner = function() { return x; }
    }

    return 0;
}
```

SimplifyCFG eliminates the code inside the `if`, but leaves the inner function, so we have an unmatched load:
```
function foo()
frame = [x, inner]
%BB0:
  %0 = ReturnInst 0 : number
function_end

function inner()
frame = []
%BB0:
  %0 = LoadFrameInst [x@foo]
  %1 = ReturnInst %0
function_end
```

The second bug is in StackPromotion for code like this:
```js
function foo() {
    let x = 10;
    if (false) {
        let inner = function() { return x; }
    }

    return 0;
}
```
Here SimplifyCFG has previously eliminated the code inside the `if`, but that is fine, since we still have a store. However then StackPromotion eliminates the store, resulting in the same problem.

Since correctness cannot depend on pass order, we are fixing both problems directly:
- SimplifyCFG will delete all functions that have become unreachable.
- We actually have a new version of StackPromotion called SimpleStackPromotion which is less aggressive and doesn't have this problem. We are replacing the former with the latter, precisely because the former is poorly understood and too subtle.

cc: @neildhar @avp

PR #916, opened 2023-02-17
Update BuildingAndRunning.md

radex:
As far as I can tell, `ninja check-hermes` does not work, or at least not without extra steps not mentioned in the docs

```
$ ninja check-hermes
ninja: error: loading 'build.ninja': No such file or directory
```

However this works:

```
cmake --build ./build --target check-hermes
```


neildhar:
@radex If you're following the instructions for Linux/macOS and passing in `-G Ninja` when configuring the build, you need to ensure that you run `ninja check-hermes` from the `build` directory.

That said, this PR is certainly an improvement, since it is consistent with the rest of the README, and works with other build systems.

Issue #914, opened 2023-02-16
Intl.NumberFormat currencyDisplay 'name' does not respect currency


asherLZR:
## Bug Description

Intl.NumberFormat does not respect the `currency` input when `currencyDisplay` is "name". Without `currencyDisplay`, `currency` is correctly taken into account.

Hermes version:
React Native version (if any): 0.70.6 & 0.71.3
OS version (if any): iOS 15.4
Platform (most likely one of arm64-v8a, armeabi-v7a, x86, x86_64):  arm64-v8a

## Steps To Reproduce

Run the below code sample on iOS.

## The Expected Behavior

```
new Intl.NumberFormat('de-DE', {
  style: 'currency',
  currency: 'USD',
  currencyDisplay: 'name',
}).format(2)

// expected: "2,00 US-Dollar"
// actual: "2,00 Euro" - wrong

new Intl.NumberFormat('de-DE', {
  style: 'currency',
  currency: 'USD'
}).format(2)

// expected: "2,00 $"
// actual: "2,00 $" - correct
```

asherLZR:
Potentially the numberFormatter needs to have the currency defined in the locale identifier when it's of type .currencyPlural.
```
let numberFormatter = NumberFormatter()
numberFormatter.numberStyle = .currencyPlural
numberFormatter.locale = Locale(identifier: "de-DE@currency=USD")
```

PR #919, opened 2023-02-17
fix: specify currency in locale identifier when formatting currency plural

asherLZR:
## Summary
Intl.NumberFormat does not respect the currency input when currencyDisplay is "name". Without currencyDisplay, currency is correctly taken into account.

Resolves https://github.com/facebook/hermes/issues/914

## Test Plan
Build Hermes and installed on RN application, tested against multiple currency/locale pairs. 

```
new Intl.NumberFormat('de-DE', {
  style: 'currency',
  currency,
  currencyDisplay: 'name',
}).format(2)
```

### DIFF
--- a/lib/Platform/Intl/PlatformIntlApple.mm
+++ b/lib/Platform/Intl/PlatformIntlApple.mm
@@ -2468,10 +2468,8 @@ void NumberFormatApple::initializeNSFormatters() noexcept {
   // - compactDisplay is not supported.
   // - signDisplay is not supported.
   // - NSNumberFormatter has maximumIntegerDigits, which is 42 by default
-  auto nsLocale =
-      [NSLocale localeWithLocaleIdentifier:u16StringToNSString(dataLocale_)];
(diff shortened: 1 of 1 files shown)


neildhar (changes requested):
Great find, thank you for reporting this and submitting a fix. This behaviour from `NSNumberFormatter` is odd, but I'm glad specifying the currency through the locale seems to work consistently.

Let's try to minimise duplicated work, I propose the following organisation:
1. Add a new `std::u16string nsLocalStr = dataLocale_;` at the top.
2. If `style` is `currency`, set `nsLocaleStr.append(u"@currency=").append(*currency_);`. This applies regardless of what `currencyDisplay` is set to.
3. Stop setting `nsNumberFormatter_.currencyCode`, since we are already specifying the currency through the locale, and that seems to do what we want in all cases.
4. Construct the new `NSLocale` at the bottom, using `nsLocaleStr`, and set `nsNumberFormatter_.locale`.

neildhar (approved):
LGTM, thanks for working on this!

Issue #929, opened 2023-03-03
Terminating JS Execution


jacobp100:
I've maintain a [web-worker](https://github.com/jacobp100/react-native-webworker) package for React Native that used to just spin up a new bridge, but I recently added a "light" mode that just spins up JSC or Hermes directly

For my use-case on this package, I have a calculator app that has things like summations and expensive operations, so it's relatively easy to write an equation that'd take a long time to finish - so I need a way to cancel them

My current solution is to add the `-emit-async-break-check` flag when compiling and then call `_runtime->watchTimeLimit(0)` to cancel the execution

Is this the best approach for this, or is there a more optimised way of doing this?

I noticed that when I added the flag, my byte-code skyrocketed from 10kb to 170kb (no typo - 17x increase). Is this expected - or is this possibly turning on additional flags I may be able to turn off?

neildhar:
Thanks for reporting this. Could you share a repro for the JS you are compiling and the compilation command? That will help us investigate the problem.

Also, note that there is currently no thread-safe way to interrupt execution from another thread (although there should probably be one). `watchTimeLimit` may not be invoked from another thread, since it modifies runtime state in unsafe ways.

If you're imposing a fixed time limit, could you set the `watchTimeLimit` with a non-zero value before execution starts?

jacobp100:
Yep! You can grab a copy of that code here:-

https://github.com/jacobp100/jacob-does-code/blob/master/assets/technicalc/dist/worker.min.js

Using the version of Hermes in react native 0.71, I see a 17x increase when adding the async break checks

I don’t know a fixed time before running - it’s only when a loading spinner shows up and the user hits ‘cancel’

Can I call `triggerTimeoutAsyncBreak` instead? How bad would it be to call it regardless - the alternative is spinning a computation in the background 🤣

tmikov:
@jacobp100  When I compile with and without -emit-async-break-check, both versions are around 155KB, the -emit-async-break-check is only slightly larger.

I am not sure how you are getting 17KB bytecode from a 107KB minified input JS file. Can you double-check that?

jacobp100:
Sorry yes - it looks like you're right. I'm not sure how I ended up with a 10kb file - I was pretty shocked too since the minified one was 100kb. Running now I get 171kb vs. 172kb - negligible.

tmikov:
171KB seems high. I am getting 155KB and about 120KB if I strip the backtrace information. What is the exact command you are using?

[...]

PR #934, opened 2023-03-06
Add asyncTriggerTimeout method

jacobp100:
## Summary

Adds method to terminate running JS without using timers (see #929)

## Test Plan

Unsure


API/hermes/hermes.h
```diff
@@ -188,12 +188,16 @@ class HERMES_EXPORT HermesRuntime : public jsi::Runtime {
...
   void unregisterForProfiling();
 
-  /// Register this runtime for execution time limit monitoring, with a time
-  /// limit of \p timeoutInMs milliseconds.
+  /// Asynchronously terminates the current execution.
   /// All JS compiled to bytecode via prepareJS, or evaluateJS, will support the
   /// time limit monitoring.  If JS prepared in other ways is executed, care
   /// must be taken to ensure that it is compiled in a mode that supports the
   /// monitoring (i.e., the emitted code contains async break checks).
+  void notifyNeedsTermination();
```
neildhar, line 196: I'd suggest renaming this to `asyncTriggerTimeout`, to make it clear that this will throw a timeout exception.

API/hermes/hermes.h
```diff
@@ -188,12 +188,16 @@ class HERMES_EXPORT HermesRuntime : public jsi::Runtime {
   /// Unregister this runtime for sampling profiler.
   void unregisterForProfiling();
 
-  /// Register this runtime for execution time limit monitoring, with a time
-  /// limit of \p timeoutInMs milliseconds.
+  /// Asynchronously terminates the current execution.
```
neildhar, line 191: We should also mention that this may be invoked from any thread (most other API methods may not).

neildhar (commented):
Apologies for the delay in getting back to you. The PR looks good, I just have some comments regarding naming.

It would also be great if you could add a test for the new mechanism to verify it works. You can use https://github.com/facebook/hermes/blob/3a582a557f66ea1d82211187f793ed1111faefda/unittests/API/APITest.cpp#L376 as a reference, although it can be much shorter (it will need to spawn a thread though).

jacobp100:
@neildhar Sorry I took so long to fix the issue you pointed out. Hopefully it's working now. Appreciate the comments, C++ is still very new to me

unittests/API/APITest.cpp
```diff
@@ -433,6 +433,24 @@ TEST(HermesWatchTimeLimitTest, WatchTimeLimit) {
...
+  const char *forEver = "for (;;){}";
+  uint32_t ShortTimeoutMS = 123;
+  {
+    // Single runtime with ~20 minute limit that will not be reached.
+    auto rt = makeHermesRuntime();
+    std::thread t([&]() {
+      std::this_thread::sleep_for(std::chrono::milliseconds(ShortTimeoutMS));
+      rt->asyncTriggerTimeout();
+    });
+    t.detach();
```
neildhar, line 447: Why detach the thread here?

unittests/API/APITest.cpp
```diff
@@ -433,6 +433,24 @@ TEST(HermesWatchTimeLimitTest, WatchTimeLimit) {
   }
 }
 
+TEST(HermesTriggerAsyncTimeoutTest, TriggerAsyncTimeout) {
+  // Some code that exercies the async break checks.
+  const char *forEver = "for (;;){}";
+  uint32_t ShortTimeoutMS = 123;
+  {
+    // Single runtime with ~20 minute limit that will not be reached.
```
neildhar, line 441: Looks like this comment needs to be updated

unittests/API/APITest.cpp
```diff
@@ -433,6 +433,24 @@ TEST(HermesWatchTimeLimitTest, WatchTimeLimit) {
   }
 }
 
+TEST(HermesTriggerAsyncTimeoutTest, TriggerAsyncTimeout) {
+  // Some code that exercies the async break checks.
```
neildhar, line 437:
```suggestion
  // Some code that loops forever to exercise the async interrupt.
```

unittests/API/APITest.cpp
jacobp100: I don’t know. What’s the alternative?

jacobp100: The tests failed without it. My understanding is `.join` won’t work because it’ll wait for the thread to finish - so we’ll end up in deadlock. But I didn’t actually try it so I could be wrong

unittests/API/APITest.cpp
jacobp100, line 441:
```suggestion
```

unittests/API/APITest.cpp
neildhar: We should be able to `join` after the `ASSERT_THROW`.

jacobp100: You're right - that works. What happened here? Without `join` and `detach`, it did not work. But to my understanding, the `ASSERT_THROW` won't return unless the thread executes the code to stop the runtime. But by that time, the `.join` shouldn't do anything, since we basically know the thread code has run. How does the `.join` afterwards make any impact here?

neildhar: From the reference for `joinable` ([link](https://en.cppreference.com/w/cpp/thread/thread/joinable)):

> A thread that has finished executing code, but has not yet been joined is still considered an active thread of execution and is therefore joinable.

And for `~thread` ([link](https://en.cppreference.com/w/cpp/thread/thread/~thread)):

> If `*this` has an associated thread (`joinable() == true`), `std::terminate()` is called.

So even if the thread has finished executing, it should be joined before it is destroyed. That said, that there is actually no guarantee that the thread has actually finished executing. All we know when the exception is thrown is that the interrupt flag in the runtime was set. In theory, the thread could have set the flag, and then immediately been suspended before it could return.

API/hermes/hermes.h
```diff
@@ -188,12 +188,19 @@ class HERMES_EXPORT HermesRuntime : public jsi::Runtime {
   /// Unregister this runtime for sampling profiler.
   void unregisterForProfiling();
 
+  /// Define methods to interrupt JS execution and set time limits.
+  /// All JS compiled to bytecode via prepareJS, or evaluateJS, will support interruption and
+  /// time limit monitoring.  If JS prepared in other ways is executed, care
```
neildhar, line 193: Turns out that this isn't true, which is why the tests in release mode are failing in CircleCI. Async break check is actually disabled by default, and in release builds, with debugging disabled, they are not emitted at all.

We should fix this, but I don't want to block your PR on it, so let's make the requirements clear.
```suggestion
  /// All JS compiled to bytecode via prepareJS, or evaluateJS, will support interruption and
  /// time limit monitoring if the runtime is configured with AsyncBreakCheckInEval.  If JS prepared in other ways is executed, care
```

unittests/API/APITest.cpp
```diff
@@ -433,6 +433,23 @@ TEST(HermesWatchTimeLimitTest, WatchTimeLimit) {
   }
 }
 
+TEST(HermesTriggerAsyncTimeoutTest, TriggerAsyncTimeout) {
+  // Some code that loops forever to exercise the async interrupt.
+  const char *forEver = "for (;;){}";
+  uint32_t ShortTimeoutMS = 123;
+  {
+    auto rt = makeHermesRuntime();
```
neildhar, line 441: For now, we need to manually enable this.

```suggestion
    auto rt = makeHermesRuntime(hermes::vm::RuntimeConfig::Builder().withAsyncBreakCheckInEval(true).build());
```

neildhar (changes requested):
LGTM overall, this should fix the test failure and make it ready to merge

Issue #932, opened 2023-03-03
hasIndices flag for regex


roryabraham:
## Problem

We were trying to use a regex with the `-d` / `hasIndices` flag, and noticed that it worked on web but not iOS/Android. I then discovered that it looks like this flag is just not yet implemented in the Hermes engine: https://github.com/facebook/hermes/blob/417190297242f14fed540378390d28926c2a7f7e/include/hermes/Regex/RegexTypes.h#L213-L343

## Solution

Implement the `-d` / `hasIndices` flag for Hermes regex.

## Additional Context

Our use case for the `-d` flag is described [here](https://github.com/Expensify/App/pull/15611#pullrequestreview-1322663118)

roryabraham:
This was implemented in https://github.com/facebook/hermes/pull/968 and should land in the next hermes release I think?

roryabraham:
https://github.com/facebook/hermes/commit/2afc7b09ffc74201b4dd530d9414b48e61a5196f does not appear to be included in a release yet, but hopefully soon 🙂

fbmal7:
@roryabraham This feature will be in 0.73.

PR #968, opened 2023-04-12
Add support for "hasIndices" RegExp flag

fabioh8010:
## Summary

This change adds support for `hasIndices` RegExp flag, according to ES2022 specifications.

References:
https://262.ecma-international.org/13.0/#sec-regexp-regular-expression-objects
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/hasIndices

This change is also related to this issue: https://github.com/facebook/hermes/issues/932

## Test Plan

``` javascript
var str = "The Quick Brown Fox Jumps Over The Lazy Brown Dog";
var regex = /quick\s(?<color>brown).+?(?<action>jumps).+?(?<where>over_)*/dgi;
var res = regex.exec(str);

print(Object.getOwnPropertyNames(res)); // 0,1,2,3,length,index,input,groups,indices

print(Object.getOwnPropertyNames(res.indices)); // 0,1,2,3,length,groups
print(res.indices[0]); // 4,26
print(res.indices[1]); // 10,15
print(res.indices[2]); // 20,25
print(res.indices[3]); // undefined

print(Object.getOwnPropertyNames(res.indices.groups)); // color,action,where
print(res.indices.groups.constructor); // undefined
print(res.indices.groups.color); // 10,15
print(res.indices.groups.action); // 20,25
print(res.indices.groups.where); // undefined
```


include/hermes/VM/PredefinedStrings.def
```diff
@@ -318,6 +318,8 @@ STR(multiline, "multiline")
 STR(unicode, "unicode")
 STR(sticky, "sticky")
 STR(dotAll, "dotAll")
+STR(indices, "indices")
```
ArekChr, line 321: Are there two definitions needed `indicates` and `hasIndicates`?

fabioh8010: Yes, they are used [here](https://github.com/facebook/hermes/pull/968/files#diff-f4a4510e019a334caf037e7c736570bcc110022baa742adc6bdf0782ad11e70aR515), [here](https://github.com/facebook/hermes/pull/968/files#diff-f4a4510e019a334caf037e7c736570bcc110022baa742adc6bdf0782ad11e70aR123) and [here](https://github.com/facebook/hermes/pull/968/files#diff-f4a4510e019a334caf037e7c736570bcc110022baa742adc6bdf0782ad11e70aR2203).

include/hermes/Regex/RegexTypes.h
```diff
@@ -270,27 +274,31 @@ class SyntaxFlags {
...
+    if (hasIndices)
+      result.push_back('d');
     if (global)
       result.push_back('g');
     if (ignoreCase)
       result.push_back('i');
     if (multiline)
       result.push_back('m');
+    if (dotAll)
+      result.push_back('s');
```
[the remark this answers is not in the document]
fabioh8010: It is just to align the order with ES2022 specification [here](https://262.ecma-international.org/13.0/#sec-get-regexp.prototype.flags).

lib/VM/JSLib/RegExp.cpp
```diff
@@ -460,6 +461,65 @@ static void createGroupsObject(
...
 }
 
+static void createIndicesArray(
+    Runtime &runtime,
+    Handle<JSArray> matchObj,
+    Handle<JSObject> mappingObj,
+    Handle<JSArray> indices) {
+  // If there are no capture groups, then set groups to undefined.
+  if (!mappingObj) {
+    auto executionStatus = JSObject::defineNewOwnProperty(
```
fbmal7, line 471: You should refrain from using `defineNewOwnProperty`. You can only use this API when you know for certain that the property you're adding does not exist on the current object. To be safe, use `defineOwnProperty`.

lib/VM/JSLib/RegExp.cpp
```diff
@@ -460,6 +461,65 @@ static void createGroupsObject(
...
+        indices,
+        runtime,
+        Predefined::getSymbolID(Predefined::groups),
+        PropertyFlags::defaultNewNamedPropertyFlags(),
+        Runtime::getUndefinedValue());
+    assert(
```
fbmal7, line 477: These `assert`s should not be written this way. You should check the result of the `defineProperty` and return in an error in a regular `if`. Also remember to wrap these in the appropriate `LLVM_(UN)LIKELY` For example,

```
if (LLVM_UNLIKELY(JSObject::defineNewOwnProperty(...) == ExecutionStatus::EXCEPTION)){
  return ExecutionStatus::EXCEPTION;
}
```

lib/VM/JSLib/RegExp.cpp
```diff
@@ -460,6 +461,65 @@ static void createGroupsObject(
...
+          JSObject::setNamedSlotValueUnsafe(
+              *groupsObj, runtime, desc.slot, indices->at(runtime, groupIdx));
+        });
+
+    // Add groups object to indices array.
+    auto executionStatus = JSObject::defineNewOwnProperty(
+        indices,
+        runtime,
+        Predefined::getSymbolID(Predefined::groups),
+        PropertyFlags::defaultNewNamedPropertyFlags(),
```
fbmal7, line 504: This should be `getDefaultNewPropertyFlags`.

lib/VM/JSLib/RegExp.cpp
```diff
@@ -460,6 +461,65 @@ static void createGroupsObject(
...
+        executionStatus != ExecutionStatus::EXCEPTION &&
+        "failed to define .groups property inside indices array.");
+  }
+
+  // Add indices array to result.
+  auto executionStatus = JSObject::defineNewOwnProperty(
+      matchObj,
+      runtime,
+      Predefined::getSymbolID(Predefined::indices),
+      PropertyFlags::defaultNewNamedPropertyFlags(),
```
fbmal7, line 516: Ditto-`getDefaultNewPropertyFlags`.

lib/VM/JSLib/RegExp.cpp
```diff
@@ -595,6 +658,13 @@ CallResult<Handle<JSArray>> directRegExpExec(
   auto inputSHV = SmallHermesValue::encodeStringValue(*S, runtime);
   JSObject::setNamedSlotValueUnsafe(*A, runtime, inputDesc, inputSHV);
 
+  // Create indices array.
+  auto indicesRes = JSArray::create(runtime, match.size(), match.size());
```
fbmal7, line 662: Why do you create this unconditionally? This will only ever be used if `hasIndices` is true, right?

utils/testsuite/testsuite_skiplist.py
```diff
@@ -2051,7 +2051,6 @@
     "import-assertions",
     "json-superset",
     "new.target",
-    "regexp-match-indices",
```
fbmal7, line 2054: This is good to test to spec, but it should not be the only source of testing. Add some examples of using the `d` flag and making sure it's working in some basic tests cases in our lit tests. You can add these cases to `test/hermes/regexp.js`. For instance, I haven't had time to debug this further, but I am getting an error when running this code on top of your changes:

```
let re = /(?<keyword>let|const|var)\s+(?<id>[a-zA-Z_$][0-9a-zA-Z_$]*)/d;
let match = re.exec("let myVar");
print((match.indices.groups));
```
Something funny is happening here since if I instead change the line to the following:
```
print(JSON.stringify((match.indices.groups)));
```
everything works fine.

lib/VM/JSLib/RegExp.cpp
```diff
@@ -460,6 +461,65 @@ static void createGroupsObject(
   JSObject::setNamedSlotValueUnsafe(matchObj.get(), runtime, groupsDesc, shv);
 }
 
+static void createIndicesArray(
+    Runtime &runtime,
+    Handle<JSArray> matchObj,
+    Handle<JSObject> mappingObj,
+    Handle<JSArray> indices) {
+  // If there are no capture groups, then set groups to undefined.
+  if (!mappingObj) {
```
fbmal7, line 470: small nit- is there a point to flipping this conditional if both cases are written out?

fbmal7:
Overall the code is looking good so far, thanks for taking this on :)

lib/VM/JSLib/RegExp.cpp
fabioh8010: I tried to do this, but since my `createIndicesArray` is a `void` method this approach wouldn't work, so I followed the same way was done to `createGroupsObject` which is also a `void` method and uses `assert`. What's your advice? Change the method's signature to return something like `CallResult<void>` ?

lib/VM/JSLib/RegExp.cpp
fabioh8010: I created at this point because I need this array already declared in order to be used inside the loop [here](https://github.com/fabioh8010/hermes/blob/feature/regexp-indices-flag/lib/VM/JSLib/RegExp.cpp#L691), and would also need the same array outside the loop [here](https://github.com/fabioh8010/hermes/blob/feature/regexp-indices-flag/lib/VM/JSLib/RegExp.cpp#L724). Do you think it's problematic?

lib/VM/JSLib/RegExp.cpp
fabioh8010: It was just preference I guess, I can flip it if you like :)

utils/testsuite/testsuite_skiplist.py
fabioh8010: Yes I was going to say that, this behavior also happens to `match.groups` because it's a null-prototype object just like `match.indices.groups` is.

This issue also happens in other engines if you do like this for example (although they'll work if you just `console.log` the variable without string conversion):

```javascript
let re = /(?<keyword>let|const|var)\s+(?<id>[a-zA-Z_$][0-9a-zA-Z_$]*)/d;
let match = re.exec("let myVar");
console.log('match.groups ' + match.groups); // Uncaught TypeError: Cannot convert object to primitive value
console.log('match.indices.groups ' + match.indices.groups); // Uncaught TypeError: Cannot convert object to primitive value
```

Some references:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#return_value
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#null-prototype_objects

fabioh8010: About the testing, I'll implement some tests as requested 👍 . Should I implement tests for this "null-prototype object" case too?

lib/VM/JSLib/RegExp.cpp
fbmal7: You change the type to `ExecutionStatus`.

utils/testsuite/testsuite_skiplist.py
fbmal7: Yeah I don't think I would classify this as a bug. JSC and SpiderMonkey both exhibit the same behavior as Hermes in this case. You can test this by trying to print out the expression `Object.create(null)` in the different engines.

You should not write tests for this case.

lib/VM/JSLib/RegExp.cpp
fbmal7: Again it's a nit, but I personally feel like it would be cleaner without the negation.

fabioh8010:
Hi @fbmal7, I addressed all the changes requested.

fbmal7:
Actually, I think this code could be cleaner if we consolidate the all the work related to populating the indices object into one method. This way we can also make the code read closer to the spec. We can also address the point about eagerly initializing `auto indicesRes = JSArray::create(runtime, match.size(), match.size());`. So, you can remove the check for `hasIndices` inside the loop (which would have been turned into 3 hasIndices checks to avoid the initialization.)

And instead, move all that logic into a new method `makeMatchIndicesIndexPairArray`, which will be aligned to [the spec](https://tc39.es/ecma262/#sec-makematchindicesindexpairarray) with the same name. Then we will only have one check for `hasIndices`. This means you will construct and populate `Handle<JSArray> indices` inside of `makeMatchIndicesIndexPairArray`. So it will also need to take in `match` as a parameter. Also, try to include inline comments in source code for the corresponding spec rules. I will leave a comment on a specific piece of code as an example.

lib/VM/JSLib/RegExp.cpp
```diff
@@ -460,6 +461,67 @@ static void createGroupsObject(
...
+            JSObject::defineOwnProperty(
+                indices,
+                runtime,
+                Predefined::getSymbolID(Predefined::groups),
+                DefinePropertyFlags::getDefaultNewPropertyFlags(),
+                groupsObj) == ExecutionStatus::EXCEPTION)) {
+      return ExecutionStatus::EXCEPTION;
+    }
+  } else {
+    // If there are no capture groups, then set groups to undefined.
```
fbmal7, line 499: Here, you can leave an inline comment that copy-pastes the spec. In this case it's coming from this [definition](https://tc39.es/ecma262/#sec-makematchindicesindexpairarray).

```
// 7a. Else, Let groups be undefined.
if (LLVM_UNLIKELY(
            JSObject::defineOwnProperty(
                indices,
                runtime,
                Predefined::getSymbolID(Predefined::groups),
                DefinePropertyFlags::getDefaultNewPropertyFlags(),
                Runtime::getUndefinedValue()) == ExecutionStatus::EXCEPTION)) {
      return ExecutionStatus::EXCEPTION;
    }
```

fabioh8010:
Hi @fbmal7 thanks for your review and thoughts. Yes, I agree, let's keep the implementation closer to the specification 👍 
I will be OOO for some days and will return on May 4th to start working on these changes.

fabioh8010:
Hi @fbmal7 , I addressed the requested changes, tried to keep the implementation as close as possible to the specification, please review it again when you have some time.

lib/VM/JSLib/RegExp.cpp
```diff
@@ -595,6 +701,10 @@ CallResult<Handle<JSArray>> directRegExpExec(
   auto inputSHV = SmallHermesValue::encodeStringValue(*S, runtime);
   JSObject::setNamedSlotValueUnsafe(*A, runtime, inputDesc, inputSHV);
 
+  // If R contains any GroupName, then let hasGroups be true.
+  auto groupNames = regexp->getGroupNameMappings(runtime);
```
fbmal7, line 705: You can write out the type here, it's not obvious from the name.

`Handle<JSObject> groupNames = regexp->getGroupNameMappings(runtime);`

lib/VM/JSLib/RegExp.cpp
```diff
@@ -460,6 +461,108 @@ static void createGroupsObject(
...
+    idx++;
+  }
+
+  // If hasGroups is true, then
+  if (hasGroups) {
+    // Let groups be OrdinaryObjectCreate(null).
+    auto clazzHandle = runtime.makeHandle(mappingObj->getClass(runtime));
+    auto groupsObjRes = JSObject::create(
+        runtime, Runtime::makeNullHandle<JSObject>(), clazzHandle);
+    auto groupsObj = runtime.makeHandle(groupsObjRes.get());
```
fbmal7, line 522: nit: `groupsObj` -> `groups`

fabioh8010:
Hi @fbmal7 , thanks for the review. I've addressed all the changes, but I also had to use `JSArray::setStorageEndIndex()` in my arrays to be able to work with `JSArray::unsafeSetExistingElementAt()`, because without resizing the storage I was getting this kind of error.

```
Assertion failed: (index >= self->beginIndex_ && index < self->endIndex_ && "array index out of range"), function unsafeSetExistingElementAt, file JSArray.h, line 77.
```

Please let me know if it makes sense, thank you!

fbmal7:
Using `setStorageEndIndex` is expected.

Given that this PR will require a bytecode bump, the team is currently discussing the best way to merge these changes in.

AndrewGable:
Hey @fbmal7 - We are trying to plan around this PR, would this be included in a new version of Hermes (`0.12.1` ?) that would land with `0.73.x` of React Native? Thanks in advanced!

fbmal7:
@AndrewGable We expect this PR to land in time for 0.73, barring any surprises. We will keep this thread updated.

PR #1000, opened 2023-05-18
feat: Add support for tvOS (Apple TV)

douglowder:
## Summary

Add support for tvOS, so that Hermes will work with React Native for Apple TV.

## Test Plan

New test target `ApplePlatformsIntegrationTVTests` in the integration test app for Apple platforms, similar to the existing ones for iOS and MacOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,14 @@ if(HERMES_APPLE_TARGET_PLATFORM MATCHES "catalyst")
   set(THREADS_PREFER_PTHREAD_FLAG ON)
 endif()
 
+if(HERMES_APPLE_TARGET_PLATFORM MATCHES "appletvos")
+  add_definitions(-DTARGET_OS_TV=1)
+endif()
+
+if(HERMES_APPLE_TARGET_PLATFORM MATCHES "appletvsimulator")
+  add_definitions(-DTARGET_OS_TV=1)
+endif()
+
 # This must be consistent with the release_version in:
 # - android/build.gradle
 # - npm/package.json
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,6 +57,8 @@ if(HERMES_APPLE_TARGET_PLATFORM MATCHES "catalyst")
   set(THREADS_PREFER_PTHREAD_FLAG ON)
 endif()
 
+# The build for Apple platforms is not automatically picking up TARGET_OS_TV from TargetConditionals.h
+# for Apple TV. For now, we add this define here.
 if(HERMES_APPLE_TARGET_PLATFORM MATCHES "appletvos")
   add_definitions(-DTARGET_OS_TV=1)
 endif()
--- a/external/icu_decls/unicode/platform.h
+++ b/external/icu_decls/unicode/platform.h
@@ -148,8 +148,6 @@
 #   include <TargetConditionals.h>
 #   if defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE  /* variant of TARGET_OS_MAC */
 #       define U_PLATFORM U_PF_IPHONE
-#   elif defined(TARGET_OS_TV) && TARGET_OS_TV  /* variant of TARGET_OS_MAC */
-#       define U_PLATFORM U_PF_IPHONE
 #   else
 #       define U_PLATFORM U_PF_DARWIN
 #   endif
--- a/external/icu_decls/unicode/platform.h
+++ b/external/icu_decls/unicode/platform.h
@@ -148,6 +148,8 @@
 #   include <TargetConditionals.h>
 #   if defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE  /* variant of TARGET_OS_MAC */
 #       define U_PLATFORM U_PF_IPHONE
+#   elif defined(TARGET_OS_TV) && TARGET_OS_TV  /* variant of TARGET_OS_MAC */
+#       define U_PLATFORM U_PF_IPHONE
 #   else
 #       define U_PLATFORM U_PF_DARWIN
 #   endif
--- a/external/llvh/lib/Support/Unix/Program.inc
+++ b/external/llvh/lib/Support/Unix/Program.inc
@@ -200,7 +200,7 @@ static bool Execute(ProcessInfo &PI, StringRef Program,
   // If this OS has posix_spawn and there is no memory limit being implied, use
   // posix_spawn.  It is more efficient than fork/exec.
 #ifdef HAVE_POSIX_SPAWN
-#if defined(TARGET_OS_TV) && TARGET_OS_TV
+#if defined(__APPLE__) && defined(TARGET_OS_TV) && TARGET_OS_TV
   {
 #else
   if (MemoryLimit == 0) {
@@ -270,7 +270,7 @@ static bool Execute(ProcessInfo &PI, StringRef Program,
   }
 #endif
 
-#if !(defined(TARGET_OS_TV) && TARGET_OS_TV)
+#if !(defined(__APPLE__) && defined(TARGET_OS_TV) && TARGET_OS_TV)
   // Create a child process.
   int child = fork();
   switch (child) {
@@ -329,8 +329,8 @@ static bool Execute(ProcessInfo &PI, StringRef Program,
   PI.Pid = child;
   PI.Process = child;
 
-  return true;
 #endif // TARGET_OS_TV
+  return true;
 }
 
 namespace llvh {
--- a/external/llvh/lib/Support/Unix/Program.inc
+++ b/external/llvh/lib/Support/Unix/Program.inc
@@ -47,7 +47,7 @@
 #include <TargetConditionals.h>
 #endif
 
-#if defined(__APPLE__) && !(defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE)
+#if defined(__APPLE__) && !(defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE) && !(defined(TARGET_OS_TV) && TARGET_OS_TV)
 #define USE_NSGETENVIRON 1
 #else
 #define USE_NSGETENVIRON 0
@@ -200,7 +200,11 @@ static bool Execute(ProcessInfo &PI, StringRef Program,
   // If this OS has posix_spawn and there is no memory limit being implied, use
   // posix_spawn.  It is more efficient than fork/exec.
 #ifdef HAVE_POSIX_SPAWN
+#if defined(TARGET_OS_TV) && TARGET_OS_TV
+  {
+#else
   if (MemoryLimit == 0) {
+#endif
     posix_spawn_file_actions_t FileActionsStore;
     posix_spawn_file_actions_t *FileActions = nullptr;
 
@@ -266,6 +270,7 @@ static bool Execute(ProcessInfo &PI, StringRef Program,
   }
 #endif
 
+#if !(defined(TARGET_OS_TV) && TARGET_OS_TV)
   // Create a child process.
   int child = fork();
   switch (child) {
@@ -325,6 +330,7 @@ static bool Execute(ProcessInfo &PI, StringRef Program,
   PI.Process = child;
 
   return true;
+#endif // TARGET_OS_TV
 }
 
 namespace llvh {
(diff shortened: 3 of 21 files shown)


tmikov:
Thank you! We will review this PR in the next few days. Please keep in mind that even if we merge it, we won't be able to provide actual support (answer questions, provide fixes, etc) for Hermes on tvOS, we will have to rely on the community.

douglowder:
Thanks for reviewing this, @tmikov and the rest of the team :)

I would very much appreciate you merging this so that I can use the Hermes engine as is in the React Native TV repo, and not have to create a fork.

Needless to say, if there are issues, they will probably be coming from people using the React Native TV repo, so I will definitely be available to help with such issues if they arise.

douglowder:
Hmm it seems like there is something weird going on in Circle CI -- the same `xcodebuild test` invocation that works fine on my desktop can't find an Apple TV simulator in the Circle CI environment.

neildhar:
Thank you for contributing this, looking at the PR, some high level comments and questions:

1. The iOS build for Hermes has actually been migrated to RN. So changes to Hermes source and our CMake files should be made here, but things like `build-ios-framework` and `build-apple-framework` are now maintained by RN (we will delete them from Hermes in the near future). Correspondingly, these should live wherever React Native TV is.
2. It looks like many of the changes here are to build tools like the `hermes` CLI, and `node-hermes`. In general, building these should not be necessary to use Hermes in RN on tvOS, you should just need to build the `libhermes` target. Is there a reason your PR addresses these other targets?

I also have some questions about specific lines, but these can wait since the above two points may significantly reduce the scope of necessary changes.

douglowder:
> The iOS build for Hermes has actually been migrated to RN. So changes to Hermes source and our CMake files should be made here, but things like build-ios-framework and build-apple-framework are now maintained by RN (we will delete them from Hermes in the near future). Correspondingly, these should live wherever React Native TV is.

@neildhar thanks very much for letting me know about the above, and I will make the corresponding changes to those files in the React Native TV repo.

> It looks like many of the changes here are to build tools like the hermes CLI, and node-hermes. In general, building these should not be necessary to use Hermes in RN on tvOS, you should just need to build the libhermes target. Is there a reason your PR addresses these other targets?

It appeared to me that all these changes were needed in order for `build-ios-framework.sh` to complete successfully for tvOS. If you could identify which changes are specifically for building the CLI and node-hermes, I'll try removing them from the PR and see if everything is still ok. The difficulty with tvOS is that it is almost iOS but not quite -- certain APIs (even fundamental ones like `fork()`) are removed and will not compile if they are left in the source by the preprocessor.

neildhar:
> prevent regressions that would otherwise only be detected downstream in the React Native repo?

This is a great question. Empirically, this setup has been much more reliable. In the past, we've had problems at the point of producing the Hermes CocoaPod, and consuming it on the RN side. Having that process subsumed into the RN build has eliminated that class of problems entirely.
But to make sure that we don't break the RN build, we need to add a simple end-to-end test in our CI that clones the RN repo and builds RNTester against Hermes source from our repo. We already have such a [test for Android](https://github.com/facebook/hermes/blob/473727fa808595ad828a953d46d627f35e76ee99/.circleci/config.yml#L531-L562), and it's a matter of setting up something similar for iOS. Once we have that, we can eliminate the existing iOS testing, and the build scripts in the Hermes repo, in favour of the end-to-end approach.

> which changes are specifically for building the CLI and node-hermes

At the very least, the changes in the node-hermes directory should not be needed. I tried tinkering with some of your changes locally and I do see it complaining about certain system calls. I'd like to give some more thought to how best to address these.

douglowder:
@neildhar I think you absolutely need the changes I put in `node-hermes` directory. They are to exclude code from Apple TV builds (`!TARGET_OS_TV`), in the same way that the existing preprocessor directives are excluding code from iOS builds (`!TARGET_OS_IPHONE`)

CMakeLists.txt
```diff
@@ -57,6 +57,14 @@ if(HERMES_APPLE_TARGET_PLATFORM MATCHES "catalyst")
   set(THREADS_PREFER_PTHREAD_FLAG ON)
 endif()
 
+if(HERMES_APPLE_TARGET_PLATFORM MATCHES "appletvos")
+  add_definitions(-DTARGET_OS_TV=1)
```
neildhar, line 61: Why do we need to define this ourselves, that is, why doesn't it defined automatically as it is for iOS?

neildhar (commented):
@douglowder nothing in the `node-hermes` directory gets built when preparing the Apple frameworks. The library you modified is a third-party library, and had those iOS ifdefs to begin with, but we don't actually build it for iOS in Hermes.

To verify, I checked out your changes and was able to complete a successful build without any of the changes in that directory.

CMakeLists.txt
douglowder: I don't know the answer to that question..... I just found that it was needed.

douglowder: I'll back out the node-hermes changes and see if CI succeeds and has the tvOS artifacts.

external/icu_decls/unicode/platform.h
```diff
@@ -148,6 +148,8 @@
 #   include <TargetConditionals.h>
 #   if defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE  /* variant of TARGET_OS_MAC */
 #       define U_PLATFORM U_PF_IPHONE
+#   elif defined(TARGET_OS_TV) && TARGET_OS_TV  /* variant of TARGET_OS_MAC */
```
neildhar, line 151: It looks like this header is only used on Windows, why is this change needed?

external/llvh/lib/Support/Unix/Program.inc
```diff
@@ -200,7 +200,11 @@ static bool Execute(ProcessInfo &PI, StringRef Program,
   // If this OS has posix_spawn and there is no memory limit being implied, use
   // posix_spawn.  It is more efficient than fork/exec.
 #ifdef HAVE_POSIX_SPAWN
+#if defined(TARGET_OS_TV) && TARGET_OS_TV
```
neildhar, line 203: What is the issue being worked around here?

lib/Support/OSCompatPosix.cpp
```diff
@@ -778,10 +778,12 @@ bool unset_env(const char *name) {
 void *SigAltStackLeakSuppressor::stackRoot_{nullptr};
 
 SigAltStackLeakSuppressor::~SigAltStackLeakSuppressor() {
+#if !(defined(TARGET_OS_TV) && TARGET_OS_TV)
```
neildhar, line 781: It seems a little safer (and more common) to use something like `#if defined(__APPLE__) && TARGET_OS_TV`. That way, forgetting to include `TargetConditionals.h` will result in a preprocessor error. In practice, this would matter most for code that actually can compile on the target platform, but it seems good to be consistent.

external/llvh/lib/Support/Unix/Program.inc
```diff
@@ -266,6 +270,7 @@ static bool Execute(ProcessInfo &PI, StringRef Program,
   }
 #endif
 
+#if !(defined(TARGET_OS_TV) && TARGET_OS_TV)
```
neildhar, line 273: I'm not sure this is the right approach. ifdef-ing out part of this function is risky, we now have a control path through the function that doesn't return anything.

If we do need to compile this, I would not be opposed to ifdef-ing out the whole body on tvOS, and using `llvh::report_fatal_error` to trigger a runtime crash.

CMakeLists.txt
neildhar: I think it would be great to understand what's going on with this. Based on some cursory searches, it looks like the tvOS define should be in `TargetConditionals.h` just like the iOS one. This may suggest something is wrong with how the build is set up, or warrant a comment to avoid confusion in the future.

neildhar (commented):
Thank you for making the changes to the node-hermes directory. I've done a pass over the individual ifdefs to get your input.

external/llvh/lib/Support/Unix/Program.inc
```diff
@@ -325,6 +330,7 @@ static bool Execute(ProcessInfo &PI, StringRef Program,
   PI.Process = child;
 
   return true;
+#endif // TARGET_OS_TV
```
douglowder, line 333:
```suggestion
#endif // TARGET_OS_TV
  return true;
```

external/llvh/lib/Support/Unix/Program.inc
douglowder: Apple TV needs to use posix_spawn (does not have fork()) but MemoryLimit did not exist for Apple TV (at least at the time I was working on this patch).

CMakeLists.txt
douglowder: I will have a look at `TargetConditionals.h`

external/llvh/lib/Support/Unix/Program.inc
neildhar: What do you mean? MemoryLimit is a parameter to this function.

CMakeLists.txt
douglowder: I am not able to determine why the build is not picking up the tvOS version of `TargetConditionals.h`. I will add a comment here and hopefully we can figure that out in the future.

external/icu_decls/unicode/platform.h
douglowder: I will back out this change and make sure tvOS still builds. It appeared to me, in reading this file, that it was not only used on Windows, since it referenced Apple defines.

external/llvh/lib/Support/Unix/Program.inc
douglowder: Sorry I didn't say that correctly. Apple TV needs to use this path even if MemoryLimit is not 0.

external/llvh/lib/Support/Unix/Program.inc
douglowder: Done

CMakeLists.txt
neildhar: I apologise for the back and forth on this. But we can't merge changes that we don't really understand. In this instance, it may suggest a subtle issue in the build setup that could break in the future, or add overhead when managing the integration of our Apple builds with upstream RN.

lib/Support/OSCompatPosix.cpp
```diff
@@ -778,10 +778,12 @@ bool unset_env(const char *name) {
 void *SigAltStackLeakSuppressor::stackRoot_{nullptr};
 
 SigAltStackLeakSuppressor::~SigAltStackLeakSuppressor() {
+#if defined(__APPLE__) && !(defined(TARGET_OS_TV) && TARGET_OS_TV)
```
neildhar, line 781: My suggestion was to use just `defined(__APPLE__) && !TARGET_OS_TV`. As noted in my previous comment, doing this correctly also ensures that `TargetConditionals.h` is properly imported.

In addition to the suggested change above, we should also add a comment that this is not supported on tvOS.

external/llvh/lib/Support/Unix/Program.inc
```diff
@@ -200,7 +200,11 @@ static bool Execute(ProcessInfo &PI, StringRef Program,
   // If this OS has posix_spawn and there is no memory limit being implied, use
   // posix_spawn.  It is more efficient than fork/exec.
 #ifdef HAVE_POSIX_SPAWN
+#if defined(__APPLE__) && defined(TARGET_OS_TV) && TARGET_OS_TV
```
neildhar, line 203: Thank you for explaining what was going on with `MemoryLimit` here. I think the right thing to do here would be to restrict the ifdef to just the region below with the `fork`, and then use `ErrMsg` to report an error if we're on tvOS and a `MemoryLimit` is specified. I suspect something similar would happen on iOS anyway, since I don't think we're allowed to actually call `fork`.

neildhar (changes requested):

CMakeLists.txt
douglowder: I totally understand. I'll close this for now, and once the React Native TV 0.71 release is out, will come back to this and see what is going on with TargetConditionals. My guess is that somewhere in the Hermes build, the Apple CMake configuration is picking up iPhone includes instead of Apple TV includes.

Issue #926, opened 2023-02-27
Add support for dateStyle and timeStyle for Android


dlebedynskyi:
## Problem

Intl on Android is still missing dateStyle and timeStyle. iOS already has it [since v0.12](https://github.com/facebook/hermes/blob/v0.12.0/lib/Platform/Intl/PlatformIntlApple.mm#L517).

## Solution

Ask is to add support for missing [Intl.DateTimeFormat API for Android](https://github.com/facebook/hermes/blob/main/doc/IntlAPIs.md)  [dateStyle/timeStyle](https://tc39.es/proposal-intl-datetime-style/).

## Additional Context

If possible, would be great to add rest of missing API 
* [dayPeriod](https://github.com/tc39/ecma402/issues/29)
* [fractionalSecondDigits](https://github.com/tc39/ecma402/pull/347)

to close rest of missing options on Intl.DateTimeFormat  as polyfill for Intl.DateTimeFormat is complex and heavy, and is currently required when dateStyle/timeStyle are used. 

Related https://github.com/facebook/hermes/issues/23#issuecomment-1446878656

mganandraj:
I've created a PR : https://github.com/facebook/hermes/pull/927

@neildhar

dlebedynskyi:
@mganandraj  sorry for late reply. 

We think there is a bug in how it is implemented. 

The way it is done, it does not match the spec - https://tc39.es/ecma402/#datetimeformat-objects. 

```
`If required is "date" or "any", then
a. For each property name prop of « "weekday", "year", "month", "day" », do
i. Let value be ? Get(options, prop).
ii. If value is not undefined, let needDefaults be false.
```

as example 
```
const date = new Date(Date.UTC(2012, 11, 20, 3, 0, 0, 200));
// request a weekday along with a long date
let options = {
  weekday: "narrow"
};
console.log(new Intl.DateTimeFormat("en-US", options).format(date)); 
// outputs "W, 12/19/2012" which is wrong,
```

this should just be "W" since using "weekday" negates the need for defaults.

Fix that we tested was here in https://github.com/facebook/hermes/blob/main/lib/VM/JSLib/Intl.cpp#L272

```diff
- {u"weekday", platform_intl::Option::Kind::String, 0},
+ {u"weekday", platform_intl::Option::Kind::String, kDateRequired},
```


We had tested your change prior to merge. So maybe we can do PR now. WDYT?

mganandraj:
[...]

PR #1014, opened 2023-05-28
Fix Android Intl to clear needDefault flag correctly

Maturia:
Fixed an issue where needDefault flag was not being cleared when the option `weekday` was set.

## Summary
The current Intl implementation on Android does not correctly follow the RFC for Intl implementation. Specifying a value for the option `weekday`, `year`, `month`, or `day` should clear the `needDefaults` flag, which will prevent the auto-formatted date from automatically being appended to the formatted string (4a here: https://tc39.es/ecma402/#sec-todatetimeoptions).

Current implementation:
```
const date = new Date(Date.UTC(2012, 11, 20, 3, 0, 0, 200));
// request a narrow weekday format
let options = {
  weekday: "narrow"
};
console.log(new Intl.DateTimeFormat("en-US", options).format(date)); 
// outputs "W, 12/19/2012" which is wrong, this should just be "W" since using "weekday" negates the need for defaults.
```

With fix:
```
const date = new Date(Date.UTC(2012, 11, 20, 3, 0, 0, 200));
let options = {
  weekday: "narrow"
};
console.log(new Intl.DateTimeFormat("en-US", options).format(date)); 
// outputs "W" which is correct.
```

### DIFF
--- a/lib/VM/JSLib/Intl.cpp
+++ b/lib/VM/JSLib/Intl.cpp
@@ -269,7 +269,7 @@ constexpr OptionData kDTFOptions[] = {
     {u"hourCycle", platform_intl::Option::Kind::String, 0},
     {u"timeZone", platform_intl::Option::Kind::String, 0},
     {u"formatMatcher", platform_intl::Option::Kind::String, 0},
-    {u"weekday", platform_intl::Option::Kind::String, 0},
+    {u"weekday", platform_intl::Option::Kind::String, kDateRequired},
     {u"era", platform_intl::Option::Kind::String, 0},
     {u"year",
      platform_intl::Option::Kind::String,


neildhar:
Hey @Maturia, thanks for reporting and investigating this. I have a question regarding the test case you've shared. While this seems to be a legitimate bug, running `new Intl.DateTimeFormat("en-US", options).format(date)` does not exercise the code modified here, and seems to work correctly. However, I am able to reproduce the bug you describe by running `console.log(date.toLocaleString('en-US', options));`.

Can you confirm the issue you're seeing?

dlebedynskyi:
I believe issues in this PR is same from https://github.com/facebook/hermes/issues/926#issuecomment-1535461210

neildhar:
@dlebedynskyi I understand the issue being reported here. This fixes a legitimate bug in Hermes. I'm just confused because the example code deals with `DateTimeFormat`, when the bug should appear with `toLocaleString`. So it would be helpful to get confirmation that this is indeed the bug you're observing, and you aren't using any polyfills perhaps that might in turn delegate the work to `toLocaleString`.

dlebedynskyi:
I've done some quick test on 0.71 demo app 
```
date = new Date(Date.UTC(2012, 11, 20, 3, 0, 0, 200));
options = {
  weekday: "narrow"
};
new Intl.DateTimeFormat("en-US", options).format(date);
// 'W' - Correct 

console.log(date.toLocaleString('en-US', options))
// W, 12/19/2012, 7:00:00 PM  - WRONG, expected W instead
```

@neildhar  looks like you are correct and fix should be in `toLocaleString` part 
@Maturia  Can you please update fix?

neildhar:
@dlebedynskyi Thank you for confirming. This fix is correct for the bug in `toLocaleString`, I just wanted to make sure we were looking at the same problem.

@Maturia could you update the summary and add a test for this? The easiest place to do it would be to add it to `date-time-format-apple.js`, which will test that we get the right behaviour on macOS (the bug is also present there).

PR #1038, opened 2023-06-26
Flow API Translator - Fix variables that have identifiers as values

nmn:
Summary: ...to Use `typeof` for values.

Differential Revision: D46993719


nmn:
I'm not sure how to fix the failing tests here. I'll need some help.

NickGerleman:
Failure looks unrelated. Maybe the OSS CI here needs some love?

noahlemen:
i suspect this is an OSS CI issue too – definitely looks unrelated. i've only worked on this repo through internal phabricator and have not seem a similar failure there

PR #1040, opened 2023-06-27
Add missing cstdint includes

theoparis:
I added the missing <cstdint> includes to fix build errors when using uint*_t with clang 16. I also added .cache to gitignore so I don't acidentally commit the clangd LSP cache directory.

The errors that this PR fixes look like this:
```
          ^
/home/theo/dev/hermes/public/hermes/Public/../../hermes/Public/DebuggerTypes.h:40:11: error: unknown type name 'uint64_t'
constexpr uint64_t kInvalidBreakpoint = 0;
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ android/intltest/java/com/facebook/hermes/test/assets/**
 # VS Code and Visual Studio
 .vs
 .vscode
+
+# clangd cache
+.cache/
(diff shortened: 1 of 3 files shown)


theoparis:
It seems like the CI (test-e2e) is broken due to gradle toolchains 👀

tmikov:
These changes certainly look valid.

PR #1058, opened 2023-07-18
Implement some more functions

neildhar:
Differential Revision: D47564304


API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,270 @@
...
+
+#ifndef HERMES_ABI_HERMES_ABI_H
+#define HERMES_ABI_HERMES_ABI_H
+
+#include <assert.h>
+#include <stdalign.h>
+#include <stddef.h>
+#include <stdint.h>
+
+#ifndef HERMES_EXPORT
```
vmoroz, line 16: It is better to optimize the header file for users, not for the implementation.
We better reverse the logic here and not to define the HERMES_EXPORT unless we have another definition that says that we are implementing the API rather than consuming it.

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,270 @@
...
+
+/// Define the structure for references to pointer types in JS (e.g. string,
+/// object, BigInt).
+/// TODO: Replace jsi::PointerValue itself with this C implementation to
+/// eliminate pointer management overhead in the JSI wrapper.
+struct HermesABIManagedPointer;
+struct HermesABIManagedPointerVTable {
+  /// Pointer to the function that should be invoked when this reference is
+  /// released.
+  void (*invalidate)(HermesABIManagedPointer *);
```
vmoroz, line 36: Please add explicit calling conventions to all functions and function pointers.
We technically need it only for MSVC x86.
The best way to do it is to define a macro, and then use it everywhere. E.g. see the `NAPI_CDECL` in [Node-API](https://github.com/nodejs/node/blob/main/src/js_native_api_types.h).

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,270 @@
...
+#endif // _MSC_VER
+#endif // !defined(HERMES_EXPORT)
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// Define the structure for references to pointer types in JS (e.g. string,
+/// object, BigInt).
+/// TODO: Replace jsi::PointerValue itself with this C implementation to
```
vmoroz, line 30: In my experiments I saw that we can `reinterpret_cast` C++ `jsi::PointerValue` to a struct that has `HermesABIManagedPointerVTable` pointer as long as we specify explicit calling conventions.

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,270 @@
...
+  HermesABIErrorCodeJSINativeException,
+  HermesABIErrorCodeJSError,
+};
+
+static_assert(
+    alignof(HermesABIManagedPointer) % 4 == 0,
+    "HermesABIManagedPointer must be at least aligned to pointer size.");
+
+/// Define simple wrappers for the different pointer types.
+#define HERMES_ABI_POINTER_TYPES(V) \
```
vmoroz, line 52: While the macros are easier to write, they are very difficult to debug and to search for symbols.
It would be better to write all these definitions explicitly.
The API is usually written once and read many times.
Being explicit here significantly increases readability of the code.

API/hermes_abi/HermesABIHelpers.h
```diff
@@ -0,0 +1,197 @@
...
+      HermesABIErrorCode err) {                                            \
+    HermesABI##name##OrError res;                                          \
+    res.is_error = true;                                                   \
+    res.data.error = err;                                                  \
+    return res;                                                            \
+  }
+HERMES_ABI_TRIVIAL_OR_ERROR_TYPES(DECLARE_HERMES_ABI_TRIVIAL_OR_ERROR_HELPERS)
+#undef DECLARE_HERMES_ABI_TRIVIAL_OR_ERROR_HELPERS
+
+static inline HermesABIVoidOrError createVoidOrError(void) {
```
vmoroz, line 47: Header files must never use `static` keyword for standalone functions.
It results in extremely inefficient generated code because linker is required to create a copy of each such function per .cpp file.

API/hermes_abi/HermesABIHelpers.h
```diff
@@ -0,0 +1,197 @@
...
+
+#ifndef HERMES_ABI_HERMES_ABI_HELPERS_H
+#define HERMES_ABI_HERMES_ABI_HELPERS_H
+
+#include "hermes_abi/hermes_abi.h"
+
+namespace facebook::hermes::abi {
+
+#define DECLARE_HERMES_ABI_POINTER_HELPERS(name)                             \
+  static inline HermesABI##name create##name(HermesABIManagedPointer *ptr) { \
```
vmoroz, line 16: It is better to keep the C-API header clean from any implementation code.
If such code is needed for C++ implementation, then it is better to put it in a separate C++ header which includes the C-API header.
We must be able to use the C API from any other language.

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,270 @@
...
+  V(Array)                          \
+  V(String)                         \
+  V(BigInt)                         \
+  V(Symbol)                         \
+  V(Function)                       \
+  V(ArrayBuffer)                    \
+  V(PropNameID)
+
+#define DECLARE_HERMES_ABI_POINTER_TYPE(name) \
+  struct HermesABI##name {                    \
```
vmoroz, line 63: It should be `typedef struct`

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,270 @@
...
+struct HermesABIManagedPointerVTable {
+  /// Pointer to the function that should be invoked when this reference is
+  /// released.
+  void (*invalidate)(HermesABIManagedPointer *);
+};
+struct HermesABIManagedPointer {
+  const HermesABIManagedPointerVTable *vtable;
+};
+
+enum HermesABIErrorCode {
```
vmoroz, line 42: It should be `typedef enum`

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,270 @@
...
+HERMES_ABI_TRIVIAL_OR_ERROR_TYPES(DECLARE_HERMES_ABI_TRIVIAL_OR_ERROR_TYPES)
+#undef DECLARE_HERMES_ABI_TRIVIAL_OR_ERROR_TYPES
+
+struct HermesABIVoidOrError {
+  bool is_error;
+  uint16_t error;
+};
+
+/// Always set the top bit for pointers so they can be easily checked.
+#define HERMES_ABI_POINTER_MASK (1 << (sizeof(int) * 8 - 1))
```
vmoroz, line 97: I wonder why `sizeof(int)`. Since it is related to a pointer, it should be `sizeof(intptr_t)`.

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,270 @@
...
+  } data;
+};
+
+/// Define the structure for buffers containing JS source or bytecode. This is
+/// designed to fully mirror the functionality of jsi::Buffer.
+struct HermesABIBuffer;
+struct HermesABIBufferVTable {
+  size_t (*size)(const HermesABIBuffer *);
+  const uint8_t *(*data)(const HermesABIBuffer *);
+  void (*release)(HermesABIBuffer *);
```
vmoroz, line 127: Should we always use `release` as the first entry in v-table for consistency?

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,270 @@
...
+      *clone_prop_name_id)(HermesABIContext *, HermesABIPropNameID);
+  HermesABIStringOrError (*clone_string)(HermesABIContext *, HermesABIString);
+  HermesABISymbolOrError (*clone_symbol)(HermesABIContext *, HermesABISymbol);
+  HermesABIObjectOrError (*clone_object)(HermesABIContext *, HermesABIObject);
+  HermesABIBigIntOrError (*clone_big_int)(HermesABIContext *, HermesABIBigInt);
+
+  size_t (
+      *get_max_utf8_length_from_string)(HermesABIContext *, HermesABIString);
+  HermesABISizeTOrError (
+      *get_utf8_from_string)(HermesABIContext *, HermesABIString, uint8_t *);
```
vmoroz, line 254: Consider using a different pattern where the same function can return either number of required bytes or fill the target buffer. Otherwise, the `get_utf8_from_prop_name_id` looks quite dangerous because it does not provide the buffer size.
See [napi_get_value_string_utf8](https://github.com/nodejs/node/blob/3af7cfe7d4b55ed88a68d2a04212d72540a4951a/src/js_native_api.h#L171)

API/hermes_abi/hermes_abi.h
neildhar: Yep, I'm aware of this, just haven't gotten around to fixing issues with C compilation (same for `typedef struct`).

API/hermes_abi/hermes_abi.h
neildhar: This API has actually completely changed since this PR was last published. I'm exporting the latest version now and am keen to get your take.

API/hermes_abi/hermes_abi.h
neildhar: While it may work on certain platforms, we certainly shouldn't rely on it. I'm planning to replace `jsi::PointerValue` with `HermesABIManagedPointer` in the future, which should eliminate most of the wrapper overhead.

API/hermes_abi/hermes_abi.h
neildhar: The function list here definitely needs a lot of work (documentation, parameter names, calling conventions)

API/hermes_abi/hermes_abi.h
neildhar: This was only used in one place at the bottom of this header, so I've actually removed the definition for now.

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,400 @@
...
+};
+
+struct HermesABINativeStateVTable {
+  void (*release)(HermesABINativeState *);
+};
+struct HermesABINativeState {
+  const HermesABINativeStateVTable *vtable;
+};
+
+struct HermesABIVTable {
```
vmoroz, line 202: Should we name it HermesABIContextVTable? To match the naming pattern for the code above.

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,400 @@
...
+
+  HermesABIPropNameID (
+      *clone_prop_name_id)(HermesABIContext *, HermesABIPropNameID);
+  HermesABIString (*clone_string)(HermesABIContext *, HermesABIString);
+  HermesABISymbol (*clone_symbol)(HermesABIContext *, HermesABISymbol);
+  HermesABIObject (*clone_object)(HermesABIContext *, HermesABIObject);
+  HermesABIBigInt (*clone_big_int)(HermesABIContext *, HermesABIBigInt);
+
+  /// Check if the given buffer contains Hermes bytecode.
+  bool (*is_hermes_bytecode)(const uint8_t *, size_t);
```
vmoroz, line 236: Should we add the `HermesABIContext *` parameter for consistency even if it is not used?

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,400 @@
...
+  HermesABIVoidOrError (*set_object_property_from_prop_name_id)(
+      HermesABIContext *,
+      HermesABIObject,
+      HermesABIPropNameID,
+      const HermesABIValue *);
+
+  HermesABIArrayOrError (
+      *get_object_property_names)(HermesABIContext *, HermesABIObject);
+
+  HermesABIArrayOrError (*create_array)(HermesABIContext *, size_t);
```
vmoroz, line 285: I wonder how important is the passing of the array size?
In V8 creating an array with a known size produces the most inefficient array representation.

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,400 @@
...
+      size_t);
+  HermesABIPropNameIDOrError (*create_prop_name_id_from_utf8)(
+      HermesABIContext *,
+      const uint8_t *,
+      size_t);
+  HermesABIPropNameIDOrError (
+      *create_prop_name_id_from_string)(HermesABIContext *, HermesABIString);
+  HermesABIPropNameIDOrError (
+      *create_prop_name_id_from_symbol)(HermesABIContext *, HermesABISymbol);
+  bool (*compare_prop_name_id)(
```
vmoroz, line 307: Usually term "compare" is used in C++ when we expect to get -1, 0, or 1.
They use term "equals" to return bool. E.g. [std::equal](https://en.cppreference.com/w/cpp/algorithm/equal)

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,400 @@
...
+  HermesABIVoidOrError (*set_native_state)(
+      HermesABIContext *,
+      HermesABIObject,
+      HermesABINativeState *);
+
+  HermesABIWeakObjectOrError (
+      *create_weak_object)(HermesABIContext *, HermesABIObject);
+  HermesABIValue (*lock_weak_object)(HermesABIContext *, HermesABIWeakObject);
+
+  void (*get_utf8_from_string)(
```
vmoroz, line 355: It is an interesting solution! I see how we can wrap std::string by using it.

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,400 @@
...
+  void (*release)(HermesABIBuffer *);
+};
+struct HermesABIBuffer {
+  const HermesABIBufferVTable *vtable;
+  const uint8_t *data;
+  size_t size;
+};
+
+struct HermesABIByteBufferVTable {
+  void (*grow_by)(HermesABIByteBuffer *, size_t);
```
vmoroz, line 143: What if the buffer cannot grow and we get out of memory error?
We cannot propagate an exception here, but should we return bool or an error code?

API/hermes_abi/hermes_abi.h
vmoroz: Below you use term `strict_equals`.

API/hermes_abi/hermes_abi.h
```diff
@@ -0,0 +1,400 @@
...
+      HermesABIString,
+      HermesABIString);
+  bool (*strict_equals_object)(
+      HermesABIContext *,
+      HermesABIObject,
+      HermesABIObject);
+};
+
+#ifdef _MSC_VER
+__declspec(dllexport)
```
vmoroz, line 390: We need a flag here to avoid exporting if we consume this header from other code.

API/hermes_abi/hermes_abi.h
vmoroz: Thank you for sharing the new version! It looks much cleaner, and you had found some interesting solutions.

API/hermes_abi/hermes_abi.h
neildhar: On the other hand, this is the current naming convention used in JSI.

API/hermes_abi/hermes_abi.h
neildhar: Why?

I'm asking because the existing Hermes and JSI headers always include the visibility specifier.

API/hermes_abi/HermesABIRuntime.cpp
```diff
@@ -0,0 +1,904 @@
...
+    static HermesABIValueOrError call(
+        HermesABIHostFunction *hf,
+        HermesABIContext *ctx,
+        const HermesABIValue *thisArg,
+        const HermesABIValue *args,
+        size_t count) {
+      auto *self = static_cast<HostFunctionWrapper *>(hf);
+      auto &rt = self->rt_;
+      try {
+        std::vector<Value> jsiArgs;
```
vmoroz, line 181: It is better to use `SmallVector` here. The `std::vector` always allocates heap memory.

API/hermes_abi/HermesABIHelpers.h
tmikov: Hmm, I believe usage of "static inline" is pretty common for this case. Given that it is both static and inline, I don't think any modern compiler would emit a copy unless it is actually used.

API/hermes_abi/hermes_abi.h
neildhar: I considered it, and the API does permit allocation failure  (you just wouldn't update `available` if the allocation fails). 

At the same time, correctly detecting and handling OOMs is extremely difficult. Even more so when we're using non-exception compiled code as is the case in the ABI implementation. Given that it is going to be rarely useful in practice, I don't think native OOM handling should be a first-class design consideration.

What do you think?

API/hermes_abi/HermesABIRuntime.cpp
neildhar: The set of Hermes source used here is intentionally restricted. It is likely that integrators will want to build this C++ wrapper outside the main Hermes build.

API/hermes_abi/hermes_abi.h
neildhar: Arrays in Hermes have far fewer representations than in V8, so that is less of a concern. The primary purpose of this API is to efficiently and conveniently provide this functionality on `jsi::Runtime`

API/hermes_abi/hermes_abi.h
neildhar: I think not, this certainly isn't the only function in the vtable that doesn't take a `HermesABIContext`, and it seems unreasonable to require creating a context to check if something is bytecode (since that is currently a static function in our C++ API).

API/hermes_abi/HermesABIRuntime.cpp
vmoroz: Yeah, you are right. I did not realize that it is part of the API for the user code.
Then, maybe we can have a hybrid strategy here: use a fixed array until a specified limit and only after that allocate on heap.
The issue is that this code is called for each function call and should avoid the memory allocation if we can.

API/hermes_abi/hermes_abi.h
vmoroz: It maybe not a big deal for Clang, but for MSVC we must be very specific.
The DLL that exports a function uses `__declspec(dllexport)`.
Then, code that imports the function must use `__declspec(dllimport)` or maybe nothing.
Using `__declspec(dllexport)` in user code is an error and linker may complain about it.

API/hermes_abi/hermes_abi.h
vmoroz: I think you have a good point about not updating the `available` field. No need to provide any extra result then.
Please make sure that this case is handled correctly on both ends of the API.

API/hermes_abi/hermes_abi.h
vmoroz: You are changing some names anyway. Using `strict_equals_prop_name_id` will align it with other functions and clean up the semantics by using the verb `equals`.

API/hermes_abi/hermes_abi.h
neildhar: I view this as the list of exported functions for our ABI, rather than as the vtable for `HermesABIContext`. `HermesABIContext` doesn't actually have a vtable, since it is just an opaque pointer.

But that raises the reasonable question of why `HermesABIContext` is meaningfully different than the other types which do define vtables. For most of those types, they are vtables that are fed into the ABI from outside (e.g. `HostObject`), so it is reasonable that integrators may want to create different instances with different vtables. This property is useful in practice to check whether a `HostObject` was in fact created by the current C++ wrapper (see the `isHostObject` implementation).

The only exception is `HermesABIManagedPointer`'s `invalidate` function. For that, it is reasonable that the implementation may provide different wrappers for different types (e.g. `Object` vs `WeakObject`). In addition, given that `jsi::Runtime::PointerValue` can be invalidated using just a pointer to it, if we want to replace it, we have to recreate that behaviour.

API/hermes_abi/HermesABIHelpers.h
neildhar: Could you elaborate on your concerns? These functions are very small and intended to be inlined. As long as the address of these functions is not taken, they will be either inlined and completely eliminated, or they will be deduplicated by ICF.

vmoroz: If they are inlined, then there is no penalty.
We are religiously avoid `static` for standalone C functions in headers because it prevents linker from merging/reusing them. It is the most common cause for the binary code bloat.
In this context having only `inline` keyword will do exactly the same.
The `static` is mostly bad when functions cannot be inlined. The `inline` keyword is just a hint which can be ignored. It is important to have to avoid compiler complaining about duplicated symbols.
See the [Stack overflow discussion](https://stackoverflow.com/questions/780730/c-c-static-function-in-header-file-what-does-it-mean) for some details.

API/hermes_abi/hermes_abi.h
tmikov: I am with @vmoroz on this one

API/hermes_abi/HermesABIHelpers.h
tmikov: @vmoroz having only `inline` would not work here. This is technically a C file, so we should follow the C rules. In C, `inline` alone is not sufficient to guarantee successful compilation; C requires that an out-of-line function is **explicitly defined** in exactly one module. So, we would need to distribute a separate .c file containing instantiations of inline functions. That is not practical.

This is very easy to see with a simple example like this:
```c
#include <stdio.h>

inline int func(void) { return printf("%d\n", 42); }

int main(void) {
    func();
    return 0;
}
```

This example fails to link when compiled with `gcc main.c`, because there is no out-of-line definition for `func()`. The solution is to either provide it, or to use `static inline`, s very common and widely recommended.

Of course this is not a problem when the header is consumed in C++ mode, C++ has different rules, but we can't rely on that, since it is technically a C header.

neildhar: I should clarify that this particular header is a C++ header just containing a bunch of helper functions. I wrote this in C++ so I could namespace them and avoid clunky names and use overloading.

API/hermes_abi/HermesABIRuntime.cpp
neildhar: It is worth keeping in mind that once the representations of ABI values and JSI values are the same, we won't need any intermediate storage here at all.

API/hermes_abi/HermesABIHelpers.h
tmikov: That's a C++ header? Then I am an idiot and nothing I said applies! Damn.

Please, carry on 🤣

API/hermes_abi/README.md
```diff
@@ -0,0 +1,5 @@
+# Experimental C based stable ABI for Hermes
+
+This directory contains ongoing work to develop a stable C-based ABI for Hermes and an accompanying C++ JSI wrapper. It is a work in progress and is not supported for general use.
+
+The goal of this ABI is to allow Hermes to be updated independently of the rest of a React Native application. Note that this does not solve the general problem of ABI stability for RN extensions, since they are still written against the C++ JSI, and consume a C++ `jsi::Runtime` provided by RN.
```
abhishah85, line 5: does that mean that the hermes bytecode will also be made stable at some point so that hermes can be updated independently of React Native app?

neildhar: Great question. Bytecode stability is not currently planned, but is certainly on our mind for the next major version of Hermes. Note that it is not strictly necessary for swapping Hermes, since you can also swap the compiler binary for a newer release.

API/hermes_abi/hermes_abi.h
abhishah85: @neildhar  can you provide an example usage of `HermesABIByteBuffer`? is this managed by hermes or you expect the caller to manage this object?

neildhar: For most Hermes integrators, this entire API will be an implementation detail. The intention is for this to provide a stable ABI boundary, and then for the integrator to use the C++ wrapper in `HermesABIRuntime.cpp` to use it as a normal JSI runtime.

API/hermes_abi/HermesABIHelpers.h
abhishah85: since you have both c and c++ code, maybe use `.hpp` extension for c++ headers and `.h` for c headers?

API/hermes_abi/hermes_abi.h
abhishah85: maybe i should have been more clear in my question. I did not find a factory function that creates and returns a  `HermesABIByteBuffer*` .

neildhar:
This was merged

PR #1071, opened 2023-07-27
Optimize Equality Operations Indifferent To Null & Undefined

sebholl:
## Summary

`null` and `undefined` are two of the most common JavaScript literals, and yet for the `==` and `!=` operations, they can be treated interchangeably. As they are semantically equivalent, it's really a stylistic preference which one is used. Recognizing this indifference however provides opportunities to:

* Normalizing operand literals to provide more opportunities for subsequent optimization passes, like CSE.
* Reduce the number of load constant statements required in byte-code.

The first opportunity is realized in `lib/Optimizer/Scalar/InstSimplify.cpp`, by rewriting any `null` literals in `==` or `!=` operators with `undefined`. Whilst we could have opted to instead rewritten `undefined` to `null`, empirically `undefined` yielded better optimized output as `undefined` is more commonly used in other boiler-plate byte-code (for example, when calling functions).

The second opportunity is realized in the `LoadConstants::runOnFunction` pass in `lib/BCGen/HBC/Passes.cpp`, by replacing `null` or `undefined` literal operands in indifferent`BinaryOperatorInstr` with nearby operand literals from instructions that *do* discriminate.

As shown in the test plan, both of these optimizations yield similar efficiencies but together perform optimally as the second avoids regressions in the first where a `null` suffices in a function without an `undefined`; and the first happens earlier on in the optimization pipeline and so allows other optimizations to kick-in that the second-alone can't achieve.

Note: I did consider pulling the second opportunity into a separate pass outside of `LoadConstants::runOnFunction` but on-balance it felt like performing an additional full pass would be expensive, and the code should not be felt to interfere along the way.

### Clean-Up:

* In `lib/IR/Instrs.cpp`, I introduced a new `bool BinaryOperatorInst::looseToStrictEqualityOperator(...)` helper method.
* In `lib/Optimizer/Scalar/InstSimplify.cpp`, I collapse the `==` and `!=` switch cases  as the code was already near identical and introducing this optimization block to both cases was adding further duplication.
* In `lib/BCGen/HBC/Passes.cpp`, I refactor the unique literal memoization code into a `getLoadEntry` lambda function so it can be re-used outside of the for-loop in the clean-up assignment pass.

## Test Plan
We execute `../build/bin/hermes -dump-bytecode -O ../bk.js` against a 33MB React Native production bundle.

```
wc -l ../byte*
 2871165 ../bytecode.bk.none.out
 2867422 ../bytecode.bk.first.out (saved 3,743 statements)
 2867409 ../bytecode.bk.second.out (saved 3,756 statements)
 2867329 ../bytecode.bk.both.out (saved 3,836 statements)
```

We also add tests for the specific use-cases that are being optimized:

```
cmake --build ../build --target check-hermes all -j 4
```


tmikov:
Wow, I don't think we have ever received such thoughtful and high quality compiler PR. This will be a meaningful improvement to Hermes, plus it starts the work on normalization, which we have been talking about for a long time.

The PR looks good, but there are complications around the code it touches, which make it difficult to land it with its current structure. Specifically:
- IntSimplify and LoadConstants have diverged in the next major release of Hermes (Static Hermes), making changes difficult to forward-port. At the same time, Static Hermes is changing very rapidly and doesn't have a release date, so is not a suitable target.
- We are in a process of landing a stack addressing a major problem in object literal performance, which again changes LoadConstants in pretty significant ways. There will be merge conflicts.

To make it practical to land this, I think you should:
1. Split it in two separate PRs. I believe both have sufficient value to stand on their own. The second PR will have to wait until after we have landed our changes and I am afraid it will not be a trivial rebase.
2. Split the changes into separate passes, where practical. Specifically, I think it makes sense to create a new normalization pass instead of folding it into `InstSimplify`. We can then start adding more stuff to the normalization pass, for example always putting constants as a second operand when possible, etc.

So, basically I think the first PR should add a new `NormalizeInst` pass (suggestions for a better name are welcome).

At this time we are not too concerned about the overhead of adding new passes. The benefit of easier review, easier maintenance, and perhaps most importantly, easier forward porting to Static Hermes, outweigh the perf cost for now. 

In the future we are planning for a mechanism of composing multiple "peephole" pass into a single instruction scan, while keeping their logic separate. So, any performance we lose we should be able to recover.

sebholl:
Awww, shucks 😇 . I'm actually worked at Facebook for 6 years, and remember the challenges of reconciling external PRs with internal roadmap so now I'm out the outside, I'm trying to do my best to over-communicate. Anyway, I really appreciate the speedy feedback.

I'll take stab at splitting out the PRs. Definitely would love to learn more about Hermes Stable if you have anything already written out. There were a couple other optimizations I was hoping to submit (specifically to do with de-babeling) which would love your 2c on the best way to approach. Performance seems to suffer from over-polyfilling and I feel like there may be a lot of opportunity to deconstruct that.

For the `NormalizeInst` pass, my only question is at what point in the optimization pipeline should it run? We do it before invocations of `InstSimplify` and we might not have yet reduced some operands to literals and so will miss some identifications. Yet, we execute it too late and we'll miss the new transforms being worked through a lot of the passes. It almost feels like we'd want to pair-up alternate runs between `InstSimplify` and `NormalizeInst`?

At least combining in `InstSimplify`, we don't have that additional sequencing problem.

tmikov:
> For introducing the NormalizeInst pass, my only question is at what point in the optimization pipeline should it run?

I agree that it makes sense to start by inserting an NormalizeInst after every SimplifyInst in the pipeline. That would be acceptable for landing. Afterwards we could look at stat counters to see whether it needs further tweaking. 

> Definitely would love to learn more about Hermes Stable if you have anything public already written out.

Do you mean "Static Hermes" (the next major version)? There is nothing out yet - while it is not a secret, it hasn't been announced. We are planning to put up a detailed README one of these days, which I will be glad to share.

> There were a couple other optimizations I was hoping to submit (specifically to do with de-babeling) which would love your 2c on whether/how it's worth it to approach.

That is great! I think the best approach is to submit an issue with the proposed optimization (please ignore the issue template), so we can discuss the details, etc. It would avoid having to refactor stuff like in this PR.

Specifically about object spread, Hermes already supports it natively, but we haven't done much optimization.

tmikov:
FYI, we just enabled the "Discussions" tab, for more informal questions and ideas.

sebholl:
Followed up with a canonicalization PR here: https://github.com/facebook/hermes/pull/1075

tmikov:
Hi, sorry we haven't commented here in a while,I haven't forgotten about these PRs, though admittedly https://github.com/facebook/hermes/pull/1175 made me feel more guilty.

Since we have now publicly announced Static Hermes, I can finally be fully transparent. When we started working on Static Hermes, which will be a drop-in replacement for Hermes, among many other things, Hermes has been frozen except for bugfixes and small changes that can trivially be ported to the Static Hermes branch without rebase conflicts. 

Accepting a change this in Hermes would create **a lot** of work for us porting to Static Hermes, where the compiler has diverged, especially given that Hermes is frozen.

I realize that this is creating even more churn for you, bit this and https://github.com/facebook/hermes/pull/1075 should really be against the `static_h` branch. Then we will integrate it, and it will be maintained, etc.

PR #1091, opened 2023-08-11
fix unresponsive video container

ranjitbudhathoki:
There was a video container on the documentation site that was not responsive in smaller devices and was not looking good. 

Currently it looks something like this on smaller screen.
image

So, I have fixed that unresponsive video container and after the fix, it looks like this.

image

This is my first pull request on open source. So, if I have missed something please let me know.


tmikov:
Hi, AFAICT, the change looks reasonable (I am not a web developer), but please try to avoid noise by changing quotation marks and formatting. Can you please resubmit the PR with only the actual change.

ranjitbudhathoki:
Yeah sure. Sorry, for that.

ranjitbudhathoki:
I have made the changes. Please let me know if I am missing something else.

mattbfb:
Perhaps this styling should target just the video in question, rather than setting a style for all `iframe` elements on the page. I think this could be done by giving the video container a `className` and making the CSS apply only to that.

The `banner` style on the page can looks like a good example:

https://github.com/facebook/hermes/pull/1091/files#diff-ac42b5536aba05e65aa12e65c4f44017645c48c18c9dd4893b93da1d2a1acf28L80

https://github.com/facebook/hermes/pull/1091/files#diff-f4de6df0b1da89a2c0603ab69cddacae0264fd9cc03e44d498968586137594a7L46

ranjitbudhathoki:
@mattbfb Thanks. You  meant something like this?

PR #1107, opened 2023-08-24
Update README.md

DominicTobias:
Thanks for this plugin. It wasn't obvious to me from the instructions this needed to be added to plugins and spent quite a while debugging why the parser couldn't be found


pieterv:
Oh interesting looking at the prettier docs it looks like this is a change with prettier v3 where you now need to explicitly list the plugins rather than them being searched for you.

Could you also remove this line from the readme as well since its no longer accurate:

> Prettier will automatically load this plugin if you install it in the same `node_modules` directory where Prettier is located.

Thanks for the fix!

pieterv:
Btw if you are using prettier v3 i believe most features this plugin provides should be available there. Did you have an example of why this plugin was needed for you? This was mostly created as a way to release Flow features during the time prettier was blocked from releasing while waiting for the v3 release. Just keep in mind we still use prettier v2.8.8 internally so this combination is not as well tested.

DominicTobias:
Yeah I thought so from their changelog but as soon as I don't specify hermes Mapped Types don't work 🤔
Screenshot 2023-08-25 at 7 58 53 AM

pieterv:
Oh i see the issue you are using the babel parser in your prettier config, if you switch it to the "flow" parser it will work as expected in prettier v3 without needing this plugin. Up to you though, there are some advantages to using hermes parser (faster perf, gets features first etc).

Example: https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACACgJwgwB5cA+bAXm2AG0BrONAHSm3Y86+wEs3G0EAGZ4AukjwBfEABoQRGD2gBnZKACG+QgHdcmhKpTqANtvVpVcgEb51YRjADK6gLZwAMnzjIhJ5XGtbezgnDDs+AHNkGHwAVwCQfxceaLiE5UjjOABFWIh4Hz8EgCtlAA9HTJy8gqRfY385AEcauAIiQxB1ZQBaKDg4ABMh2VRbHmNIgGEIFxd1ZC7jY1GMqAisgEEYGJ4rWPhcOHxPfsKGhIALGBdjAHVLnnhlMLA4RwMnngA3J7RFsDKSwgb7xACSUGGsEcYHwPAwME2kMc6Cy50aIAwhH8d1sGEWWLg-nw328cj4xJgBHUEXm6ISYXwxMWQmMEG0oyxfBgdx4gxgl2QAA4AAxyfBwFo8CXU2kLOpFOQwdRWXn8wVIABMcli-gAKirDPUMXAXFYhsNBu51OtYjS4AAxCD4eY7SKLdQHCAgSSSIA

DominicTobias:
> Oh i see the issue you are using the babel parser in your prettier config, if you switch it to the "flow" parser it will work as expected in prettier v3 without needing this plugin. Up to you though, there are some advantages to using hermes parser (faster perf, gets features first etc).
> 
> Example: https://prettier.io/playground/#N4Igxg9gdgLgprEAuEMCeAHOACACgJwgwB5cA+bAXm2AG0BrONAHSm3Y86+wEs3G0EAGZ4AukjwBfEABoQRGD2gBnZKACG+QgHdcmhKpTqANtvVpVcgEb51YRjADK6gLZwAMnzjIhJ5XGtbezgnDDs+AHNkGHwAVwCQfxceaLiE5UjjOABFWIh4Hz8EgCtlAA9HTJy8gqRfY385AEcauAIiQxB1ZQBaKDg4ABMh2VRbHmNIgGEIFxd1ZC7jY1GMqAisgEEYGJ4rWPhcOHxPfsKGhIALGBdjAHVLnnhlMLA4RwMnngA3J7RFsDKSwgb7xACSUGGsEcYHwPAwME2kMc6Cy50aIAwhH8d1sGEWWLg-nw328cj4xJgBHUEXm6ISYXwxMWQmMEG0oyxfBgdx4gxgl2QAA4AAxyfBwFo8CXU2kLOpFOQwdRWXn8wVIABMcli-gAKirDPUMXAXFYhsNBu51OtYjS4AAxCD4eY7SKLdQHCAgSSSIA

Ah I see good to know! Yeah I'll keep hermes now that it's setup

Brianzchen:
Will also note that using hermes instead of built in prettier plugin will allow usage of flow comment types

PR #1111, opened 2023-08-31
[ts2flow] Convert built-in type primitives and literal types

seonggun-kim:
## Summary
[ts2flow] Convert built-in type primitives and literal types.

## Test Plan

Added unit and regression tests.


tmikov:
We will probably need the parser and ESTree.def changes to be split in a separate PR, because we keep the parser synchronized between Hermes and Static Hermes. But before you do that, let's wait for @avp to review.

avp:
This looks good to me, and the simple `NodeList` override in the visitor is a good idea.

But yeah we need to split off the ESTree.def/JSParserImpl changes because I will port those back to the `main` branch. Can you make a separate PR with just those changes against `static_h`? Then we can land them separately and not block on the sync to `main`.

seonggun-kim:
Thank you for your review, @avp. I separated out the parser changes into #1115.

avp:
5c498d75aa5462f1e9382041e0ee898a0b2cdeb5

PR #1134, opened 2023-09-25
Fix wasm optimization compilation errors

farwayer:
## Summary

For now enabling the HERMES_RUN_WASM flags causes compilation errors

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -50,7 +50,6 @@ add_hermes_library(hermesFrontend
   IR/Instrs.cpp
   Utils/Dumper.cpp
   Optimizer/Wasm/WasmIntrinsics.cpp
-  Optimizer/Wasm/EmitWasmIntrinsics.cpp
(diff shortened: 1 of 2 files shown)


lib/CMakeLists.txt
```diff
@@ -20,6 +20,7 @@ add_hermes_library(hermesOptimizer
   Optimizer/Scalar/InstSimplify.cpp
   Optimizer/Scalar/Auditor.cpp
   Optimizer/Wasm/WasmSimplify.cpp
+  Optimizer/Wasm/EmitWasmIntrinsics.cpp
```
tmikov, line 23: If we are moving EmitWasmIntrinsics into the optimizer lib, it should be removed from the frontend lib below.

tmikov:
Closing, since no response from PT author in two weeks.

farwayer:
Ah sorry, missed the message.
Removed from the frontend lib. Should `Optimizer/Wasm/WasmIntrinsics.cpp` be moved to optimizer lib too? I'm not familiar with the Hermes internal kitchen.

PR #1139, opened 2023-10-01
docs: fix typo in CrossCompilation.md

eltociear:
## Summary
happend -> happened


avp:
Thanks for the PR. These bullets should likely be in the present tense so this fix doesn't seem completely correct.

PR #1162, opened 2023-10-18
refactor(promise):  don't include the promise rejection tracker hook in the bytecode

ecreeth:


ecreeth:
I'm getting an `Uncaught TypeError: Promise rejection tracking hook was not registered` from [`hermesInternalEnablePromiseRejectionTracker`](https://github.com/facebook/hermes/blob/main/lib/VM/JSLib/HermesInternal.cpp#L720). I need to investigate a little, so I'm going to convert the PR to draft.

avp:
This part of the code has a couple moving parts, so maybe I can help explain what's going on here:

The code in `promise/index.js` in Hermes are calling `setPromiseRejectionTrackingHook`, which performs a different operation from `enablePromiseRejectionTracker`.

* `setPromiseRejectionTrackingHook` is entirely internal to Hermes - it's only got the one callsite in InternalBytecode. It doesn't actually enable promise rejection tracking; it only sets up a single pointer inside the Runtime to point to the `enable` closure (https://github.com/facebook/hermes/blob/main/lib/VM/JSLib/HermesInternal.cpp#L712)

* `enablePromiseRejectionTracker` is called (in development mode) by RN or any other integrator with any options they choose. 

The reason you're getting a TypeError is that Hermes expects the hook to be set, because it has no way of knowing whether a caller wants to enable it.

The doc-comments on those two functions here might also help: https://github.com/facebook/hermes/blob/main/lib/VM/JSLib/HermesInternal.cpp#L697

Issue #1169, opened 2023-10-19
The Math.random implementation in Hermes is collision prone for the App use case


psaha-rbi:
## Bug Description

The implementation is 
```
CallResult<HermesValue> mathRandom(void *, Runtime &runtime, NativeArgs) {
  RuntimeCommonStorage *storage = runtime.getCommonStorage();
  if (!storage->randomEngineSeeded_) {
    std::minstd_rand::result_type seed;
    seed = std::random_device()();
    storage->randomEngine_.seed(seed);
    storage->randomEngineSeeded_ = true;
  }
  std::uniform_real_distribution<> dist(0.0, 1.0);
  return HermesValue::encodeUntrustedNumberValue(dist(storage->randomEngine_));
}
```
As you can see, it uses `std::random_device()` as seed. It is an unsigned int and therefore there are only about 4B possible values.  There is a 71% chance of a single collision if you generate 100k values using the [Birthday problem formulation](https://en.wikipedia.org/wiki/Birthday_problem#Approximations). So many different instances of this will go through the exact same sequence. While this is not used for security, this is also unusable for things like UUID v4 generation since there will be tons of collisions.

Please consider using a much larger seed space - ideally, 100+ bits. A simple way would be to concatenate multiple outputs from std::random_device().

- [ ] I have run `gradle clean` and confirmed this bug does not occur with JSC

Hermes version: latest
React Native version (if any): all
OS version (if any): all
Platform (most likely one of arm64-v8a, armeabi-v7a, x86, x86_64): all

## Steps To Reproduce

Theoretical but we have seen collisions matching the birthday problem formula out in the wild.

code example:

## The Expected Behavior
Much harder to produce collisions.

tmikov:
Hi, I removed the "bug" label, since this not a bug - EcmaScript specifies it as implementation defined and puts no requirements for its "randomness".

Can you provide links to the relevant implementations in v8 or JSC? If we are to change this, we should probably match theor behavior. 

The best way to improve this would probably be to submit a PR.

tmikov:
Also, can you please clarify your 4B values comment. That is just the seed, which is used once. How does it affect the quality of the values that are generated after?

psaha-rbi:
[...]

PR #1171, opened 2023-10-22
Optimize the Math.random implementation

woshiccm:
Summary: Fix [The Math.random implementation in Hermes is collision prone for the App use case](https://github.com/facebook/hermes/issues/1169)

### DIFF
--- a/lib/VM/JSLib/Math.cpp
+++ b/lib/VM/JSLib/Math.cpp
@@ -242,10 +242,10 @@ CallResult<HermesValue> mathRandom(void *, Runtime &runtime, NativeArgs) {
     if (storage->randomState_ == 0) {
       storage->randomState_ = 1;
     }
-      storage->randomEngineInited = true;
-    }
-    uint64_t val = XorShift64(&storage->randomState_);
-    return HermesValue::encodeUntrustedNumberValue(ToDouble(val));
+    storage->randomEngineInited = true;
+  }
+  uint64_t val = XorShift64(&storage->randomState_);
+  return HermesValue::encodeUntrustedNumberValue(ToDouble(val));
 }
 
 CallResult<HermesValue> mathFround(void *, Runtime &runtime, NativeArgs args)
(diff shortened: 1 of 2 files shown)


woshiccm:
@tmikov @psaha-rbi pls help to review, thanks

lib/VM/JSLib/Math.cpp
```diff
@@ -205,18 +207,45 @@ CallResult<HermesValue> mathPow(void *, Runtime &runtime, NativeArgs args) {
...
   RuntimeCommonStorage *storage = runtime.getCommonStorage();
-  if (!storage->randomEngineSeeded_) {
-    std::minstd_rand::result_type seed;
-    seed = std::random_device()();
-    storage->randomEngine_.seed(seed);
-    storage->randomEngineSeeded_ = true;
+  if (!storage->randomEngineInited) {
+    struct timeval tv;
+    gettimeofday(&tv, NULL);
+    storage->randomState_ = static_cast<uint64_t>((tv.tv_sec * SECONDS_TO_SUBTLE) + tv.tv_usec);
```
psaha-rbi, line 240: I think this is problematic because 
- clock resolution maybe more coarse grained than microseconds. e.g. on Linux it is usually 10us. 
- Even if there is perfect microsecond granularity, you'll start getting collisions if thousands of uuids are generated every second across many devices since many of them will get the same read of microsecond. In general, even though the "microseconds since epoch" is a large value it has very strong temporal locality in the sense that at any point in time all values will be tightly clustered.
- I also feel not so warm about rolling a custom RNG where there are so many available in the standard library specially, without running any of the standard randomness test suites on it.

I'd recommend using `std::random_device()()` to generate multiple 32-bit numbers and concatenating those to form a 64 bit seed or better yet, use a `std::seed_seq()` with enough members to initialize a Mersenne twister to get a very long period.

woshiccm: OK, thanks, will try it

woshiccm: I have changed the implement

neildhar:
Closing since we merged https://github.com/facebook/hermes/commit/34f0b2bcd3020e45d5f0b38b230a38324c8ef042 instead.

PR #1175, opened 2023-10-24
Switch Math.random() to faster 64-bit seeded implementation.

sebholl:
The `std::minstd_rand` implementation has its `seed()` value type defined as `result_type` which is aliased to `unsigned int`, a 32-bit value on supported platforms. Using 32-bit seeds and results can lead to real-world [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem) collisions. As per Purujit's recommendation, we update the implementation of `Math.random()` to utilize a 64-bit LCG that can accept a 64-bit seed and also generate a 64-bit result.

There is a lot of [analysis](https://prng.di.unimi.it/) and [blog posts](https://v8.dev/blog/math-random) around the best way to implement random number generators. This PR attempts to deliver two improvements on the current implementation:

* Having 64-bit seeded values and result types instead of 32-bit.
* Ideally, being faster or at least the same speed.
* (optional) having better randomness properties.

## Summary

Addressing #1169, we benchmark the `xoroshiro128+` (used by browsers); `randomDevice` (cryptographically secure); `mt19937_64` (complex but fast) and `lcg64` (simple & fast) unsigned 64-bit integer PRNGs against three different uniform random distribution implementations:

1. **`std::uniform_real_distribution()`**: standard library implementation, but slow as it requires multiple random numbers to be generated for a single [0, 1) value; plus it has known bugs where it can in fact return 1 because of rounding.
2. **Bit Twiddle Approach 1**: use the `(0x3FFL << 52 | uint64_t(x) >>> 12) - 1.0` from Java and Firefox that pairs a fixed exponent with 52 random mantissa bits to generate a double between [1, 2) and then subtracting 1. 
3. **Bit Twiddle Approach 2**: use the `(uint64_t(x) >> 11) * 0x1.0p-53` approach which generates a double between [0, 1) directly with 53 bits worth of possible output (twice as many values as the 52 bits from Approach 1).

The bit twiddling approaches are described [here](https://prng.di.unimi.it/#:~:text=Generating%20uniform%20doubles%20in%20the%20unit%20interval).

Given the large number of platforms and architectures that Hermes is compiled to, we make copious use of `static_assert`s to make sure that we don't encounter unexpected rug-pulls or changing bit-widths when upgrading compilers; or adding new targets. These should of course be elided after compilation and so will not affect runtime performance.

### Alternatives Considered

- Instead of manually bit packing a 64-bit seed, an implementation leveraging C++11's `seed_seq` was explored that would ideally provide even more entropy, however the author of [this article](https://www.pcg-random.org/posts/cpp-seeding-surprises.html) asserts that 64 bits of seed seq data does not necessarily produce 64 bits of output. As such, we keep it simple with a single seed value so it is easier to reason with in the future.

## Test Plan

We compile the following `benchmark.js` to byte-code using `../bin/hermes --emit-binary -fstatic-builtins -O ./benchmark.js -out ./benchmark.out.bin`:

```
const n_iterations = 10_000_000;
let sum = 0, min = 2, max = -2;
const rndFunction = Math.random;
for (const i = 0; i < n_iterations; i++) {
    const rnd = rndFunction();
    if (rnd < min)
        min = rnd;
    if (rnd > max)
        max = rnd;
    sum += rnd;
}
print(JSON.stringify({min, avg: sum / n_iterations, max}));
```

And then we execute this bytecode on each of the implementation's resulting `hvm` binary (I'm using `hyperfine`) on my local development machine (Apple MacBook Air M2 with ARM64):

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `./hvm-main ./benchmark.out.bin` | 8.371 ± 0.065 | 8.310 | 8.508 | 1.05 ± 0.01 |
| `./hvm-xoroshiro128plus-twiddlebits ./benchmark.out.bin` | 8.075 ± 0.023 | 8.040 | 8.127 | 1.01 ± 0.00 |
| **`./hvm-xoroshiro128plus-twiddlebits2 ./benchmark.out.bin`** | 8.019 ± 0.033 | 7.975 | 8.068 | 1.00 ± 0.01 |
| `./hvm-lcg64-urd ./benchmark.out.bin` | 8.288 ± 0.041 | 8.232 | 8.369 | 1.04 ± 0.01 |
| `./hvm-lcg64-twiddlebits ./benchmark.out.bin` | 7.988 ± 0.030 | 7.949 | 8.039 | 1.00 |
| `./hvm-lcg64-twiddlebits2 ./benchmark.out.bin` | 8.010 ± 0.029 | 7.965 | 8.058 | 1.00 ± 0.01 |
| `./hvm-mt19937_64-urd ./benchmark.out.bin` | 8.109 ± 0.047 | 8.058 | 8.190 | 1.02 ± 0.01 |
| `./hvm-mt19937_64-twiddlebits ./benchmark.out.bin` | 8.135 ± 0.055 | 8.063 | 8.262 | 1.02 ± 0.01 |
| `./hvm-mt19937_64-twiddlebits2 ./benchmark.out.bin` | 8.143 ± 0.039 | 8.067 | 8.187 | 1.02 ± 0.01 |
| `./hvm-mt19937_64-urd-hoisted ./benchmark.out.bin` | 8.028 ± 0.029 | 7.977 | 8.082 | 1.01 ± 0.01 |
| `./hvm-randomDevice-urd ./benchmark.out.bin` | 9.859 ± 0.036 | 9.800 | 9.918 | 1.23 ± 0.01 |
| `./hvm-randomDevice-twiddlebits ./benchmark.out.bin` | 9.514 ± 0.028 | 9.463 | 9.550 | 1.19 ± 0.01 |
| `./hvm-randomDevice-twiddlebits2 ./benchmark.out.bin` | 9.558 ± 0.033 | 9.498 | 9.610 | 1.20 ± 0.01 |

Although `xoroshiro128plus-twiddlebits2` appears optimal, @neildhar preferred `mt19937_64-urd` as it's less invasive a change and has less complexity to maintain.

### DIFF
--- a/include/hermes/VM/JSLib/RuntimeCommonStorage.h
+++ b/include/hermes/VM/JSLib/RuntimeCommonStorage.h
@@ -38,7 +38,7 @@ struct RuntimeCommonStorage {
   MockedEnvironment tracedEnv;
 
   /// PRNG used by Math.random()
-  std::minstd_rand randomEngine_;
+  std::mt19937_64 randomEngine_;
   bool randomEngineSeeded_ = false;
 };
 
--- a/lib/VM/JSLib/Math.cpp
+++ b/lib/VM/JSLib/Math.cpp
@@ -210,8 +210,14 @@ CallResult<HermesValue> mathPow(void *, Runtime &runtime, NativeArgs args) {
 CallResult<HermesValue> mathRandom(void *, Runtime &runtime, NativeArgs) {
   RuntimeCommonStorage *storage = runtime.getCommonStorage();
   if (!storage->randomEngineSeeded_) {
-    std::minstd_rand::result_type seed;
-    seed = std::random_device()();
+    std::random_device randDevice;
+
+    auto randValue = randDevice();
+    static_assert(sizeof(randValue) == 4, "expecting 32 bits from std::random_device()");
+    
+    uint64_t seed = (uint64_t(randValue) << (8*sizeof(randValue))) | randDevice();
+    static_assert(sizeof(decltype(storage->randomEngine_)::result_type) >= 8, "expecting at least 64-bit result_type for PRNG");
+
     storage->randomEngine_.seed(seed);
     storage->randomEngineSeeded_ = true;
   }


include/hermes/VM/JSLib/RuntimeCommonStorage.h
```diff
@@ -38,7 +38,7 @@ struct RuntimeCommonStorage {
   MockedEnvironment tracedEnv;
 
   /// PRNG used by Math.random()
-  std::minstd_rand randomEngine_;
+  std::mt19937_64 randomEngine_;
```
purujit, line 41: Have you considered using the 64 bit LCG like the one in the article you shared in your PR description?
```
typedef std::linear_congruential_engine<
    uint64_t, 6364136223846793005U, 1442695040888963407U, 0U>
        knuth_lcg;  /* Knuth's preferred 64-bit LCG */
```
I have slight preference for this one since it is easier to reason about than the Mersenne twister and also, minstd_rand is just the 32 bit version of it and it will therefore be less of a change to just use a 64 bit LCG.

lib/VM/JSLib/Math.cpp
```diff
@@ -205,18 +205,26 @@ CallResult<HermesValue> mathPow(void *, Runtime &runtime, NativeArgs args) {
...
   RuntimeCommonStorage *storage = runtime.getCommonStorage();
   if (!storage->randomEngineSeeded_) {
-    std::minstd_rand::result_type seed;
-    seed = std::random_device()();
+    auto rnd_dev = std::random_device();
+
+    auto rnd_value = rnd_dev();
+    static_assert(sizeof(rnd_value) == 4, "expecting 32 bits from std::random_device()");
+    
+    auto seed = (uint64_t(rnd_value) << (8*sizeof(rnd_value))) | rnd_dev();
```
purujit, line 221: Perfect! This is what I had in mind.

sebholl: <3

include/hermes/VM/JSLib/RuntimeCommonStorage.h
sebholl: I ran benchmarks with `knuth_lcg` (updated in PR description above) and it appears to appear slower for some reason.

purujit (commented):
Lgtm.

lib/VM/JSLib/Math.cpp
```diff
@@ -205,18 +205,26 @@ CallResult<HermesValue> mathPow(void *, Runtime &runtime, NativeArgs args) {
   return HermesValue::encodeUntrustedNumberValue(expOp(x, y));
 }
 
+// hoisting to top-level benchmarks faster than static local
+std::uniform_real_distribution<> _std_real_dist_01(0.0, 1.0);
```
neildhar, line 209: In general, we avoid having any global constructors. Keeping this as a static local or a global may also not be thread safe, as multiple runtimes can exist in the same process running on different threads. (`std::uniform_real_distribution<RealType>::operator()` is not const)

We can just keep this as a local to avoid these issues and minimise the scope of this change.

lib/VM/JSLib/Math.cpp
```diff
@@ -205,18 +205,26 @@ CallResult<HermesValue> mathPow(void *, Runtime &runtime, NativeArgs args) {
...
+
 // ES5.1 15.8.2.14
 // Returns a Hermes-encoded pseudo-random number uniformly chosen from [0, 1)
 CallResult<HermesValue> mathRandom(void *, Runtime &runtime, NativeArgs) {
   RuntimeCommonStorage *storage = runtime.getCommonStorage();
   if (!storage->randomEngineSeeded_) {
-    std::minstd_rand::result_type seed;
-    seed = std::random_device()();
+    auto rnd_dev = std::random_device();
```
neildhar, line 216: nit

```suggestion
    std::random_device rndDev;
```

lib/VM/JSLib/Math.cpp
neildhar, line 221: nit
```suggestion
    uint64_t seed = (uint64_t(rnd_value) << (8*sizeof(rnd_value))) | rnd_dev();
```

neildhar (commented):
Thank you for working on this! Just a few nits.

Please also use camel case instead of snake case, as that is the style in Hermes.

lib/VM/JSLib/Math.cpp
sebholl: Sounds good.

lib/VM/JSLib/Math.cpp
sebholl: Done.

lib/VM/JSLib/Math.cpp
sebholl: Done.

sebholl (commented):
Without hoisting, benchmarks now make the Purujit's recommended LCG 64 algorithm faster so swapped back over to that. I instead decided to swap up in the bit twiddle the browsers use to map from `uint64` to `double` which are way faster anyway and need minimal state/memory.

sebholl:
I'm going to bechmark an implementation with `xoshiro128+` too which is what the other browsers use.

sebholl:
Okies, @neildhar - I think I'm wrapped up on this now. Ended up where @tmikov postulated with just implementing the `xoroshiro128+` algorithm that the other browsers user, and then included the bit twiddle speed up too. So faster and hopefully more "random" all round.

neildhar:
@sebholl Thanks for digging more into this. Given that the performance differences here are relatively small and likely vary across platforms, the previous approach seems much simpler and less invasive (unless there are legitimate issues with the quality of the random values produced).

If there are other performance or quality issues with the existing implementation, I think we should consider them separately from this change to fix #1169.

shollington-rbi:
@neildhar - do you mean revert the bit twiddle to the `std::unique_real_distribution<>` or change the actual PRNG implementation back to a lcg or mt?

The challenge is that there is a slight coupling between the two. This PRNG the browsers use only generates 59 good upper bits but it's fine because the bit twiddle discards the bottom 11. The standard library `std::unique_real_distribution()` doesn't do this and so we'd have to move to a different algo from the browsers to continue using this distribution mechanism.

Also, I believe there will be additional complexity wrapping this function if we want it to conform to the generator semantics so it can be used by `std::unique_real_distribution()`. Providing min/max constants etc.

Finally there is a known issue with `std::generate_canonical()` (used by `std::unique_real_distribution()`) where it can return 1 which, whilst unlikely, I believe would be against the ECMA spec for `Math.random()` behavior:

https://stackoverflow.com/questions/25668600/is-1-0-a-valid-output-from-stdgenerate-canonical

I understand the desire to minimize invasiveness and have to of course  defer to you, but my recommendation in this case is to make these changes as one.

neildhar:
> do you mean revert the bit twiddle to the std::unique_real_distribution<> or change the actual PRNG implementation back to a lcg or mt?

I meant both. I have no expertise in random number generation, but the initial implementation using the MT and `uniform_real_distribution` seems like the simplest path to solving the issue. (as long as it is correct)

> there is a known issue with std::generate_canonical()

My understanding from reading through your link is that this problem applies to the `float` specialisation, but not to the `double` specialisation we are using, so this shouldn't be a concern. Is that correct?

sebholl:
> I meant both.

Ah got it, makes sense @neildhar. I've reverted PR back to that state.

> My understanding from reading through your link is that this problem applies to the float specialisation, but not to the double specialisation we are using, so this shouldn't be a concern. Is that correct?

I believe you are correct, all of the references I see appear to be for `float`s. There's a lot of other literature [[1](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0952r0.html); [2](http://www.blahonga.org/2015/04/state-of-art-of-random-double-generation.html)] but this is getting far outside my expertise so happy to let sleeping dogs lie.

PR #1176, opened 2023-10-25
Added support for ES6 classes

rFlex:
## Summary

We are currently experimenting with a Hermes integration as a potential replacement for the QuickJS engine. As part of the migration, we've realized that the lack of support for ES6 classes and block scoped variables within Hermes is a bigger deal than expected for us. We have 3 engine integrations which all support ES2017+, and without support for ES6 classes and block scoped variables we have to either downgrade the ES target to ES5, or integrate Babel in our compilation pipeline which are both invasive and risky changes for us. In order to fix this, I've taken a stab at supporting ES6 classes somewhat natively within Hermes.

I've went with the simplest and the less risky approach that could get the job done. This PR doesn't change the IR, it instead transforms the AST at compile time to convert ES6 classes into plain ES5 classes, very much like what the TypeScript and Babel compiler do. A more advanced implementation would likely add support for classes within the IR. This PR ended up to be fairly straightforward to do, as the JS parser implemented within Hermes already supports most of the ES6 syntax, it's only the IR Gen pass which doesn't support it.

## Supported ES6 class features
- Define a class with a parent class (`class Child extends Parent {}`)
- Method and static method definitions (`method() {}` and `static method() {}`)
- Property and static property definitions (`get prop() {}`, `static get prop() {}`, `set prop(v) {}`, `static set prop(p) {}`
- Super call expression in method and static method definitions (`super.method()`)
- Super member expression in property and static property definitions (`super.property`) 
- User defined constructor (`constructor() {}`)
- Properties that are immediately initialized. (`property = someValue`)
- Class expressions (`const MyClass = class {}`)
- Class expressions with private named identifier (`const MyClass = class Self { constructor() { Self.hello = 'world' }})`)

## Unsupported ES6 class features
- static block initializers (`static { // code to eval at class load time }`). The existing parser is not able to parse these blocks yet.
- Private methods and properties (`#privateMethod() {}`)
- Inheritance from built-in types like `Error` or `Map` (`class CustomError extends Error {}`)

## Changes

- Added a `enableES6Classes` compile flag. When enabled, a new post processing step will occur on the AST which will transform ES6 classes into ES5 functions, similar to what Babel or the TypeScript compiler does.
- Added `lib/AST/ES6Class`, an AST visitor which visits the ClassNodeDeclaration and ClassExpressionNode and will turn them into plain ES5 functions. The generated AST relies on a new helper lib that is meant to be available on the global object as `HermesES6Internal`.
- Added the `HermesES6Internal` helper module, which implements `defineClass`, `defineClassProperty`, ` defineStaticClassProperty`, `defineClassMethod` and `defineStaticClassMethod`.
- Added a few tests

## Test Plan
- All Hermes tests are passing


include/hermes/CompilerDriver/CompilerDriver.h
```diff
@@ -101,5 +101,8 @@ extern llvh::cl::opt<bool> EnableEval;
 extern llvh::cl::list<std::string> InputFilenames;
 extern llvh::cl::opt<bool> OptimizedEval;
 extern llvh::cl::opt<bool> VerifyIR;
+extern llvh::cl::opt<bool> VerifyIR;
```
tmikov, line 104: Nit: duplicate line

lib/CompilerDriver/CompilerDriver.cpp
```diff
@@ -297,6 +297,12 @@ static llvh::cl::alias _EnableBlockScoping(
     Hidden,
     llvh::cl::aliasopt(EnableBlockScoping));
 
+opt<bool> ES6Class(
+    "Xes6-class",
+    init(false),
```
tmikov, line 302: Let's also add a "Hidden" for now.

lib/InternalBytecode/03-ES6Class.js
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  function defineStaticClassMethod(cls, methodName, fn) {
+    Object.defineProperty(cls, methodName, {
+      enumerable: false,
+      value: fn,
+    });
+  }
+
+  globalThis.HermesES6Internal = {
```
tmikov, line 50: Consider freezing this object?

lib/AST/ES6Class.cpp
```diff
@@ -0,0 +1,675 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+#include "hermes/AST/ES6Class.h"
+#include "hermes/AST/RecursiveVisitor.h"
+#include "hermes/Parser/JSLexer.h"
+
+namespace hermes {
+struct ClassMemberKey {
```
tmikov, line 13: If this class (and similar types) is intended to be used only here, perhaps it should be in an anonymous namespace?

tmikov:
Looks good. I think we can land this when some minor comments are addressed. 

Can you please run it through clang-format (`./utils/format.sh`). Also, it would be good for future maintenance if there was a file level documentation block in `ES6Class.cpp` explaining what is happening - the transformation, etc.

As a rule we require all declarations (types, fields, methods/functions) to have doc comments. We do not enforce it strictly but at least the major ones should have it.

lib/InternalBytecode/03-ES6Class.js
```diff
@@ -0,0 +1,61 @@
+/* @nolint */ var initES6Class = function () {
+  function defineClass(ctor, superClass) {
+    ctor.prototype = Object.create(superClass && superClass.prototype, {
```
avp, line 3: We generally recommend caching global object lookups as local variables on first run, so you can rely on them to not be overridden by the user. As a bonus, it'll be faster to call functions that are already local variables instead of having to do a lookup on the global object.

```
// inside initES6Class
var Object_create = Object.create;

// inside defineClass
ctor.prototype = Object_create(/* args */);
```

Similarly for `Object.defineProperty`, etc.

rFlex:
@tmikov thanks for your review. Addressed the comments, ran clang format, added documentation around the main public API (`transformES6Classes`) as well as on a few key methods and helpers in `ES6Class.cpp`

avp (changes requested):
Thanks for the updates. Looks like this is pretty close to being able to merge, just some minor comments.

lib/CompilerDriver/CompilerDriver.cpp
tmikov: Can you please make this a hidden option?

PR #1180, opened 2023-10-31
[static_h] Switch Math.random() to faster 64-bit seeded implementation. (#1175)

sebholl:
Cherry-picking https://github.com/facebook/hermes/pull/1175 for `static_h`.


avp:
This was picked up by the periodic sync we do from main to static_h, so this PR can be closed.

PR #1186, opened 2023-11-10
Optimize UnicodeICU localeCompare

sujankh:
## Summary
This change eliminates repeated calls to `ucol_open` and `ucol_close` in `hermes::platform_unicode::localeCompare` for HERMES_PLATFORM_UNICODE_ICU. On a sample sort.js(that sorted 30k strings using localeCompare), the runtime goes down from ~500 ms to ~120ms.

**Before:**
I observed in my trace that ucol_open and ucol_close was called in every localeCompare invocation and each ucol_open took around 500-900 ns. ucol_close is tiny ~100ish ns in each call.
Note that the first call to ucol_open, it takes around 100us even after applying this diff.
image

Here is localeCompare call stack before the patch:
image

**After**
After making UCollator construction static, we avoid repeated ucol_open and ucol_close(except during static initialization)  thus making the sort.js sorting fast. 
Here is the localeCompare stack after the patch:
image

Looking at the code, i found out that HERMES_PLATFORM_UNICODE_ICU is not used in Android/iOS but can make desktop usage faster (it it useful??).

Here is the sort.js script which i got from one of the issues/posts in hermes.
```
print("gen...");
  const randInt = (end, start = 0) =>
    Math.floor(Math.random() * (end - start) + start);
  const alphabet = 'abcdefghijklmnopqrstuvwxyz';
  const names = Array.from({length: 30000}, () =>
     Array.from({length: 8}, () => alphabet[randInt(alphabet.length)]).join(''),
   );

  for(var i = 0; i < 5; ++i){
    print(names[i])
  }
  print('sorting...');
  const s = Date.now();
  names.sort(localeCompare);
  print(`...done, took ${Date.now() - s}ms`);

  function localeCompare(a, b) {
    const s = Date.now();
    const result = a.localeCompare(b);
    const e = Date.now();
    if (e - s > 1) {
      print(`slow localeCompare: ${e - s}ms for '${a}' vs '${b}'`);
    }
    return result;
  }

  for(var i = 0; i < 5; ++i){
    print(names[i])
  }
```

## Test Plan
**Execute sort.js before applying this patch**
```
hermes (main)$ ../build_release/bin/hermesc -emit-binary -out sort.hbc samples/sort.js
hermes (main)$ ../build_release/bin/hvm ./sort.hbc 
gen...
sorting...
...done, took 466ms
hermes (main)$ ../build_release/bin/hvm ./sort.hbc 
gen...
sorting...
...done, took 486ms
```

**Execute sort.js after applying this patch**
```
$ ../build_release/bin/hermesc -emit-binary -out sort.hbc samples/sort.js
hermes (icu_localecompare_optimize)$ ../build_release/bin/hvm ./sort.hbc 
gen...
sorting...
...done, took 119ms
hermes (icu_localecompare_optimize)$ ../build_release/bin/hvm ./sort.hbc 
gen...
sorting...
...done, took 127ms
```

**Run check-hermes:**
```
hermes (icu_localecompare_optimize)$ cmake --build ../build_release/ --target check-hermes
[0/1] Running the Hermes regression tests
Testing Time: 15.91s
  Expected Passes    : 1773
  Unsupported Tests  : 66
hermes (icu_localecompare_optimize)$ echo $?
0
```

### DIFF
--- a/lib/Platform/Unicode/PlatformUnicodeICU.cpp
+++ b/lib/Platform/Unicode/PlatformUnicodeICU.cpp
@@ -13,34 +13,59 @@
 
 #include <time.h>
 
+#include <memory>
+
 namespace hermes {
 namespace platform_unicode {
+
+namespace {
+struct UCollatorDeleter {
+  void operator()(UCollator *coll) {
+    ucol_close(coll);
+  }
+};
+
+/// Returns a pointer to the singleton UCollator. It is thread-safe because
+/// threads can only access const APIs using this handle(which are thread-safe
+/// as per the documentation). Please refer to
+/// https://unicode-org.github.io/icu/userguide/icu/design.html#thread-safe-const-apis,
+/// for more details. This
+/// function invokes ucol_open only once and keeps it as a singleton for better
+/// performance instead of ucol_open/ucol_close in each invocation.
+const UCollator *getUCollatorInstance() {
+  using UCollatorPtr = std::unique_ptr<UCollator, UCollatorDeleter>;
+  static UCollatorPtr coll = [] {
+    UErrorCode err{U_ZERO_ERROR};
+    UCollator *coll = ucol_open(uloc_getDefault(), &err);
+
+    if (U_FAILURE(err)) {
+      // Failover to root locale if we're unable to open in default locale.
+      err = U_ZERO_ERROR;
+      coll = ucol_open("", &err);
+    }
+    assert(U_SUCCESS(err) && "failed to open collator");
+
+    // Normalization mode allows for strings that can be represented
+    // in two different ways to compare as equal.
+    ucol_setAttribute(coll, UCOL_NORMALIZATION_MODE, UCOL_ON, &err);
+    assert(U_SUCCESS(err) && "failed to set collator attribute");
+    return UCollatorPtr(coll, UCollatorDeleter());
+  }();
+
+  return coll.get();
+}
+} // namespace
 
 int localeCompare(
     llvh::ArrayRef<char16_t> left,
     llvh::ArrayRef<char16_t> right) {
-  UErrorCode err{U_ZERO_ERROR};
-  UCollator *coll = ucol_open(uloc_getDefault(), &err);
-  if (U_FAILURE(err)) {
-    // Failover to root locale if we're unable to open in default locale.
-    err = U_ZERO_ERROR;
-    coll = ucol_open("", &err);
-  }
-  assert(U_SUCCESS(err) && "failed to open collator");
-
-  // Normalization mode allows for strings that can be represented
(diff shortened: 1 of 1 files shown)


lib/Platform/Unicode/PlatformUnicodeICU.cpp
```diff
@@ -13,12 +13,21 @@
...
 namespace hermes {
 namespace platform_unicode {
 
-int localeCompare(
-    llvh::ArrayRef<char16_t> left,
-    llvh::ArrayRef<char16_t> right) {
+namespace {
+struct UCollatorDeleter {
+  void operator()(UCollator *coll) {
+    if (coll) {
```
neildhar, line 24: I don't believe the deleter is invoked if the pointer is null.

lib/Platform/Unicode/PlatformUnicodeICU.cpp
```diff
@@ -32,16 +41,22 @@ int localeCompare(
...
+  return std::unique_ptr<UCollator, UCollatorDeleter>(coll, UCollatorDeleter());
+}
+} // namespace
+
+int localeCompare(
+    llvh::ArrayRef<char16_t> left,
+    llvh::ArrayRef<char16_t> right) {
+  static auto coll = openUCollator();
 
   auto result = ucol_strcoll(
```
neildhar, line 53: Could we add a comment about the safety of calling `ucol_strcoll` on the same `UCollator` from multiple threads?

I believe it is safe given https://unicode-org.github.io/icu/userguide/icu/design.html#thread-safe-const-apis, but we should document it. (and perhaps use a `std::unique_ptr<const UCollator>` to make it obvious)

neildhar (changes requested):
Thank you for working on this, I just have some minor comments.

lib/Platform/Unicode/PlatformUnicodeICU.cpp
sujankh: Fixed. Thanks, re-read cppreference to make sure.

lib/Platform/Unicode/PlatformUnicodeICU.cpp
sujankh: Added a comment. 
I tried `std::unique_ptr<const UCollator>` but turns out `ucol_close` does not accept a `const UCollator*`. I could use const_cast and reading `ucol_close`'s implementation I could see that there will be no UB due to const cast but it is not guaranteed on the API.

So I came up with current implementation where `getUCollator()` returns a `const UCollator*`. This will ensure that all the APIs invoked on the returned pointer will be thread-safe.

getUCollator() also owns the lifetime of the `unique_ptr<UCollator>`.

PR #1193, opened 2023-11-22
Enable prefabPublishing for aar on android platform

cmzy:
## Enable prefabPublishing for aar on android platform

This will enable we can use aar by prefab 

## Test Plan

NAN


dannysu:
hi @cmzy, I'm less familiar with the Android AAR import. Do you have an example project that uses the new AAR from your diff that I can look at?

Thanks

dannysu:
@cmzy
I gave this PR a quick try. It seems like the generated `hermesConfig.cmake` tries to reference `liblibhermes.so`, which doesn't exist, rather than the `libhermes.so` packaged in the AAR. So I can't actually get a project to build yet.

Do you have an example project that successfully compiles using header imports from the AAR with the new prefab settings? If so, would you be able to share that?
Thanks

cmzy:
> @cmzy I gave this PR a quick try. It seems like the generated `hermesConfig.cmake` tries to reference `liblibhermes.so`, which doesn't exist, rather than the `libhermes.so` packaged in the AAR. So I can't actually get a project to build yet.
> 
> Do you have an example project that successfully compiles using header imports from the AAR with the new prefab settings? If so, would you be able to share that? Thanks

Could you share me how did you try that ?  It is ok when I try to build it.

PR #1210, opened 2023-12-06
Update footer logo and links

neerajkumarc:
Updated the old oss logo with new one, also updated some of the old links to new ones.


tmikov:
We are deprecating the website.

PR #1219, opened 2023-12-12
Draft: Implement formatRange and formatRangeToParts on Apple and Android platforms

ewlsh:
## Summary

We want to support `formatRange` and `formatRangeToParts` for `Intl.DateTimeFormat` on Android and iOS. Our company uses the Android implementation internally and will use the iOS implementation once we move iOS to Hermes.

I've added a basic fallback for older Android versions, we don't use this internally but I'll test it.

For Apple this introduces support for `formatToParts` using the skeleton string to "build" each part - formatting the parts individually. This is expensive but can be avoided in the future if we move the implementation to Swift and utilize Swift's [`AttributedString`](https://developer.apple.com/documentation/foundation/attributedstring). Apple [has been working on](https://github.com/apple/swift-cmake-examples) Swift/C++ cmake interop.

For Android and Apple this introduces support for `formatRange` uses the respective interval formatters on both platforms.

## Test Plan

Adding support for existing test262 tests and porting some tests to the internal testsuite where the data won't align to do limitations in the Apple implementation


ewlsh:
I'm working on breaking apart our internal work into distinct MRs, this is currently rebased on https://github.com/facebook/hermes/pull/1218 but I may drop those commits soon.

lib/Platform/Intl/Intl.iml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ewlsh, line 1: note: drop

PR #1220, opened 2023-12-12
Improve test262 test suites on Apple and Android

ewlsh:
## Summary

For Android, this parameterizes the testsuite so each individual testcase/file is displayed separately and can be opened for debugging.

Additionally this ports the "exclude features" feature from the Apple testsuite to the Android one. This avoids having to exclude numerous test files for the same reason.

For iOS, this adds an "allowlist" to ensure we aren't excluding modern tests which use "const" but don't rely on its actual behavior.

## Test Plan

The existing testsuite should pass.


tmikov:
Hi, sorry for the delayed response. Unfortunately we have concluded that the current Intl implementations are a dead end, so we are not investing in them. Please see https://github.com/facebook/hermes/discussions/1211.

PR #1259, opened 2024-01-17
Fix `error: signed shift result`

r-barnes:
## Summary

Fixes
```
hermes/Support/StringTableEntry.h:19:44: error: signed shift result (0x80000000) sets the sign bit of the shift expression's type ('int') and becomes negative [-Werror,-Wshift-sign-overflow]
```
which allows the `-Wshift-sign-overflow` flag to be set.

## Test Plan

Standard CI

### DIFF
--- a/include/hermes/Support/StringTableEntry.h
+++ b/include/hermes/Support/StringTableEntry.h
@@ -16,7 +16,7 @@ namespace hermes {
 
 /// An entry in the string table inside the ConsecutiveStringStorage.
 class StringTableEntry {
-  static constexpr uint32_t UTF16_MASK = 1 << 31;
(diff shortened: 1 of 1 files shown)


include/hermes/Support/StringTableEntry.h
```diff
@@ -16,7 +16,7 @@ namespace hermes {
 
 /// An entry in the string table inside the ConsecutiveStringStorage.
 class StringTableEntry {
-  static constexpr uint32_t UTF16_MASK = 1 << 31;
+  static constexpr uint32_t UTF16_MASK = static_cast<uint32_t>(1) << 31;
```
tmikov, line 19: We prefer C-style casts for trivial cases like this. `(uint32_t)1`

PR #1262, opened 2024-01-18
Added support for providing bytecode generation option to BytecodeProviderFromSrc

rFlex:
## Summary

This change adds support for providing the bytecode generation options to the `BytecodeProviderFromSrc` API. This makes it possible to strip the function names and debug information when using the high level `BytecodeProviderFromSrc` API without having to re-write what `BytecodeProviderFromSrc` or `CompilerDriver` are already doing. For context, we currently statically embed Hermes into one of our own CLI binary that is able to do a lot of different compile related tasks for us, as such we are not using the Hermes CLI API. We still would like to be able to compile modules as byte code, and the only thing missing from `BytecodeProviderFromSrc` to be at parity with the `CompilerDriver` CLI option that does bytecode generation was the ability to configure the bytecode generation phase.

## Test Plan

Tests pass

### DIFF
--- a/include/hermes/BCGen/HBC/BytecodeProviderFromSrc.h
+++ b/include/hermes/BCGen/HBC/BytecodeProviderFromSrc.h
@@ -95,6 +95,9 @@ class BCProviderFromSrc final : public BCProviderBase {
   ///     and this is non-null, invoke this callback with the IR module to
   ///     perform optimizations. This allows us to defer the decision of
   ///     whether to link all optimizations to the caller.
+  /// \param bytecodeGenerationOptions if provided, will be used during the
+  ///     bytecode generation phase, otherwise a set of default options will
+  ///     be used.
   ///
   /// \return a BCProvider and an empty error, or a null BCProvider and an error
   ///     message (if diagHandler was provided, the error message is "error").
@@ -107,7 +110,8 @@ class BCProviderFromSrc final : public BCProviderBase {
       const ScopeChain &scopeChain,
       SourceErrorManager::DiagHandlerTy diagHandler,
       void *diagContext,
-      const std::function<void(Module &)> &runOptimizationPasses);
+      const std::function<void(Module &)> &runOptimizationPasses,
+      llvh::Optional<BytecodeGenerationOptions> bytecodeGenerationOptions);
 
  public:
   static std::unique_ptr<BCProviderFromSrc> createBCProviderFromSrc(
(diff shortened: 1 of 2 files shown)


include/hermes/BCGen/HBC/BytecodeProviderFromSrc.h
```diff
@@ -173,7 +177,8 @@ class BCProviderFromSrc final : public BCProviderBase {
       const ScopeChain &scopeChain,
       SourceErrorManager::DiagHandlerTy diagHandler = nullptr,
       void *diagContext = nullptr,
-      const std::function<void(Module &)> &runOptimizationPasses = {});
+      const std::function<void(Module &)> &runOptimizationPasses = {},
+      llvh::Optional<BytecodeGenerationOptions> bytecodeGenerationOptions = {});
```
neildhar, line 181: nit: we can make this more explicit
```suggestion
      llvh::Optional<BytecodeGenerationOptions> bytecodeGenerationOptions = llvh::None);
```

include/hermes/BCGen/HBC/BytecodeProviderFromSrc.h
```diff
@@ -95,6 +95,9 @@ class BCProviderFromSrc final : public BCProviderBase {
   ///     and this is non-null, invoke this callback with the IR module to
   ///     perform optimizations. This allows us to defer the decision of
   ///     whether to link all optimizations to the caller.
+  /// \param bytecodeGenerationOptions if provided, will be used during the
```
neildhar, line 98: Let's rename this to `defaultBytecodeGenerationOptions`, and document that flags specified in compileFlags may override conflicting flags specified here. That way, the behaviour is clear for cases where they may conflict (like format, verifyIR, etc.)

lib/BCGen/HBC/BytecodeProviderFromSrc.cpp
```diff
@@ -240,10 +243,15 @@ BCProviderFromSrc::createBCProviderFromSrcImpl(
     runOptimizationPasses(M);
 
   BytecodeGenerationOptions opts{compileFlags.format};
```
neildhar, line 245: As described above, we should make the new parameter provide the defaults, and keep the existing code specifying particular fields. When organised this way, the parameter likely doesn't even need to be optional.

neildhar (commented):
Looks good conceptually, just some comments to make the interface simpler and more clearly defined.

include/hermes/BCGen/HBC/BytecodeProviderFromSrc.h
rFlex: cool, I I was wondering if there was an equivalent of `std::nullopt` for `llvh::Optional`

include/hermes/BCGen/HBC/BytecodeProviderFromSrc.h
rFlex: sounds good, updated!

neildhar (approved):
LGTM

PR #1315, opened 2024-02-18
[Windows] Fix incorrect use of GetProcessMemoryInfo's return value.

mylhyz:
## Summary

According to the API documentation [GetProcessMemoryInfo function (psapi.h)
](https://learn.microsoft.com/en-us/windows/win32/api/psapi/nf-psapi-getprocessmemoryinfo)

> If the function succeeds, the return value is nonzero.
> If the function fails, the return value is zero. To get extended error information, call [GetLastError](https://learn.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror).

## Test Plan

- Turn on ShouldRecordStats ( GCConfig.h )
- Be able to get the actual value instead of 0


tmikov:
Thank you for this! Yet another example of how harmful `auto` can be in C++.

PR #1330, opened 2024-02-28
Allow using `babel-plugin-syntax-hermes-parser` with Babel 8 alpha

nicolo-ribaudo:
## Summary

The plugin currently requires Babel 7. Babel 8 alpha has no changes to the API that the plugin is using, so it will work if we allow it.

Note that there are some AST breaking changes in Babel 8, so the plugin will have to be updated to fully support it: https://next.babeljs.io/docs/v8-migration-api#ast-changes. However, this PR at least allows experimenting with the plugin in Babel 8 (and making sure that everything else works) for people not affected by those AST changes.

## Test Plan

Existing tests pass.


pieterv:
@nicolo-ribaudo thanks so much for the patch! We should be able to cut a new release soon.

nicolo-ribaudo:
Thank you! :)

pieterv:
@nicolo-ribaudo This was released in version 0.20.1: https://github.com/facebook/hermes/blob/main/tools/hermes-parser/js/CHANGELOG.md

PR #1358, opened 2024-03-19
feat: add visionOS support

okwasniewski:
## Summary

This PR adds support for visionOS. 

Note: Unreleased version of [CMake 3.28.4](https://gitlab.kitware.com/cmake/cmake/-/milestones/144#tab-issues) is required to build for visionOS. We might need to wait until it gets released.

### Things to sort out:
- [x] Add CI 

## Test Plan

Build test app for visionOS

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,7 +40,7 @@ workflows:
 # Default settings for Apple jobs (apple-runtime, test-apple-runtime)
 apple_defaults: &apple_defaults
   macos:
-    xcode: 13.4.1
+    xcode: 15.3
   resource_class: macos.m1.medium.gen1
   working_directory: ~/hermes
   environment:
@@ -168,7 +168,10 @@ jobs:
       - run:
           name: Install dependencies
           command: |
-            brew install cmake ninja
+            brew install ninja
+            # TODO: See comment in build-apple-runtime.
+            brew update
+            brew install cmake
       - run:
           name: Build the test application
           command: |
@@ -189,8 +192,17 @@ jobs:
             xcodebuild test \
               -workspace ApplePlatformsIntegrationTests.xcworkspace \
               -configuration Debug \
-              -destination 'platform=iOS Simulator,name=iPhone 11' \
+              -destination 'platform=iOS Simulator,name=iPhone 15' \
               -scheme ApplePlatformsIntegrationMobileTests
+          working_directory: test/ApplePlatformsIntegrationTestApp
+      - run:
+          name: Test Apple Vision application
+          command: |
+            xcodebuild test \
+              -workspace ApplePlatformsIntegrationTests.xcworkspace \
+              -configuration Debug \
+              -destination 'platform=visionOS Simulator,name=Apple Vision Pro' \
+              -scheme ApplePlatformsIntegrationVisionOSTests
           working_directory: test/ApplePlatformsIntegrationTestApp
 
   build-apple-runtime:
@@ -207,8 +219,13 @@ jobs:
       - run:
           name: Install dependencies
           command: |
-            brew install cmake ninja
+            brew install ninja
             sudo gem install cocoapods
+            # TODO: Building for Apple Vision Pro requires a newer version of
+            # CMake than is available by default. Remove "brew update" once the
+            # CircleCI image contains CMake > 3.29.2.
+            brew update
+            brew install cmake
       - run:
           name: Build the iOS frameworks
           command: ./utils/build-ios-framework.sh
@@ -222,6 +239,8 @@ jobs:
             - ~/hermes/build_catalyst
             - ~/hermes/build_iphonesimulator
             - ~/hermes/build_macosx
+            - ~/hermes/build_xros
+            - ~/hermes/build_xrsimulator
             - ~/hermes/destroot
 
   package-apple-runtime:
@@ -236,8 +255,10 @@ jobs:
       - run:
           name: Install dependencies
           command: |
-            brew install cmake ninja
             sudo gem install cocoapods
+            # TODO: See comment in build-apple-runtime.
+            brew update
+            brew install cmake
       - run:
           name: Package the framework
           command: |
@@ -265,7 +286,7 @@ jobs:
 
   macos:
     macos:
-      xcode: 13.4.1
+      xcode: 13.4.1 
     steps:
       - checkout:
           path: hermes
@@ -303,7 +324,7 @@ jobs:
 
   test-macos:
     macos:
-      xcode: 13.4.1
+      xcode: 13.4.1 
     steps:
       - checkout:
           path: hermes
@@ -680,7 +701,7 @@ jobs:
 
   test-macos-test262:
     macos:
-      xcode: 13.4.1
+      xcode: 13.4.1 
     steps:
       - checkout:
           path: hermes
--- a/hermes-engine.podspec
+++ b/hermes-engine.podspec
@@ -24,13 +24,14 @@ Pod::Spec.new do |spec|
   # The podspec would be serialized to JSON and people will download prebuilt binaries instead of the source.
   # TODO(use the hash field as a validation mechanism when the process is stable)
   spec.source      = ENV['hermes-artifact-url'] ? { http: ENV['hermes-artifact-url'] } : { git: "https://github.com/facebook/hermes.git", tag: "v#{spec.version}" }
-  spec.platforms   = { :osx => "10.13", :ios => "11.0" }
+  spec.platforms   = { :osx => "10.13", :ios => "12.0", :visionos => "1.0" }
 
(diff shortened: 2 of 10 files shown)


matthargett:
both cmake 3.28.4 and 3.29 are released! https://cmake.org/download/

tmikov:
Out of curiosity, what functionality was needed in CMake?

hermes-engine.podspec
```diff
@@ -24,13 +24,14 @@ Pod::Spec.new do |spec|
   # The podspec would be serialized to JSON and people will download prebuilt binaries instead of the source.
```
tmikov, line 24: I don't think React Native uses this file. It is pretty outdated, I think.

matthargett:
> Out of curiosity, what functionality was needed in CMake?

There were a few bugs/enhancements like this one when linking a simulator binary versus a real device binary:
https://gitlab.kitware.com/cmake/cmake/-/issues/25188

hermes-engine.podspec
okwasniewski: It was needed to make the internal hermes test's work and properly build in Xcode

neildhar:
Thanks for working on this, most of the changes seem safe and in line with the existing implementation (apart from the tests failures). I know this isn't quite ready yet, but we took a look and wanted to share some thoughts.

1. Why do we need to edit the CircleCI Android job?
2. Can we avoid using `brew update`? We deliberately disable automatically running brew update in our CircleCI jobs because we have had problems with it in the past. Could we omit that step? (potentially by using a newer version of the image if necessary)
3. What happens if an older version of CMake is used? If CMake doesn't already provided a descriptive error, perhaps we should add one.

okwasniewski:
Hey @neildhar,

1. For some reason when I updated the file Android job started failing with this error, I followed suggestions from the CircleCI blog post and set it to `default`:
CleanShot 2024-03-25 at 11 40 19@2x
2. I've updated to Xcode 15.3 image let's see if this will pull newer version of Cmake
3. The older version of CMake contained a bug that lead to always building a version for the real device which wouldn't allow us to run the Xcode test. If you prefer I can remove Xcode test

okwasniewski:
So without a brew upgrade, we are installing CMake 3.28.3 (and 3.28.4) is required for Apple Vision 😕

tmikov:
The problem with running `brew update` is that we end up potentially performing every test run with different versions of tools. 
- Can we specify explicit versions to `brew install`?
- If we are building for VisionOS Simulator, can we add a cmake version check. Something like:
```
if(CMAKE_VERSION VERSION_LESS 3.28.4)
  message("VisionOS Simulator requires CMake version >= 3.28.4")
endif()
```

neildhar:
@okwasniewski I think the Android image issue should be addressed by #1364, so you should be able to remove those changes from this PR

okwasniewski:
@neildhar I've rebased the PR - for some reason everything works locally but `pod install` fails on the CI. 

CleanShot 2024-04-02 at 12 34 07@2x

Also regarding `brew update` we can wait for next circleci image to be released which will include a newer version of CMake as its not possible to install a specific version newer than what homebrew has set in its current version

lib/Platform/Logging.cpp
```diff
@@ -11,6 +11,7 @@
 #include <android/log.h>
 #elif defined(__APPLE__)
 #include <os/log.h>
+#include <stdio.h>
```
okwasniewski, line 14: This was the source of issues 😄 `vsnprintf` function wasn't automatically imported for visionOS

okwasniewski:
Hey @neildhar, @tmikov 

I finally figured out why the build is failing and it's now all green. The last thing left to sort out is the homebrew upgrade. I've removed `HOMEBREW_NO_AUTO_UPDATE` env variable as it was linking to an issue from 2019 (https://discuss.circleci.com/t/brew-install-fails-while-updating/32992) do you think it's okay to remove it?

lib/Platform/Logging.cpp
matthargett: from what I read in the docs, it's probably also a problem on macOS 11.3+ toolchains. weird they'd make such a fundamental change to their core libc headers after so many years 🤷

neildhar:
Hey @okwasniewski, thanks for getting everything up and passing. Regarding the Homebrew updates, I propose that we leave `HOMEBREW_NO_AUTO_UPDATE` in the file. Instead, we should split up installing the homebrew packages into two steps, where the second step only installs CMake, pinned to whatever version we need. Then we should run `brew update` only before the CMake installation, with a comment explaining why.

That way, all packages will continue to be installed at a deterministic version.

okwasniewski:
@neildhar I've applied your comment. CMake is now installed separately.

.circleci/config.yml
```diff
@@ -5,6 +5,16 @@ orbs:
...
 
+commands:
+  # Install latest CMake version, needed for building Hermes for visionOS which requires at least CMake 3.29. 
+  brew-install-cmake:
+    steps:
+      - run:
+          name: Install CMake
+          command: |
+            brew update
+            brew install cmake
```
neildhar, line 16: Can we pin the CMake version here?
```suggestion
            brew install cmake@3.29.0
```

okwasniewski: It needs to be 3.29.2, I'll update this shortly

okwasniewski: Updated!

okwasniewski: It looks like it doesn't work :( `brew install cmake@3.29.2` fails..

From Homebrew docs: 

> Homebrew is a [rolling release](https://en.wikipedia.org/wiki/Rolling_release) package manager so it does not support installing arbitrary older versions of software. If your software needs specific pinned versions, consider [whalebrew](https://github.com/whalebrew/whalebrew) lines in your Brewfile to install [Docker](https://www.docker.com/) containers.

okwasniewski:
Hey! Is there anything I can do to help you resolve those internal build issues?

neildhar:
Hey @okwasniewski, I tried doing some cleanup internally to minimise changes but can't export them back out to your PR.

Could you apply this patch to the CI config? It addresses some lints and removes changes in jobs that do not build for visionOS.
```
@@ -5,16 +5,6 @@
   android: circleci/android@2.0.3
   node: circleci/node@4.7.0

-commands:
-  # Install latest CMake version, needed for building Hermes for visionOS which requires at least CMake 3.29.
-  brew-install-cmake:
-    steps:
-      - run:
-          name: Install CMake
-          command: |
-            brew update
-            brew install cmake
-
 workflows:
   version: 2
   build:
@@ -176,11 +166,6 @@
       - restore_cache:
           key: v4-repo-{{ .Environment.CIRCLE_SHA1 }}
       - run:
-          name: Install dependencies
-          command: |
-            brew install ninja
-      - brew-install-cmake
-      - run:
           name: Build the test application
           command: |
             pod install
@@ -227,9 +212,12 @@
       - run:
           name: Install dependencies
           command: |
-            brew install ninja
             sudo gem install cocoapods
-      - brew-install-cmake
+            # TODO: Building for Apple Vision Pro requires a newer version of
+            # CMake than is available by default. Remove "brew update" once the
+            # CircleCI image contains CMake > 3.29.2.
+            brew update
+            brew install cmake
       - run:
           name: Build the iOS frameworks
           command: ./utils/build-ios-framework.sh
@@ -259,9 +247,10 @@
       - run:
           name: Install dependencies
           command: |
-            brew install cmake
             sudo gem install cocoapods
-      - brew-install-cmake
+            # TODO: See comment in build-apple-runtime.
+            brew update
+            brew install cmake
       - run:
           name: Package the framework
           command: |
@@ -289,7 +278,7 @@

   macos:
     macos:
-      xcode: 15.3
+      xcode: 13.4.1
     steps:
       - checkout:
           path: hermes
@@ -327,12 +316,15 @@

   test-macos:
     macos:
-      xcode: 15.3
+      xcode: 13.4.1
     steps:
       - checkout:
           path: hermes
-      - brew-install-cmake
       - run:
+          command: |
+            brew install cmake
+      - run:
           name: Run MacOS regression tests in debug mode
           command: |
             cmake -S hermes -B build -GXcode
@@ -701,7 +693,7 @@

   test-macos-test262:
     macos:
-      xcode: 15.3
+      xcode: 13.4.1
     steps:
       - checkout:
           path: hermes
```

okwasniewski:
Hey @neildhar, sorry for the delay I was on time off. I've updated the PR, hopefully the tests will be green now 🤞

.circleci/config.yml
```diff
@@ -165,10 +165,6 @@ jobs:
       - checkout
       - restore_cache:
           key: v4-repo-{{ .Environment.CIRCLE_SHA1 }}
-      - run:
```
neildhar, line 168: I apologise, I thought testing didn't require CMake, but it looks like pod install is looking for it. We should add the same thing here with a comment.

lib/Platform/Logging.cpp
douglowder: @okwasniewski @matthargett it is the same for tvOS

Issue #1350, opened 2024-03-13
Using Intl with standalone Hermes app


manuelpuyol:
👋 I've been experimenting running JS using Hermes as a standalone binary in Linux and wanted to use `Intl.DateTimeFormat`, but looks like `Intl` support is tied to Android/iOS runtimes.

Is there any way to use `Intl` in a Linux runtime without having to use polyfills?

I created this simple gist to reproduce the current `Intl` behavior https://gist.github.com/manuelpuyol/dbb1e41c748dbaf80da3ca3556648611

neildhar:
Hey @manuelpuyol, we're aware of this limitation, and have a plan to address Intl support more comprehensively in the future by migrating to ICU (see #1211). Once we do that, we will be able to support Intl across platforms, including in a standalone linux application.

In the interim, if you're up for getting your hands dirty, Microsoft maintains an implementation for Windows that is also based on ICU (see [here](https://github.com/microsoft/hermes-windows/tree/main/lib/Platform/Intl)). It should be possible to pull in that implementation and port it to work on Linux. We would also welcome a PR to make this change in Hermes, since migrating to ICU is part of our plan for Intl support.

manuelpuyol:
great! thanks for pointing that windows implementation, I'll take a look and see if that's something I can PR!

PR #1361, opened 2024-03-22
Implement `getCanonicalLocales` in non apple/android environments

manuelpuyol:
Related to https://github.com/facebook/hermes/issues/1350 and https://github.com/facebook/hermes/discussions/1211

## Summary

As an initial step for `Intl` backed by ICU, I'm implementing the `getCanonicalLocales` function using BCP47Parser. The implementation is exactly the same as Apple's implementation

## Test Plan

This change should be covered by the `get-canonical-locales.js` test.


neildhar:
Hey @manuelpuyol, thanks for putting this together! The main high level suggestion I have here is to use the existing unicode tag parsing and canonicalisation functionality we use for the Apple Intl implementation. In particular, take a look at `BCP47Parser.cpp`, and how it is used in our existing Apple implementation in `PlatformIntlApple.mm`. You may be able to just copy the implementation over from the Apple platform. (we're planning to unify these in the longer term)

That should also solve the ICU issue you shared, since all the tag parsing and canonicalisation will be done by us.

manuelpuyol:
hey @neildhar! I used Apple's implementation and reverted the ICU change.
I'm not sure what is going on with  `test-e2e` though, `compileReleaseKotlin` is failing but doesn't look related to my changes and I see it failing on other PRs

lib/Platform/Intl/CMakeLists.txt
```diff
@@ -22,6 +22,10 @@ if(HERMES_ENABLE_INTL)
     set_target_properties(hermesPlatformIntl PROPERTIES UNITY_BUILD false)
     target_compile_options(hermesPlatformIntl PRIVATE -fobjc-arc)
   else()
-    add_hermes_library(hermesPlatformIntl STATIC PlatformIntlDummy.cpp LINK_LIBS hermesPublic)
+    add_hermes_library(hermesPlatformIntl STATIC PlatformIntlICU.cpp
+        LINK_LIBS
+        hermesBCP47Parser
+        hermesPublic
+    )
```
neildhar, line 29: We need to make sure we link against ICU

```suggestion
    add_hermes_library(hermesPlatformIntl STATIC PlatformIntlICU.cpp
        LINK_LIBS
        hermesBCP47Parser
        hermesPublic
    )
    hermes_link_icu(hermesPlatformIntl)
```

neildhar (commented):
LGTM, looking at your other diff, it looks like `getCanonicalLocales` might belong in `PlatformIntlShared.cpp`, but I'll leave it up to you.

We should also tweak the logic here to be more robust:
```
```

lib/Platform/Intl/CMakeLists.txt
manuelpuyol: we aren't using anything from ICU in this implementation, should we still link it?

neildhar: I suppose not in this particular PR. If you'd prefer to do it in the next PR once ICU is being used, we can just do that.

manuelpuyol: cool, I'm linking it in the DateTimeFormat PR so we should be good :)

lib/Platform/Intl/PlatformIntlICU.cpp
```diff
@@ -16,19 +17,63 @@ using namespace ::hermes;
...
+namespace {
 
+// https://tc39.es/ecma402/#sec-canonicalizelocalelist
+vm::CallResult<std::vector<std::u16string>> canonicalizeLocaleList(
+    vm::Runtime &runtime,
+    const std::vector<std::u16string> &locales) {
+  // 1. If locales is undefined, then a. Return a new empty list
+  if (locales.empty()) {
+    return std::vector<std::u16string>{};
+  }
```
neildhar, line 29: I'm about to merge this and just wanted to note two minor things to fix up. 

1. We use `///` for doc-comments, and it is generally preferable to use links to a specific version of the spec (as is done in the Apple version).
2. It looks like there's an extra line here that isn't in the Apple version. Note that given how the function is written, this check is not needed.

```suggestion
/// https://402.ecma-international.org/8.0/#sec-canonicalizelocalelist
vm::CallResult<std::vector<std::u16string>> canonicalizeLocaleList(
    vm::Runtime &runtime,
    const std::vector<std::u16string> &locales) {
  // 1. If locales is undefined, then
  //   a. Return a new empty List
  // Not needed, this validation occurs closer to VM in 'normalizeLocales'.
  // 2. Let seen be a new empty List.
  std::vector<std::u16string> seen;
```

PR #1362, opened 2024-03-27
Draft: Implement DateTimeFormat.format for non android/ios environments

manuelpuyol:
Related to https://github.com/facebook/hermes/issues/1350 and https://github.com/facebook/hermes/discussions/1211
Follow up of https://github.com/facebook/hermes/pull/1361

## Summary

I'd like to expand non iOS/android support of `Intl`. In this case, I'm implementing both `DateTimeFormat.prototype.resolvedOptions` and `DateTimeFormat.prototype.format`. The implementation is a mix from the Apple and hermes-windows' implementations.

- created `PlatformIntlShared` with helper functions that are exactly the same between Apple and ICU
- Implemented `DateTimeFormatICU` based on `DateTimeFormatApple`
    - Used an ICU equivalent whenever NS was used for Apple

## Test Plan

Testing against `test262/test/intl402/DateTimeFormat`


neildhar:
Hey @manuelpuyol, it's exciting to see this progress! I noticed you're sharing code differently than the original MS implementation, and I wanted to suggest a further slight tweak that will allow much more code to live in `PlatformIntlShared.cpp`. 

This is just something to consider as you're thinking about it, but we can move forward with this PR regardless of whether you adopt it.

The main idea is to have `PlatformIntlShared` as a middle layer between the `PlatformIntl` and `PlatformIntlICU`. So instead of having `PlatformIntlICU` call into `PlatformIntlShared` for common dependencies, we have `PlatformIntlShared` call into `PlatformIntlICU` for platform specific code. That makes it easy to share the code, since a platform only has to implement a simple well defined API.

So using `Collator` as an example, it would look roughly like:

### PlatformIntlShared.h:
```
#include "hermes/Platform/Intl/PlatformIntl.h"

class CollatorShared : public Collator {
 public:
  CollatorShared() = default;

  vm::ExecutionStatus initialize(
      vm::Runtime &runtime,
      const Options &options) noexcept;

  // Just an example, the signature might need to be different.
  void platformInitialize();

  Options resolvedOptions() noexcept;

  double compare(const std::u16string &x, const std::u16string &y) noexcept;

 private:
  std::u16string locale_;
  std::u16string usage_;
  std::u16string collation_;
  std::u16string caseFirst_;
  std::u16string sensitivity_;
  bool numeric_;
  bool ignorePunctuation_;
};
```

### PlatformIntlShared.cpp
```
vm::ExecutionStatus CollatorApple::initialize(
    vm::Runtime &runtime,
    const Options &options) noexcept {
  // Do platform agnostic init first.
  platformInitialize();
}
```

### PlatformIntlICU.cpp
```
class CollatorICU : public CollatorShared {
 public:
  CollatorICU() = default;

  // Just an example, the signature might need to be different.
  void platformInitialize();

  double compare(const std::u16string &x, const std::u16string &y) noexcept;

 private:
   // ICU specific fields
};

double CollatorShared::compare(
    const std::u16string &x,
    const std::u16string &y) noexcept {
  return static_cast<CollatorICU *>(this)->compare(x, y);
}

void CollatorShared::platformInitialize() noexcept {
  return static_cast<CollatorICU *>(this)->platformInitialize();
}
```

manuelpuyol:
👋 @neildhar 

I've been thinking more about the "Shared" implementation... If the goal is to migrate everything to ICU in the future, won't the Apple and Android implementations be deprecated? At that point having a shared library may be more burdensome that helpful since it does introduce extra code complexity. So maybe it's worth getting rid of `PlatformIntlShared` and putting all that code back in `PlatformIntlICU`.

However, if you feel like that's a distant future, I'd be happy to work on a better code sharing strategy, though I think that'd be more appropriate in a follow-up PR if that's ok for you :)

neildhar:
@manuelpuyol Of course, let's worry about that later

manuelpuyol:
Hey @neildhar, to be honest Ijust decided to follow hermes-windows approach but don't have the full background on that decision.

But, given that the DateTimeFormat frals with `char16_t` and ICU methods expect a `char`, what would you suggest as the best way to handle that conversion?

lib/Platform/Intl/PlatformIntlICU.cpp
```diff
@@ -120,40 +419,782 @@ double Collator::compare(
...
+  // 41. Set dateTimeFormat.[[Pattern]] to pattern.
+  // 42. Set dateTimeFormat.[[RangePatterns]] to rangePatterns.
+  // 43. Return dateTimeFormat.
+  dateTimeFormatter_ = getUDateFormatter(runtime);
+  return vm::ExecutionStatus::RETURNED;
+}
+
+// gets the UDateFormat with options set in initialize
+UDateFormat *DateTimeFormatICU::getUDateFormatter(vm::Runtime &runtime) {
+  static std::u16string eLong = u"long", eShort = u"short", eNarrow = u"narrow",
```
neildhar, line 851: Can we align this with `DateTimeFormatApple::initializeNSDateFormatter`? In particular, using `static constexpr std::u16string_view` instead of this.

lib/Platform/Intl/PlatformIntlICU.cpp
```diff
@@ -120,40 +419,782 @@ double Collator::compare(
 }
 
 namespace {
-struct DateTimeFormatDummy : DateTimeFormat {
-  DateTimeFormatDummy(const char16_t *l) : locale(l) {}
-  std::u16string locale;
+// Implementation of
+// https://402.ecma-international.org/8.0/#datetimeformat-objects
```
neildhar, line 423:
```suggestion
/// Implementation of
/// https://402.ecma-international.org/8.0/#datetimeformat-objects
```

lib/Platform/Intl/PlatformIntlICU.cpp
```diff
@@ -120,40 +419,782 @@ double Collator::compare(
...
+  // and "h23" start with hour 0 and go up to 11 and 23 respectively. "h12" and
+  // "h24" start with hour 1 and go up to 12 and 24. [[HourCycle]] is only used
+  // when [[Hour]] is not undefined.
+  std::optional<std::u16string> hourCycle_;
+  // [[DateStyle]], [[TimeStyle]] are each either undefined, or a String value
+  // with values "full", "long", "medium", or "short".
+  std::optional<std::u16string> dateStyle_;
+  std::optional<std::u16string> timeStyle_;
+  std::string locale8_;
+  UDateFormat *dateTimeFormatter_;
```
neildhar, line 492: Please add a doc-comment for these last two fields.

It also seems like `dateTimeFormatter_` here is being leaked? We should have a destructor for `DateTimeFormatICU` that does the necessary cleanup.

neildhar (commented):
In the interest of keeping the review lightweight, we can punt on removing the UTF-8 conversion, since that would require auditing and validating all of the places where it is used.

But to answer your question, for ASCII strings, converting between UTF-16 and UTF-8 just involves copying over the characters.

lib/Platform/Intl/PlatformIntlICU.cpp
```diff
@@ -120,40 +419,782 @@ double Collator::compare(
...
+        nullptr,
+        -1,
+        &bestpattern[0],
+        patternLength,
+        &status);
+  }
+}
+
+vm::CallResult<std::u16string> DateTimeFormatICU::getDefaultHourCycle(
+    vm::Runtime &runtime) {
```
neildhar, line 1073: The default hour cycle depends on the user specified locale, so we shouldn't be using the default locale to compute this.

lib/Platform/Intl/PlatformIntlICU.cpp
```diff
@@ -120,40 +419,782 @@ double Collator::compare(
...
+          break;
+        case 'k':
+          return u"h24";
+          break;
+      }
+    }
+  }
+
+  // There should always be a default hour cycle, return an exception if not
+  return vm::ExecutionStatus::EXCEPTION;
```
neildhar, line 1114: This is incorrect, and may result in a crash if we hit this case. If we need to raise an exception here, we should do it with `runtime.raiseError` as we do in other places in this file.

lib/Platform/Intl/PlatformIntlICU.cpp
manuelpuyol: we use `locale8_`, which comes from `resolveLocale`, to compute it

neildhar:
Hey @manuelpuyol, I just pulled this in and tried building, but it looks like it doesn't compile. Have you been able to test this on your machine?

manuelpuyol:
@neildhar that's... interesting
I can compile it locally and looks like circleci can too? What errors are you getting?

neildhar:
CircleCI doesn't currently build with `HERMES_ENABLE_INTL` on linux, so it isn't actually building the new implementation. When I turn it on and build on linux, here are some of the errors I see:

```
hermes/lib/Platform/Intl/PlatformIntlICU.cpp:275:37: error: invalid conversion from ‘const short unsigned int*’ to ‘const char16_t*’ [-fpermissive]
       auto upper = toASCIIUppercase(zoneId);
```

```
hermes/lib/Platform/Intl/PlatformIntlICU.cpp:1029:7: error: invalid conversion from ‘__gnu_cxx::__alloc_traits<std::allocator<char16_t>, char16_t>::value_type*’ {aka ‘char16_t*’} to ‘const UChar*’ {aka ‘const short unsigned int*’} [-fpermissive]
       &skeleton[0],
```

```
hermes/lib/Platform/Intl/PlatformIntlICU.cpp:1041:9: error: invalid conversion from ‘__gnu_cxx::__alloc_traits<std::allocator<char16_t>, char16_t>::value_type*’ {aka ‘char16_t*’} to ‘const UChar*’ {aka ‘const short unsigned int*’} [-fpermissive]
         &skeleton[0],
```

```
hermes/lib/Platform/Intl/PlatformIntlICU.cpp:1044:9: error: invalid conversion from ‘__gnu_cxx::__alloc_traits<std::allocator<char16_t>, char16_t>::value_type*’ {aka ‘char16_t*’} to ‘UChar*’ {aka ‘short unsigned int*’} [-fpermissive]
         &bestpattern[0],
```

```
hermes/lib/Platform/Intl/PlatformIntlICU.cpp:1060:9: error: invalid conversion from ‘__gnu_cxx::__alloc_traits<std::allocator<char16_t>, char16_t>::value_type*’ {aka ‘char16_t*’} to ‘const UChar*’ {aka ‘const short unsigned int*’} [-fpermissive]
         &bestpattern[0],
```

Of these, the first one seems the most odd, since we are casting between different types.

manuelpuyol:
hmm I wonder if that's being caused by different ICU versions, do you know which version you are using to build?

neildhar:
Taking a closer look, I think this might be a problem of differing definitions of char16_t. This build was with GCC, are you building with Clang?

manuelpuyol:
Looking at ICU's source, they did change the return type of `unext` and technically Hermes supports both versions since it looks for ICU > 52:

version 66:
https://github.com/unicode-org/icu/blob/3ca5d8ca1559fb52d08bad61e350f0285f3edb37/icu4c/source/common/unicode/strenum.h#L157

version 57:
https://github.com/unicode-org/icu/blob/b7c2034642ce01ba6ffb9418f26234efb2549d08/icu4c/source/common/unicode/strenum.h#L154

manuelpuyol:
> Taking a closer look, I think this might be a problem of differing definitions of char16_t. This build was with GCC, are you building with Clang?

I'm using gcc

manuelpuyol:
ok, looks like ICU lets us define the `UChar` type to be less platform dependent. I pushed https://github.com/facebook/hermes/pull/1362/commits/49037cebca4e8fd48ce6e4a6ab5e516898756a04 which may solve the incompatibility we are seeing

include/hermes/Platform/Intl/PlatformIntl.h
```diff
@@ -9,6 +9,8 @@
 #define HERMES_PLATFORMINTL_PLATFORMINTL_H
 
 #ifdef HERMES_ENABLE_INTL
+#define UCHAR_TYPE char16_t
```
neildhar, line 12: Let's put this in `PlatformIntlICU.cpp` right before the unicode headers with a comment explaining what it does.

I'm slightly surprised that this works fine, but I suppose it is designed to.

lib/Platform/Intl/PlatformIntlICU.cpp
```diff
@@ -7,73 +7,372 @@
...
-    const std::vector<std::u16string> &locales) {
-  // 1. Let ll be ? CanonicalizeLocaleList(locales).
-  // 2. Return CreateArrayFromList(ll).
-  return canonicalizeLocaleList(runtime, locales);
+namespace {
+/// Thread safe management of time zone names map.
+class TimeZoneNames {
+ public:
+  /// Initializing the underlying map with all known time zone names in
+  /// NSTimeZone.
```
neildhar, line 271: nit: this comment still mentions NSTimezone

lib/Platform/Intl/PlatformIntlICU.cpp
```diff
@@ -120,40 +419,787 @@ double Collator::compare(
...
+  // slots that are computed by the constructor:
+  // [[Locale]] is a String value with the language tag of the locale whose
+  // localization is used for formatting.
+  std::u16string locale_;
+  // [[Calendar]] is a String value with the "type" given in Unicode Technical
+  // Standard 35 for the calendar used for formatting.
+  std::optional<std::u16string> calendar_;
+  // [[NumberingSystem]] is a String value with the "type" given in Unicode
+  // Technical Standard 35 for the numbering system used for formatting.
+  // NOTE: Even though NSDateFormatter formats date and time using different
```
neildhar, line 457: nit: NSDateFormatter

neildhar (commented):
That fixed it, thanks!

I was able to get it building, but it looks like it's still failing tests. To start, I ran it against our `date-time-format-apple.js` test, and it looks like there might be a bug in handling `dateStyle`/`timeStyle` when only one of them is specified.

Some minor differences are expected from Apple of course, but at least that one looked like a bug.

We should also manually test this against test262 DateTimeFormat tests to verify that it works on the tests that the Apple implementation is currently passing. You can take a look at how our CI does that [here](https://github.com/facebook/hermes/blob/c2f7dcd5e4fa88b6bea12e2b705d53a7ffd54ecf/.circleci/config.yml#L701), and you should be able to narrow the runner to only run on `DateTimeFormat` tests by specifying the directory where those tests live (instead of all of `test262/test`).

manuelpuyol:
I'm gonna take a look at the apple test! As for the test262:

```
$ utils/testsuite/run_testsuite.py --test-intl -b build/bin/ ~/test262/test/intl402/DateTimeFormat/

| Total                |      176 |
| Pass                 |       50 |
| Skipped              |      125 |
```

manuelpuyol:
> To start, I ran it against our date-time-format-apple.js test, and it looks like there might be a bug in handling dateStyle/timeStyle when only one of them is specified.

mind sharing a specific example I can take a look?

neildhar:
Sure, if you run `LIT_FILTER=date-time-format-apple.js ninja check-hermes`, that should run that specific test and print failures.

manuelpuyol:
@neildhar https://github.com/facebook/hermes/pull/1362/commits/7c578953e745b995a401d6cddd59562190e90509 should fix the case when `timeStyle` or `dateStyle` is missing!

neildhar:
Minor differences are inevitable, we should relax the match patterns slightly where necessary to accommodate those differences, since they likely just come up from different ICU versions being used.

We run this test using `FileCheck`, which allows you to place a regex in the pattern (see [here](https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-regex-matching-syntax)).

It should just be a matter of turning a few characters into wildcards. If the differences are more substantial, that may indicate a bug in one of the implementations.

manuelpuyol:
just to keep track of it, these are the differences I'm seeing in that test:

```js
const date = new Date(Date.UTC(2020, 0, 2, 3, 45, 00, 30));
const oldDate = new Date(Date.UTC(1952, 0, 9, 8, 04, 03, 05));
// apple and node: 3
// ICU: 03
new Intl.DateTimeFormat('en-GB', {second: '2-digit'}).format(oldDate)

// apple: 午前3:45
// ICU and node: 3:45
new Intl.DateTimeFormat('ja-JP', {hour: 'numeric', minute: 'numeric'}).format(date);

// apple: 午前3:45
// ICU: 午前03:45
// node: 03:45
new Intl.DateTimeFormat('ja-JP', {hour: '2-digit', minute: '2-digit'}).format(date);

// apple and ICU: zh
// node: zh-CN
new Intl.DateTimeFormat('zh-CN').resolvedOptions().locale

// apple and ICU: undefined
// node: latn
new Intl.DateTimeFormat('en-US').resolvedOptions().numberingSystem;

// apple: SGT
// node and ICU: Error Invalid time zone specified
new Intl.DateTimeFormat('en-US', { timeZone: 'SGT'}).resolvedOptions().timeZone
```

manuelpuyol:
@neildhar if you are comfortable with those differences, I can push a change to the test :)

neildhar:
The Japanese date patterns seem potentially wrong. It looks like the DateTime pattern generation here doesn't quite follow the spec rules. For example, the 23 hour cycle should be using `H` and `HH` rather than `k` and `KK`.

manuelpuyol:
indeed! I updated it and now ICU outputs the same as node for that case

neildhar:
The rest seem to be genuine differences in the defaults. We don't need to solve this right away though, given that the implementation is incomplete (so we can't run tests anyway).

PR #1363, opened 2024-03-28
[Debugger] Modify ES6 Method names instead of removing them

Beanyy:
## Summary
The method names of ES6 classes were being stripped out to prevent name conflicts when there's both a function and method with identical names. However, when connecting the Debugger or Profiler, this causes all class methods to show up with the name `anonymous` instead of its actual function name.

This change modifies the ES6 method names to prefix and suffix them with '#' instead of removing it, so that the debugger and profiler give more detailed information, while still preventing the name conflicts.

Before:
Screenshot 2024-03-28 at 2 46 57 PM

After:
Screenshot 2024-03-28 at 2 42 36 PM


tmikov:
Makes sense

PR #1365, opened 2024-03-29
Implement `TypedArray.prototype.with`

robik:
## Summary

This PR implements ES2023  `TypedArray.prototype.with` method. This is a follow-up of PR #1286 which added `with` method on standard `Array` prototype.

## Test Plan

```bash
$ echo "new Int8Array([0]).with(0, 5)" | ./bin/hermes
# >> Int8Array [ 5 ]
```


robik:
Also, as an future improvement area, is it possible to determine if an array is an RValue, as in a temporary value, which reference to was not captured yet? If so, I might create a follow-up PR that if checks if target array is an RValue, and then changes it in-place for some non mutating methods so that we do not allocate new array everytime. 

This would be beneficial for cases like below:
```
const finalArray = [0, 0, 2]
  .with(1, 1)
  .toReversed();
```

tmikov:
@robik, high level optimizations like that are definitely something we are planning for in Static Hermes, but unfortunately they are almost impossible to do ahead of time in a regular untyped JS.

In regular JS, when we encounter `x.with(1, 1).toReversed()`, we don't know whether `x` is an Array, what `with()` does, whether it has been overridden by a polyfill, etc.

Things change dramatically in Static Hermes:
```
let x : any[];
...
x.with(1,1).toReversed()
```

lib/VM/JSLib/TypedArray.cpp
```diff
@@ -1746,6 +1746,126 @@ typedArrayPrototypeToLocaleString(void *, Runtime &runtime, NativeArgs args) {
   return HermesValue::encodeStringValue(*builder->getStringPrimitive());
 }
 
+inline static bool isValidIntegerIndex(
+    Runtime &runtime,
+    Handle<JSTypedArrayBase> O,
+    const HermesValue &index) {
```
avp, line 1752: Taking a `HermesValue` here is potentially error-prone. It looks like this is only ever called with a number - could we change this to just take `double` directly and avoid the potential of accidentally passing in an object pointer?

lib/VM/JSLib/TypedArray.cpp
```diff
@@ -1746,6 +1746,126 @@ typedArrayPrototypeToLocaleString(void *, Runtime &runtime, NativeArgs args) {
...
+  if (actualIndex > 0) {
+    if (LLVM_UNLIKELY(
+            JSTypedArrayBase::setToCopyOfTypedArray(
+                runtime, A, 0, self, 0, actualIndex) ==
+            ExecutionStatus::EXCEPTION)) {
+      return ExecutionStatus::EXCEPTION;
+    }
+  }
+
+  auto targetValue = runtime.makeHandle(res.getValue());
```
avp, line 1846: This handle needs to be created much earlier - right after `res` is checked for `EXCEPTION`. Otherwise we're potentially holding it across allocations.

avp (changes requested):
Thanks for the PR, it looks mostly good. I've left a couple comments regarding pointer safety that will have to be fixed before we can land it.

lib/VM/JSLib/TypedArray.cpp
robik: My idea was to encapsulate the value and type checking. If this function would accept a double, 3 if conditions would need to be moved to caller making this a one-liner. Is there some alternative for this? 🤔

PR #1366, opened 2024-03-29
Implement `TypedArray.prototype.toReversed`

robik:
## Summary

This PR implements ES2023 `TypedArray.prototype.toReversed` method. This is a follow-up of PR https://github.com/facebook/hermes/pull/1286 which added with method on standard Array prototype.

See also:
 - PR for `TypedArray.prototype.with`: #1365
 - PR for `TypedArray.prototype.toSorted`: #1367

I am creating separate PRs for each method in case there are any suggestions or change requests from your side, so that PRs are more independent.

## Test Plan

```bash
$ echo "new Int8Array([0, 1, 2, 3]).toReversed()" | ./bin/hermes
# >> Int8Array [ 3, 2, 1, 0 ]
```

### DIFF
--- a/lib/VM/JSLib/TypedArray.cpp
+++ b/lib/VM/JSLib/TypedArray.cpp
@@ -1761,8 +1761,8 @@ typedArrayPrototypeToReversed(void *, Runtime &runtime, NativeArgs args) {
   auto self = args.vmcastThis<JSTypedArrayBase>();
 
   // 3. Let len be O.[[ArrayLength]].
-  JSArrayBuffer::size_type len = self->getLength();
-
+  auto len = self->getLength();
+  auto byteLength = self->getByteLength();
   auto byteWidth = self->getByteWidth();
 
   // 4. Let A be ? TypedArrayCreateSameType(O, « 𝔽(len) »).
@@ -1775,19 +1775,19 @@ typedArrayPrototypeToReversed(void *, Runtime &runtime, NativeArgs args) {
   // 5. Let k be 0.
   JSArrayBuffer::size_type k = 0;
 
-  auto aBuffer = A->getBuffer(runtime);
-  auto srcBuffer = self->getBuffer(runtime);
+  auto dstBlock = A->getBuffer(runtime)->getDataBlock(runtime);
+  auto srcBlock = self->getBuffer(runtime)->getDataBlock(runtime);
 
   // 6. Repeat, while k < len,
-  while (k < len) {
+  while (k < byteLength) {
     // 6a. Let from be ! ToString(𝔽(length - k - 1)).
-    JSArrayBuffer::size_type from = len - k - 1;
+    size_t from = byteLength - k - byteWidth;
 
     // 6d. Perform ! Set(A, Pk, fromValue, true).
-    JSArrayBuffer::copyDataBlockBytes(runtime, aBuffer, k * byteWidth, srcBuffer, from * byteWidth, byteWidth);
+    memcpy(dstBlock + k, srcBlock + from, byteWidth);
 
     // 6e. Set k to k + 1.
-    ++k;
+    k += byteWidth;
   }
 
   return A.getHermesValue();


lib/VM/JSLib/TypedArray.cpp
```diff
@@ -1746,6 +1746,64 @@ typedArrayPrototypeToLocaleString(void *, Runtime &runtime, NativeArgs args) {
...
+  auto marker = gcScope.createMarker();
+  // 6. Repeat, while k < len,
+  while (k < len) {
+    gcScope.flushToMarker(marker);
+
+    // 6a. Let from be ! ToString(𝔽(length - k - 1)).
+    double from = len - k - 1;
+
+    // 6b. Let Pk be ! ToString(𝔽(k)).
+    kHandle = HermesValue::encodeTrustedNumberValue(k);
```
dannysu, line 1787: `kHandle` is unused other than being assigned here. Could you please remove it?

dannysu (changes requested):
Thank you for the PR! A minor comment on something caught by our internal linter.

lib/VM/JSLib/TypedArray.cpp
```diff
@@ -1746,6 +1746,61 @@ typedArrayPrototypeToLocaleString(void *, Runtime &runtime, NativeArgs args) {
...
+  // 4. Let A be ? TypedArrayCreateSameType(O, « 𝔽(len) »).
+  auto aRes = JSTypedArrayBase::allocateSpecies(runtime, self, len);
+  if (LLVM_UNLIKELY(aRes == ExecutionStatus::EXCEPTION)) {
+    return ExecutionStatus::EXCEPTION;
+  }
+  auto A = aRes.getValue();
+
+  // 5. Let k be 0.
+  double k = 0;
+  MutableHandle<> kHandle{runtime};
```
dannysu, line 1775: The unused variable is still here.

lib/VM/JSLib/TypedArray.cpp
```diff
@@ -1746,6 +1746,61 @@ typedArrayPrototypeToLocaleString(void *, Runtime &runtime, NativeArgs args) {
...
+    // 6a. Let from be ! ToString(𝔽(length - k - 1)).
+    double from = len - k - 1;
+
+    // 6c. Let fromValue be ? Get(O, from).
+    fromValueHandle =
+        JSObject::getOwnIndexed(createPseudoHandle(self.get()), runtime, from);
+
+    // 6d. Perform ! Set(A, Pk, fromValue, true).
+    if (LLVM_UNLIKELY(
+            A->setOwnIndexed(A, runtime, k, fromValueHandle) ==
```
tmikov, line 1792: This is all much slower than it has to be, because we are just reading and writing values from C arrays with known sizes, without side effects. We should access the pointers directly.

tmikov (changes requested):

lib/VM/JSLib/TypedArray.cpp
robik: Oh damn, sorry, Must have re-added it on rebase I think

lib/VM/JSLib/TypedArray.cpp
robik: Thanks for heads up! I was not sure of internal side effects, but that makes sense. Should be fixed now.

lavenzg:
For the other PRs, we should eliminate `setOwnIndexed` or `setToCopyOfTypedArray` as well, since we know exactly the the underlying data array and the source/target array buffer have the same type.

lib/VM/JSLib/TypedArray.cpp
```diff
@@ -1746,6 +1746,53 @@ typedArrayPrototypeToLocaleString(void *, Runtime &runtime, NativeArgs args) {
...
+  auto aBuffer = A->getBuffer(runtime);
+  auto srcBuffer = self->getBuffer(runtime);
+
+  // 6. Repeat, while k < len,
+  while (k < len) {
+    // 6a. Let from be ! ToString(𝔽(length - k - 1)).
+    JSArrayBuffer::size_type from = len - k - 1;
+
+    // 6d. Perform ! Set(A, Pk, fromValue, true).
+    JSArrayBuffer::copyDataBlockBytes(runtime, aBuffer, k * byteWidth, srcBuffer, from * byteWidth, byteWidth);
```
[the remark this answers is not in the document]
robik: Sure thing! Just wanted to first check if the current approach is corrrect :)

robik:
Sure thing! I just wanted to make sure that the current approach is correct first. If it is, I can apply it to other methods :)

lib/VM/JSLib/TypedArray.cpp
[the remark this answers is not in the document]
robik: WDYT about creating a static utility function for copying elements in `JSTypedArray` (where byte width is known) rather than copying bytes?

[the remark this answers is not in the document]
lavenzg: Then you have to specialize it for every valid width. Probably not necessary at the moment. It can be done later if we do need to dig more performance. For now, `memcpy` on the underlying `data_` array is good enough.

PR #1367, opened 2024-03-29
Implement `TypedArray.prototype.toSorted`

robik:
## Summary

This PR implements ES2023  `TypedArray.prototype.toSorted` method. 

See also:

 - PR for `TypedArray.prototype.toReversed`: #1366 
 - PR for `TypedArray.prototype.with`: #1365  

I am creating separate PRs for each method in case there are any suggestions or change requests from your side, so that PRs are more independent.

## Test Plan

```bash
$ echo "new Int8Array([3, 1, 2]).toSorted()" | ./bin/hermes
# >> Int8Array [ 1, 2, 3 ]
```


robik:
Considering comments from other PRs, I wonder if I could optimize `TypedArraySortModel` to use direct access to values rather than using VTable. WDYT?

PR #1408, opened 2024-05-23
Parse Member and TypeCast expressions in super class types

xleoooo:
### Summary

Previously, member expressions in parent classes like `class A extends Foo.Bar`, and type casts such as `class A extends (Foo: Bar)`, were not parseable by flow-api-translator. This change adds support.

## Test Plan

Previously unparseable files in `react-native` are now parseable

### DIFF
--- a/tools/hermes-parser/js/flow-api-translator/src/flowToFlowDef.js
+++ b/tools/hermes-parser/js/flow-api-translator/src/flowToFlowDef.js
@@ -49,6 +49,7 @@ import type {
   ObjectTypeAnnotation,
   ObjectTypeProperty,
   OpaqueType,
+  QualifiedTypeIdentifier,
   Program,
   RestElement,
   Statement,
@@ -977,6 +978,49 @@ function convertClassDeclaration(
   ];
 }
 
+function convertExpressionToIdentifier(
+  node: Expression,
+  context: TranslationContext,
+): DetachedNode<Identifier> | DetachedNode<QualifiedTypeIdentifier> {
+  if (node.type === 'Identifier') {
+    return t.Identifier({name: node.name});
+  }
+
+  if (node.type === 'MemberExpression') {
+    const {property, object} = node;
+    if (property.type === 'Identifier' && object.type !== 'Super') {
+      return t.QualifiedTypeIdentifier({
+        qualification: convertExpressionToIdentifier(object, context),
+        id: t.Identifier({name: property.name}),
+      });
+    }
+  }
+
+  throw translationError(
+    node,
+    `Expected ${node.type} to be an Identifier or Member with Identifier property, non-Super object.`,
+    context,
+  );
+}
+
+function convertSuperClassHelper(
+  detachedId: DetachedNode<Identifier | QualifiedTypeIdentifier>,
+  nodeForDependencies: ESNode,
+  superTypeParameters: ?TypeParameterInstantiation,
+  context: TranslationContext,
+): TranslatedResultOrNull<InterfaceExtends> {
+  const [resultTypeParams, typeParamsDeps] =
+    convertTypeParameterInstantiationOrNull(superTypeParameters, context);
+  const superDeps = analyzeTypeDependencies(nodeForDependencies, context);
+  return [
+    t.InterfaceExtends({
+      id: detachedId,
+      typeParameters: resultTypeParams,
+    }),
+    [...typeParamsDeps, ...superDeps],
+  ];
+}
+
 function convertSuperClass(
   superClass: ?Expression,
   superTypeParameters: ?TypeParameterInstantiation,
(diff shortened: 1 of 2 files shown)


tools/hermes-parser/js/hermes-estree/src/types.js
```diff
@@ -1560,7 +1560,7 @@ export interface InterfaceDeclaration extends BaseInterfaceDeclaration {
 
 export interface InterfaceExtends extends BaseNode {
   +type: 'InterfaceExtends';
-  +id: Identifier | QualifiedTypeIdentifier;
+  +id: Identifier | QualifiedTypeIdentifier | MemberExpression | TypeAnnotationType;
```
pieterv, line 1563: This type change is incorrect: https://astexplorer.net/#/gist/5c8ba16b70de7482e04a1e4bc9d481fb/3d30f92fa471d423d4569497a778a73eabd8f042

Both `MemberExpression` and `TypeAnnotationType` are invalid in these locations (although they may happened to print correct atm). You need to instead convert these back to the appropriate supported AST types.

tools/hermes-parser/js/flow-api-translator/src/flowToFlowDef.js
```diff
@@ -986,23 +986,43 @@ function convertSuperClass(
...
 
-  if (superClass.type !== 'Identifier') {
-    throw translationError(
-      superClass,
-      `SuperClass: Non identifier super type of "${superClass.type}" not supported`,
-      context,
-    );
+  switch (superClass.type) {
+    case 'Identifier':
+    case 'MemberExpression': {
```
pieterv, line 991: The member expression case needs to be converted to a `QualifiedTypeIdentifier`.

tools/hermes-parser/js/flow-api-translator/src/flowToFlowDef.js
```diff
@@ -986,23 +986,43 @@ function convertSuperClass(
...
+      const superDeps = analyzeTypeDependencies(superClass, context);
+      return [
+        t.InterfaceExtends({
+          id: asDetachedNode(superClass),
+          typeParameters: resultTypeParams,
+        }),
+        [...typeParamsDeps, ...superDeps],
+      ];
+    }
+    case 'TypeCastExpression': {
```
pieterv, line 1003: Its not safe to pass through the `TypeAnnotationType` because only `QualifiedTypeIdentifier` nodes are supported, nodes like `GenericTypeAnnotation` need to be converted to Identifiers.

pieterv (dismissed):
In addition to the inline comments please add test cases.

I'm also curious what you are using this tool chain for?

tools/hermes-parser/js/hermes-estree/src/types.js
xleoooo: Thanks for the review. I'm working with @huntie to improve the coverage in [public-api-test](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/__tests__/public-api-test.js).

Are there existing functions that do this conversion ? I see some similar logic in `TSDefToFlowDef`, but so far nothing that "just works". If not, I can try adding them -- just want to verify before committing time.

tools/hermes-parser/js/flow-api-translator/src/flowToFlowDef.js
```diff
@@ -977,6 +978,35 @@ function convertClassDeclaration(
...
+    return t.Identifier({name: node.name});
+  } else if (
+    node.type === 'MemberExpression' &&
+    node.property.type === 'Identifier' &&
+    node.object.type !== 'Super'
+  ) {
+    const id = t.Identifier({name: node.property.name});
+    return t.QualifiedTypeIdentifier({
+      qualification: convertExpressionToIdOrQualifiedTypeId(
+        // $FlowFixMe[incompatible-call] conditional already checks for non-Super type
```
xleoooo, line 995: not sure if I addressed this the right way, tried to convert the MemberExpression but hit this snag with Super type.

xleoooo:
Still working on those test cases, I'm having some trouble with jest, but was hoping to get a review on the source changes in the meantime.

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
```

tools/hermes-parser/js/flow-api-translator/src/flowToFlowDef.js
```diff
@@ -977,6 +978,35 @@ function convertClassDeclaration(
...
+        node.object,
+        context,
+      ),
+      id,
+    });
+  } else {
```
huntie, line 1001: nit: I'd remove this `else` block, since items are returned in each branch above.

tools/hermes-parser/js/flow-api-translator/src/flowToFlowDef.js
huntie: This seems practical to me, but prefer `// $FlowIssue` (indicating the typechecker should have been able to handle this, and it's not a fixme on this code).

```suggestion
        // $FlowIssue[incompatible-call] conditional already checks for non-Super type
```

tools/hermes-parser/js/flow-api-translator/src/flowToFlowDef.js
```diff
@@ -977,6 +978,35 @@ function convertClassDeclaration(
   ];
 }
 
+function convertExpressionToIdOrQualifiedTypeId(
```
huntie, line 981: Naming nit. The input type to `t.InterfaceExtends()` is `id`, so let's match "identifier" as a suitable generic term.

https://github.com/facebook/hermes/blob/831bcdc417493c04456d2c1ba2abbb32d4d21ffb/tools/hermes-parser/js/hermes-estree/src/types.js#L1563

```suggestion
function convertExpressionToIdentifier(
```

huntie (approved):
LGTM, with nits to action. cc @pieterv 

I've run this against the existing test suite (with `jest --watch`) and the affected tests all pass. Just to check: @xleoooo you've observed this produce an end-to-end fix?

image

xleoooo:
> Just to check: @xleoooo you've observed this produces an end-to-end fix?

yes, with some small caveat: some of the current RN source isn't valid input.

```
class FileReader extends (EventTarget(...READER_EVENTS): any) {}
```

AFAIK this is valid flow source code, but neither of the following typedefs are OK based on flow playground

```
declare class FileReader extends any {}
declare class FileReader extends (EventTarget(...READER_EVENTS): any) {}
```

So, for type cast super classes, they are only valid for `flowToFlowDef` if cast to `Identifier`. This means we'll still need to add our own typings for EventTarget, instead of casting `any`.

tools/hermes-parser/js/flow-api-translator/src/flowToFlowDef.js
```diff
@@ -977,34 +978,116 @@ function convertClassDeclaration(
...
+    }
+    case 'MemberExpression': {
+      return convertSuperClassHelper(
+        convertExpressionToIdentifier(superClass, context),
+        superClass,
+        superTypeParameters,
+        context,
+      );
+    }
+    case 'TypeCastExpression': {
```
xleoooo, line 1050: adding `AsExpression` case here causes a bunch of flow errors -- didn't look into it further

PR #1413, opened 2024-05-30
Add Intl.Collator implementation using ICU4C

robchu05:
## Summary

Add Intl.Collator implementation using ICU4C for non-Android / non-Apple platforms.

Classes without ICU / Platform API dependency:
* Constants - String constants for options bag’s names and values, valid option values, other constants.
* IntlUtils - Conversion between UTF-8 and UTF-16 ASCII strings, convert string to bool, i.e. case- insensitive match against “true”, lowercase ASCII strings.
* LocaleBCP47Object - Wrapper over a BCP47Parser::ParsedLocaleIdentifier, provides factory method to create instances from a BCP47 locale string, methods to get canonicalized locale string and locale string without extensions, and implements the CanonicalizeLocaleList() spec operation.
* OptionHelpers - Gets string, bool, and number option from options bag.

ICU dependent classes:
* Collator - Intl.Collator implementation using ICU collator API.
* Locale - Conversion between BCP47 and ICU locale strings.
* LocaleResolver - Implements ResolveLocale() and SupportedLocales() spec operations, supports both “lookup” and “best fit” locale matching. “best fit” locale matching uses ICU acceptLanguage API.

Fix memory leaks in existing DateTimeFormat implementation in PlatformIntlICU.cpp by wrapping created ICU object pointers in unique_ptr.

Migrate ResolveLocale, SupportedLocalesOf, GetCanonicalLocales API implementation to use the corresponding newly added supporting classes.

Add test cases to verify invalid locales and options input would result in throwing JS exception in collation.js. Add tests cases to verify resolution of locale extensions and options in a new file collation-resolved-options.js.

## Test Plan

Build succeeds on Ubuntu 20.04. Run the following JS tests on Ubuntu 20.04, and all pass except for a few test cases in date-time-format-apple.js due to locale data differences. The output of running date-time-format-apple.js test is the same as before this change.

```
$ ./build/bin/hermes ./hermes/test/hermes/intl/intl.js
$ ./build/bin/hermes ./hermes/test/hermes/intl/get-canonical-locales.js | ./build/bin/FileCheck --match-full-lines ./hermes/test/hermes/intl/get-canonical-locales.js
$ ./build/bin/hermes ./hermes/test/hermes/intl/collator.js | ./build/bin/FileCheck --match-full-lines ./hermes/test/hermes/intl/collator.js
$ LC_ALL=fr_FR _HERMES_TEST_LOCALE=fr_FR ./build/bin/hermes ./hermes/test/hermes/intl/collator-resolved-options.js
$ TZ=GMT ./build/bin/hermes ./hermes/test/hermes/intl/date-time-format-apple.js | ./build/bin/FileCheck --match-full-lines ./hermes/test/hermes/intl/date-time-format-apple.js
```

Run valgrind with above test runs and no memory leak detected.

Run Test262 tests for Collator, String-LocaleCompare, DateTimeFormat, Date, and Intl. All non-skipped test cases pass.

```
$ ./hermes/utils/testsuite/run_testsuite.py -b ./build/bin --test-intl -a ~/Downloads/test262/test/intl402/Collator/
...
| Total                |       61 |
| Pass                 |       45 |
| Skipped              |       16 |
| Permanently Skipped  |        0 |

$ ./hermes/utils/testsuite/run_testsuite.py -b ./build/bin --test-intl -a ~/Downloads/test262/test/intl402/String/prototype/localeCompare/
...
| Total                |       10 |
| Pass                 |        9 |
| Skipped              |        1 |
| Permanently Skipped  |        0 |

$ ./hermes/utils/testsuite/run_testsuite.py -b ./build/bin --test-intl -a ~/Downloads/test262/test/intl402/DateTimeFormat/
...
| Total                |      176 |
| Pass                 |       50 |
| Skipped              |      125 |

$ ./hermes/utils/testsuite/run_testsuite.py -b ./build/bin --test-intl -a ~/Downloads/test262/test/intl402/Date/
...
| Total                |       12 |
| Pass                 |       10 |
| Skipped              |        2 |
| Permanently Skipped  |        0 |

$ ./hermes/utils/testsuite/run_testsuite.py -b ./build/bin --test-intl -a ~/Downloads/test262/test/intl402/Intl/
...
| Total                |       66 |
| Pass                 |       20 |
| Permanently Skipped  |       27 |
```


lib/Platform/Intl/Constants.h
```diff
@@ -0,0 +1,322 @@
+/**
```
tmikov, line 1: I see a few problem with this file.

1. It is a header file but it contains definitions. Including it in more than one .cpp file will cause link errors.

2. It uses a lot of std::u16string for what appear to be constant strings. This will cause all of these to be heap allocated. Is that necessary?

3. All of the global constructors will cause startup delays. As a rule we try to avoid global constructors at all cost since they slow down initialization unconditionally. This file adds many many global constructirs.

tmikov (changes requested):

lib/Platform/Intl/Constants.h
robchu05: Thank you for the feedback, Tzvetan! I will work on updating the code to address the concerns raised.

I am thinking to replace all the struct variables here with namespaces and use ```inline constexpr auto&``` for all constant strings, per [this explanation](https://www.learncpp.com/cpp-tutorial/sharing-global-constants-across-multiple-files-using-inline-variables/). For example, change
```c++
const struct {
...
  const struct {
     std::u16string lookup_;
     std::u16string best_fit_;
  } matcher = {u"lookup", u"best fit"};
...
} optValue;
```
to
```c++
namespace opt_value {
...
namespace matcher {
inline constexpr auto& lookup = u"lookup";
inline constexpr auto& bestFit = u"best fit";
...
} // namespace matcher
} // namespace opt_value
```
And for the unordered_sets, I can replace them with arrays and specify inline constexpr. I believe that should address the concerns?

Any other suggestions are welcomed.

neildhar: You can follow the style used in `PlatformIntlApple.mm`, which is deliberately set up to avoid global constructors, and keep the strings close to the point where they are used (to make it easier to read).

neildhar:
Thank you for putting this together!

I haven't gone into the details yet, but wanted to first share some high level organisational comments. In particular, I'm wary of cluttering the `Platform/Intl` directory with too many files that are specific to this ICU implementation. It looks like there are also dependencies going both ways between the `Platorm/Intl` directory and `Platform/Intl/icu_impl` directory, so it's difficult to reason about how things are organised.

My preference in this instance would be to flatten more of the internals into `PlatformIntlICU.cpp` (in the same way that `PlatformIntlApple.mm` is self-contained), which will reduce the need for things like `Constants.h`. Note that we have many files in Hermes where classes that provide internal functionality live only in the relevant cpp file, so there is no hard rule requiring them to be in separate files.

Where you feel that doesn't make sense, the extra files should all be in icu_impl, so it is clear that they exist only to support the ICU implementation.

lib/Platform/Intl/Constants.h
tmikov: Instead of
```
inline constexpr auto& lookup = u"lookup";
```
can we just do:
```
inline constexpr char16_t lookup[] = u"lookup";
```

robchu05: @tmikov, yes, I will change to declare the type explicitly, rather than `auto`, which also yields char_16t[], per your guidance.

@neildhar, having the Intl public APIs' string constants in a separate Constant.h make them easier to review and check that they match the spec as a group. Another benefit is that the string constants that are semantically the same across the APIs like "localeMatcher" are then also shared across API implementation files. With the above suggested changes, there won't be global constructors involved. Your thoughts? Okay to go with this Constants.h grouping?

neildhar:
@robchu05 Sure, if you feel it makes it easier to keep things split, that's fine too. My comment was primarily about keeping the ICU implementation internals together.

lib/Platform/Intl/Constants.h
robchu05: Right. I am changing the `unordered_set` to array.

I am deleting the `unordered_map` usage in this file. It is for the Intl.DurationFormat API, which is currently a stage 3 TC39 proposal and not part of this PR. (I can put the map as a function local static inside that API implementation to avoid global constructor.)

I plan to push a new commit to update this PR next week to address the feedback so far 😀

robchu05: Updated PR with new commit to make changes as mentioned above.

robchu05:
@tmikov, @neildhar, friendly reminder to review this PR. Appreciate your feedback and help in guiding me through.

lib/Platform/Intl/impl_icu/LocaleResolver.cpp
```diff
@@ -0,0 +1,306 @@
...
+#include "Locale.h"
+#include "LocaleBCP47Object.h"
+#include "OptionHelpers.h"
+#include "hermes/Platform/Unicode/icu.h"
+
+namespace hermes {
+namespace platform_intl {
+namespace impl_icu {
+
+const LocaleResolver::AvailableLocale LocaleResolver::availableLocales_ =
```
neildhar, line 21: This introduces a global constructor. Please wrap it as a function-local static instead.

lib/Platform/Intl/impl_icu/LocaleResolver.h
```diff
@@ -0,0 +1,159 @@
...
+#include <map>
+#include <string>
+#include <string_view>
+#include <unordered_set>
+
+namespace hermes {
+namespace platform_intl {
+namespace impl_icu {
+
+class LocaleResolver {
```
neildhar, line 23: Does this need to be a class? It seems a namespace would suffice.

lib/Platform/Intl/impl_icu/Collator.cpp
```diff
@@ -0,0 +1,592 @@
...
+#include <algorithm>
+#include <optional>
+
+namespace hermes {
+namespace platform_intl {
+namespace impl_icu {
+
+constexpr auto &collationTypeDefault = u"default";
+constexpr auto &collationTypeSearch = u"search";
+constexpr auto &collationTypeStandard = u"standard";
```
neildhar, line 27: We should use the same style as `Constants.h`

lib/Platform/Intl/impl_icu/Collator.cpp
```diff
@@ -0,0 +1,592 @@
...
+    if (extensionType == collationTypeStandard ||
+        extensionType == collationTypeSearch) {
+      return false;
+    }
+    // ICU4C returns alias / legacy type ids for types that were available
+    // before Unicode extension type defined its ids to be
+    // 3 - 8 alphanumeric characters. Map these alias / legacy type ids to
+    // Unicode extension type ids.
+    // https://github.com/unicode-org/cldr/blob/main/common/bcp47/collation.xml
+    static std::unordered_map<std::u16string_view, std::u16string_view>
```
neildhar, line 548: This is pretty costly to construct and lookup given that there are only four entries. A constexpr array of pairs should be good enough for this.

lib/Platform/Intl/impl_icu/Collator.cpp
```diff
@@ -0,0 +1,592 @@
...
+  }
+  // 17. Set opt.[[kf]] to caseFirst.
+  if (caseFirstValueRes->has_value()) {
+    opt.emplace(constants::extension_key::kf, **caseFirstValueRes);
+  }
+
+  // The relevant unicode extensions accepted by Collator as specified here:
+  // https://tc39.github.io/ecma402/#sec-intl-collator-internal-slots
+  // 18. Let relevantExtensionKeys be %Collator%.[[RelevantExtensionKeys]].
+  std::unordered_set<std::u16string> relevantExtensionKeys{
```
neildhar, line 348: Can this be a constexpr array, instead of constructing a new set every time? (similar to the Apple implementation)

lib/Platform/Intl/PlatformIntlICU.cpp
```diff
@@ -591,35 +362,31 @@ vm::ExecutionStatus DateTimeFormatICU::initialize(
   // 17. Let r be ResolveLocale(%DateTimeFormat%.[[AvailableLocales]],
   // requestedLocales, opt, %DateTimeFormat%.[[RelevantExtensionKeys]],
   // localeData).
-  static constexpr std::u16string_view relevantExtensionKeys[] = {
+  static std::unordered_set<std::u16string> relevantExtensionKeys{
```
neildhar, line 365: Could we use a constexpr array here and below, similar to the apple implementation?

neildhar (commented):
I apologise for the delay in getting to this. I've done a pass over, and defer to you on the intricacies of interfacing with ICU, so I mostly have comments on potential inefficiencies.

robchu05:
Hi @neildhar,

Thanks for the review and feedback. I target to address them and post a new revision by end of next week.

lib/Platform/Intl/impl_icu/LocaleResolver.cpp
robchu05: Fixed in 3rd commit.

lib/Platform/Intl/impl_icu/LocaleResolver.h
robchu05: Done in 3rd commit.

lib/Platform/Intl/impl_icu/Collator.cpp
robchu05: Done in 3rd commit.

lib/Platform/Intl/impl_icu/Collator.cpp
robchu05: Done in 3rd commit.

lib/Platform/Intl/impl_icu/Collator.cpp
robchu05: Done in 3rd commit.

lib/Platform/Intl/PlatformIntlICU.cpp
robchu05: Done in 3rd commit.

robchu05:
Hi @neildhar,

I updated the PR to address your feedback. Please have a look 😀

I would also like to get your inputs on the tests.

The CircleCI test failure on test-macos-test262 job is on the additional test cases that I added in test/hermes/intl/. I can work on fixing PlatformIntlApple for a few test cases. For other cases that are not straightforward to fix on Apple and are not critical, we can consider splitting the Hermes intl tests so that one set is run for mac and one set is run for linux.

Another thing to consider is to add a test-linux-test262 job similar to test-macos-test262 on CircleCI. I think that would involve adding a different set of intl test skip list for the test262 test suite because the intl coverage on linux with icu4c is not yet matching Apple platform and some intl test cases in current skip list (for Apple) pass, thus do not need to be skipped, on linux.

Appreciate your thoughts on the above! Thanks!

neildhar:
Hey @robchu05 thanks for updating the PR, I haven't forgotten about this, I've been travelling for the last few weeks and will take a look once I'm back next week.

robchu05:
Thank you for letting me know, @neildhar. Hope you are having a pleasant and safe travel.

robchu05:
Hi @neildhar, hope you had a fun trip! Friendly reminder to review the new revision and also the testing update suggestions in my earlier comment. Thanks.

robchu05:
Hi @neildhar,

Please have a look at the latest changes, starting with the 4th commit "Fix handling of negative test cases and update test configurations". The commits after that are small fix-ups to circle-ci configuration to address run failures. Thank you!

Regarding the windows and test-windows job failures, they are due to compilation errors:

```sh
C:\tmp\hermes\hermes\unittests\Support\StackBoundsTest.cpp(35,14): error C3861: '__builtin_frame_address': identifier not found [C:\tmp\hermes\build\unittests\Support\HermesSupportTests.vcxproj]
C:\tmp\hermes\hermes\unittests\Support\StackBoundsTest.cpp(43,14): error C3861: '__builtin_frame_address': identifier not found [C:\tmp\hermes\build\unittests\Support\HermesSupportTests.vcxproj]
```
Error log line link: https://app.circleci.com/pipelines/github/facebook/hermes/6944/workflows/4df96b47-12a8-42a4-a9f7-5c095d23252f/jobs/71264?invite=true#step-106-84684_132

The file StackBoundsTest.cpp with the compilation error is recently introduced in https://github.com/facebook/hermes/commit/a5b2bbcf33cf7f475b3d7dce96be2371eb5fda0b.

tmikov:
@robchu05 sorry about that, by bad. I just added that test and apparently our CI didn't flag it as broken on Windows.

utils/testsuite/testsuite.py
```diff
@@ -195,7 +197,7 @@ def __str__(self):
 
 
 TIMEOUT_COMPILER = 200
-TIMEOUT_VM = 200
+TIMEOUT_VM = 600
```
neildhar, line 200: This change seems odd, are you seeing tests that need 10 minutes to run? If so, we should skip them and add a TODO to investigate why they are timing out.

.circleci/config.yml
```diff
@@ -729,3 +730,30 @@ jobs:
...
+                libreadline-dev libicu-dev zip python3
+            # Check out test262 at a pinned revision to reduce flakiness
+            git clone https://github.com/tc39/test262
+            cd test262
+            git checkout 62626e083bd506124aac6c799464d76c2c42851b
+      - run:
+          name: Run test262 with Intl
+          command: |
+            cmake -S hermes -B build -DHERMES_ENABLE_INTL=ON
+            cmake --build ./build
```
neildhar, line 755: This is going to be very slow, we should set the parallelism of the build to match the other linux jobs (which pass `-j4`)

utils/testsuite/testsuite.py
robchu05: The top 10 slowest tests from the most recent passed run of test-linux-test262 job are:

```sh
Top 10 slowest tests
test262/test/built-ins/decodeURI/S15.1.3.1_A2.5_T1.js                     388.806
test262/test/built-ins/decodeURIComponent/S15.1.3.2_A2.5_T1.js            387.991
test262/test/built-ins/RegExp/character-class-escape-non-whitespace.js    117.804
test262/test/built-ins/decodeURI/S15.1.3.1_A1.12_T2.js                     52.211
test262/test/built-ins/decodeURIComponent/S15.1.3.2_A1.12_T2.js            51.000
test262/test/built-ins/decodeURI/S15.1.3.1_A1.12_T1.js                     49.002
test262/test/built-ins/decodeURIComponent/S15.1.3.2_A1.12_T3.js            48.802
test262/test/built-ins/decodeURI/S15.1.3.1_A1.12_T3.js                     48.693
test262/test/built-ins/decodeURI/S15.1.3.1_A1.2_T1.js                      48.508
test262/test/built-ins/decodeURI/S15.1.3.1_A1.11_T1.js                     48.404
```

The Linux VM being used to run the job seems to be quite limited in CPU and memory. The longest test ends up taking 388+ seconds.

.circleci/config.yml
robchu05: Ah, good point. Let me edit this.

robchu05: Updated PR to add  `-j 4`.

utils/testsuite/testsuite.py
neildhar: If this is purely an issue for the linux job, let's just configure the build with `-O2` so the test runs faster.

robchu05: Updated PR to add `-DCMAKE_CXX_FLAGS=-O2  -DCMAKE_C_FLAGS=-O2` to cmake configure command for test-linux-test262 CircleCI job. I have not yet reduced the timeout value. Let's see how long the slowest tests take now on CircleCI, then I'll reduce the timeout value.

robchu05: Thanks for the guidance @neildhar. With the compiler optimization flag, the top 10 slowest tests are now:

```
Top 10 slowest tests
test262/test/built-ins/decodeURI/S15.1.3.1_A2.5_T1.js                                      30.295
test262/test/built-ins/decodeURIComponent/S15.1.3.2_A2.5_T1.js                             29.000
test262/test/intl402/String/prototype/localeCompare/returns-same-results-as-Collator.js     7.703
test262/test/built-ins/RegExp/character-class-escape-non-whitespace.js                      6.098
test262/test/intl402/Date/prototype/returns-same-results-as-DateTimeFormat.js               4.307
test262/test/built-ins/decodeURI/S15.1.3.1_A1.11_T1.js                                      3.498
test262/test/built-ins/decodeURIComponent/S15.1.3.2_A1.12_T1.js                             3.407
test262/test/built-ins/decodeURIComponent/S15.1.3.2_A1.11_T2.js                             3.299
test262/test/built-ins/decodeURIComponent/S15.1.3.2_A1.12_T3.js                             3.298
test262/test/built-ins/decodeURIComponent/S15.1.3.2_A1.11_T1.js                             3.290
```

I have updated the PR to revert the time out value change.

robchu05:
Hi @neildhar, anything else needed for merging this pull request?

robchu05:
@neildhar , any edits I need to make to address the Facebook Internal failures?

neildhar:
@robchu05 Looks like they're just formatting issues, I'll fix those internally before committing.

robchu05:
Hi @neildhar, will you be merging this PR some time this week?

PR #1419, opened 2024-06-04
(Cherry Pick) feat: add visionOS support

okwasniewski:
## Summary

This PR is a partial cherry-pick of this commit: dda343b80784fb75c791979c32ce943f8997dd37 from the `main` branch

Discussed offline with @cipolleschi in order to bring visionOS support to RN 0.74

## Test Plan

Not needed

### DIFF
--- a/API/hermes/CMakeLists.txt
+++ b/API/hermes/CMakeLists.txt
@@ -142,6 +142,14 @@ if(APPLE AND HERMES_BUILD_APPLE_FRAMEWORK)
     add_custom_command(TARGET libhermes POST_BUILD
       COMMAND /usr/libexec/PlistBuddy -c "Add :LSMinimumSystemVersion string ${CMAKE_OSX_DEPLOYMENT_TARGET}" $<TARGET_FILE_DIR:libhermes>/Resources/Info.plist
     )
+  elseif(HERMES_APPLE_TARGET_PLATFORM MATCHES "xr")
+    if(CMAKE_VERSION VERSION_LESS 3.28.4)
+      message("VisionOS Simulator requires CMake version >= 3.28.4")
+    endif()
+
+    add_custom_command(TARGET libhermes POST_BUILD
+      COMMAND /usr/libexec/PlistBuddy -c "Add :MinimumOSVersion string ${CMAKE_OSX_DEPLOYMENT_TARGET}" $<TARGET_FILE_DIR:libhermes>/Info.plist
+    )
   endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


tmikov:
Is this the only change from that commit? The original change also changes CMakeLists.txt, a couple of scripts, etc?

okwasniewski:
Thank you @tmikov I missed the CMakeLists.txt file. @cipolleschi mentioned the build scripts are not needed because they are separate in the React Native repo, so these should be the minimal changes needed to get this working

tmikov:
@okwasniewski I believe the the CMake files and the scripts are used by RN. The only thing that is ignored is the pod file.

cipolleschi:
The things that we use for React Native are in [this folder](https://github.com/facebook/react-native/tree/main/packages/react-native/sdks/hermes-engine).
To summarize, the build logic files we replace are:
- Cocoapods stuff (podspec and helpers)
- Some shell scripts (build-apple-framework, for example).

For everything  else, we rely on the Hermes repo. So yeah, if there are changes to other files other than the one above, we need to pick them as well.

Looking at the original commit, only `Logging.cpp` and the `CMakeList` needs to be picked. 
We can ignore the test folder.

@okwasniewski do you happen to know if we need some similar changes for tvOS? If yes, can we add another change on top of this? cc. @douglowder

One of the proposition of the [React Native Framework RFC](https://github.com/react-native-community/discussions-and-proposals/blob/main/proposals/0759-react-native-frameworks.md) is not to redistribute binaries: right now AVP and tvOS need to redistribute Hermes because you guys needs to rebuild it with the right slices. But we have an opportunity to unify all of them in Reac Native.

okwasniewski:
Thanks for clarifying @cipolleschi.

Regarding tvOS, I think similar changes should be made as tvOS is also sharing lots of code with iOS. I would guess mostly to `CMakeList.txt` as the Logging.cpp import was strictly visionOS quirk

douglowder:
@okwasniewski @cipolleschi the changes I made for tvOS are in https://github.com/react-native-tvos/hermes/tree/tvos-v0.74.0

cipolleschi:
Cool, thanks @douglowder!
I think we have a couple of commits that we need to consider in order to make this work.
Are those changes already in `main`?

douglowder:
@cipolleschi I merged my changes to current main: https://github.com/react-native-tvos/hermes/commit/be3f05c1a5b2e82b2f8aad36e1f3396f517ce67a

PR #1430, opened 2024-06-05
Expose a way to create a JSI HernesRuntime instance from an existing Hermes VM instance

rFlex:
## Summary

I understand this change might be a bit contentious, but please hear me out.

Our Hermes integration does not use the JSI API, we integrate with the Hermes Runtime C++ API directly. There are a few reasons for this:
- We already have an API that is similar to JSI for which we have 4 different JS engines integrated with it. If we were integrating Hermes through JSI we would be implementing an abstraction on top of an abstraction, which adds performance and memory overhead. We want our Hermes integration to be as fast as it can be, and for that we need our bridge layer to have minimal overhead so that it can compete on a performance level against our other JS engine integrations that are also based on lower level JS engine primitives.
- There are APIs we need to integrate that can't be implemented with JSI.
- We don't have C++ exceptions enabled, and the JSI Hermes integration makes heavy use of them and forces its consumers to handle errors through exceptions.

The Hermes Debugger implementation is however based on the JSI API, so in order for us to leverage the Hermes Debugger, this change adds a small shim that can create a JSI Runtime instance from an existing Hermes Runtime C++, so that we can pass it to the Debugger. This makes it possible for us to keep the rest of our integration intact, and just call `hermes::adoptHermesRuntime()` on debug builds to create a debugger instance. Through that system, the linker is able to strip out all the JSI code for us on release builds.

### DIFF
--- a/API/hermes/hermes.h
+++ b/API/hermes/hermes.h
@@ -25,6 +25,7 @@ struct HermesTestHelper;
 namespace hermes {
 namespace vm {
 class GCExecTrace;
+class Runtime;
 } // namespace vm
 } // namespace hermes
 
@@ -238,6 +239,9 @@ HERMES_EXPORT ::hermes::vm::RuntimeConfig hardenedHermesRuntimeConfig();
 HERMES_EXPORT std::unique_ptr<HermesRuntime> makeHermesRuntime(
     const ::hermes::vm::RuntimeConfig &runtimeConfig =
         ::hermes::vm::RuntimeConfig());
+HERMES_EXPORT std::unique_ptr<HermesRuntime> adoptHermesRuntime(
+    const std::shared_ptr<::hermes::vm::Runtime> &runtime,
+    const ::hermes::vm::RuntimeConfig &runtimeConfig);
 HERMES_EXPORT std::unique_ptr<jsi::ThreadSafeRuntime>
 makeThreadSafeHermesRuntime(
     const ::hermes::vm::RuntimeConfig &runtimeConfig =
--- a/API/hermes/DebuggerAPI.h
+++ b/API/hermes/DebuggerAPI.h
@@ -271,6 +271,9 @@ class HERMES_EXPORT Debugger {
  private:
   friend std::unique_ptr<HermesRuntime> hermes::makeHermesRuntime(
       const ::hermes::vm::RuntimeConfig &);
+  friend std::unique_ptr<HermesRuntime> hermes::adoptHermesRuntime(
+      const std::shared_ptr<::hermes::vm::Runtime> &,
+      const ::hermes::vm::RuntimeConfig &);
   friend std::unique_ptr<jsi::ThreadSafeRuntime>
   hermes::makeThreadSafeHermesRuntime(const ::hermes::vm::RuntimeConfig &);
   friend ProgramState;
--- a/API/hermes/hermes.cpp
+++ b/API/hermes/hermes.cpp
@@ -150,9 +150,16 @@ class HermesRuntimeImpl final : public HermesRuntime,
   static constexpr uint32_t kSentinelNativeValue = 0x6ef71fe1;
 
   HermesRuntimeImpl(const vm::RuntimeConfig &runtimeConfig)
+      : HermesRuntimeImpl(
+            runtimeConfig,
+            ::hermes::vm::Runtime::create(runtimeConfig)) {}
+
+  HermesRuntimeImpl(
+      const vm::RuntimeConfig &runtimeConfig,
+      const std::shared_ptr<::hermes::vm::Runtime> &rt)
       : hermesValues_(runtimeConfig.getGCConfig().getOccupancyTarget()),
         weakHermesValues_(runtimeConfig.getGCConfig().getOccupancyTarget()),
-        rt_(::hermes::vm::Runtime::create(runtimeConfig)),
+        rt_(rt),
         runtime_(*rt_),
         vmExperimentFlags_(runtimeConfig.getVMExperimentFlags()) {
 #ifdef HERMES_ENABLE_DEBUGGER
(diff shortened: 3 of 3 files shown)


neildhar:
Hey @rFlex, thanks for putting this up and sharing that context. As you probably know, the Hermes internal APIs are unstable and unsafe, so we deliberately do not expose them. That said, I understand the issues you're describing, and that this solves a legitimate problem for you.

In terms of this PR, I think it would be preferable to expose something like `vm::Runtime *getVMRuntimeUnsafe()` on `HermesRuntime`, rather than allowing you to construct a `HermesRuntime` from an existing `vm::Runtime`. Consolidating the ownership of the runtime in `HermesRuntime` makes it easier to manage the lifetimes correctly here. In particular, we want to avoid a situation where `HostFunction`s and `HostObject`s retaining `jsi::Value`s remain alive through the `vm::Runtime` after the corresponding `HermesRuntime` has been freed. Given that the lifetimes of these are managed by the GC, the safest thing to do is to ensure that the `vm::Runtime` is itself owned by the `HermesRuntime`, and therefore cannot outlive it.

I would like to hear more on your concerns though. Getting feedback from engaged Hermes integrators is extremely useful for us.

1. Have you benchmarked the relative performance of JSI against the internal APIs? It would be interesting to hear if there are scenarios where you have found the JSI overhead unacceptable, and if there are specific operations you found responsible. It would also be useful to know when you benchmarked (since Hermes' JSI performance has improved considerably in the last two years).
2. What things have you tried implementing that weren't possible in JSI? We're open to improving and extending JSI to support more use cases.
3. This is understandable. We are working on adding a stable C-based API to Hermes that is designed to serve as an intermediate layer between Hermes and JSI. The current version is in `API/hermes_abi`. This would address your concerns about exceptions, but is intended to be close to JSI in terms of performance and functionality.

rFlex:
Hey @neildhar ,

The reason we had it as `adoptHermesRuntime()` instead of `makeHermesRuntime()` + `getVMRuntimeUnsafe()` is that it allowed us to create the Hermes runtime the same way on prod + debug build, it's just that on debug build we also used `adoptHermesRuntime()` to provide an instance to the debugger. We have libcxx rtti disabled for app size reasons, which JSI requires, so we kept rtti enabled for debug builds just for being able to use JSI + Hermes Debugger, and we keep rtti disabled for prod build. That means we don't have access to the JSI classes on prod builds for us today. We could however use `makeHermesRuntime()` + `getVMRuntimeUnsafe()` on debug builds and create the Hermes C++ runtime ourselves on prod builds, not the biggest deal in the world. Happy to make the change.

Answers to your questions:
1. There were no benchmarks, it is possible that the performance would have been fine. It's just that the set of potential problems we would face would make it costly to attempt to integrate with JSI. If the other two points were not a problem, we would have likely gone that route as it'd have been cheaper to do it that way, and move to the lower level API only if we found performance problems. Since there was some risks involved, we decided to integrate with the lower level API directly, which eliminates the possibility of the performance overhead of the abstraction being a problem entirely. The JSI API is not a zero overhead abstraction, and our JS API is not as well. This was a preemptive optimization that only made sense because it was cheaper to go with the optimized route (use the lower level API) rather than the unoptimized route (use the JSI API).
2. Some of the things that I can remember, and let me know if I missed that there are ways to do this with JSI:
  - Compiling JS into JS bytecode and get the output as bytes, at runtime as well as ahead of time without using the `hermesc` binary. We use this for ahead of time bytecode compilation, and we also have a bytecode cache at runtime where we can compile js into bytecode in the background and store it on disk. This reduces the app size cost of modules as Hermes bytecode is quite large. The proposed way by Hermes on how to compile and deal with bytecode limits our options here unless I'm missing something.
  - Enabling Hermes's rejection tracker system, which we have to resolve the `HermesInternal` object and call `enablePromiseRejectionTracker` on it.
  - Manipulating strings efficiently (creating a string from UTF16 without converting it first to UTF8, avoid C++ std::string conversions)
I am likely missing some things, the Hermes integration was done in Q4 last year.
3. Yes I did notice that the C API was in progress. It's likely we would have used it if was available at the time of the integration.

I think we are unlikely to be representative of a typical Hermes user.

neildhar:
Thanks for the detailed response! Regarding the the gaps you identified in JSI:

1. Compiling JS to bytecode is indeed not available in JSI. But you can compile with our internal APIs, or using the public (but more restricted) `compileJS` API, and feed the result into JSI through `evaluateJavaScript`. That is, using the internal compiler APIs is orthogonal to using JSI.
2. Accessing the properties on `HermesInternal` can be done in JavaScript, or through JSI by getting the `HermesInternal` property off the global object. What issues did you run into doing this?
3. String conversion is definitely a gap. We intend to support 16 bit characters in JSI, which will incur less overhead, but we don't intend to support direct access to the bytes in the Hermes heap (which would be very unsafe).

rFlex:
@neildhar great callouts, we could have used the compiler APIs for bytecode compilation independently of JSI. I thought that `HermesInternal` was a private symbol and not available to user code, but reading the code that seems completely wrong and I don't remember how I got that assumption.

rFlex:
@neildhar changed to expose underlying instance of from the JSI runtime instance

API/hermes/hermes.h
```diff
@@ -214,6 +215,11 @@ class HERMES_EXPORT HermesRuntime : public jsi::Runtime {
       const std::shared_ptr<const jsi::Buffer> &sourceMapBuf,
       const std::string &sourceURL);
 
+  /// Returns the underlying low level Hermes VM runtime instance.
+  /// This should be avoided as the lower level APIs are unsafe
+  /// and they can change without notice.
```
neildhar, line 220: Strengthening the wording here a little
```suggestion
  /// This function is considered unsafe and unstable.
  /// Direct use of a vm::Runtime should be avoided as the lower level APIs are unsafe
  /// and they can change without notice.
```

neildhar (commented):
Looks good overall, just a minor comment on the wording. Could you also update the title and description? Then I'll pull it in to merge.

rFlex:
hey @neildhar , anything else I should update? could we get this merged?

neildhar:
@rFlex Just the title and the summary need to be updated, everything else looks good

rFlex:
@neildhar updated!

rFlex:
@neildhar can we get this merged? thanks!

mrousavy:
> We already have an API that is similar to JSI for which we have 4 different JS engines integrated with it

Off-topic, but I'm curious what 4 engines you have abstracted, and why? 😄

PR #1442, opened 2024-06-20
Make fallthroughs in regcomp.c explicit

r-barnes:
## Summary

Explicit fallthroughs are safer and allow folks to enable `-Wimplicit-fallthrough` for their code to avoid unintentional fallthroughs.

## Test Plan

If it builds, it works.


dannysu:
Fixed internally

PR #1445, opened 2024-06-20
Make fallthroughs explicit in dtoa.c

r-barnes:


dannysu:
Looks like we have diffs internally to fix this.

Issue #1270, opened 2024-01-23
Array.protoype.indexOf() Performance Regression


tctacm:
## Bug Description

The `Array.prototype.indexOf()` functions seems to be significantly slower on Hermes when compared to JSC. I first noticed the issue in React Native 0.64 inside the JSTimers file: https://github.com/facebook/react-native/blob/0645c38014e8310d8e387dabc860c8c5878beb6e/Libraries/Core/Timers/JSTimers.js#L62 , however, I have reproduced a case using the Hermes CLI to verify my findings.

- [X] I have run `gradle clean` and confirmed this bug does not occur with JSC
- [ ] The issue is reproducible with the latest version of React Native.

Hermes git revision (if applicable): release-v0.12? I'm using Hermes CLI with version 0.12.0
React Native version: N/A
OS: macOS Sonoma 14.2.1
Platform (most likely one of arm64-v8a, armeabi-v7a, x86, x86_64): x86_64

## Steps To Reproduce

The following code running on Hermes will take significantly longer:

`index_of_test.js`:
```
const ARR_SIZE = 10000;

// Make an array of ARR_SIZE increasing non-negative integers
// [0, 1, 2, 3, 4, 5, 6, 7, 8, ...]
var test_arr = [];
for (var i = 0; i < ARR_SIZE; i++) {
  test_arr.push(i);
}

// Call indexOf 1000 times with the last element in the array
var start = Date.now();
for (var i = 0; i < 1000; i++) {
  test_arr.indexOf(ARR_SIZE - 1);
}
var end = Date.now();
print('Array.indexOf last element time: ' + (end - start));

// Call indexOf 1000 times with the first element in the array
var start = Date.now();
for (var i = 0; i < 1000; i++) {
  test_arr.indexOf(0);
}
var end = Date.now();
print('Array.indexOf first element time: ' + (end - start));
```

1. Run with hermes cli: `hermes ./index_of_test.js`
```
Array.indexOf last element time: 253
Array.indexOf first element time: 0
```
2. Run with JSC cli: `jsc --useJIT=false ./index_of_test.js`
```
Array.indexOf last element time: 3
Array.indexOf first element time: 0
```
My example seems to indicate that Hermes is much slower at iterating through and comparing Array values when compared to JSC. 

[...]

PR #1447, opened 2024-06-24
Add fast path for Array.prototype.indexOf/lastIndexOf

lavenzg:
Summary:
If `this` is truly an Array object, go fast path, otherwise, falls to
slow path. In fast path, use specialized loop given the type of search
value and reverse mode. For checking Number, add a new function that
returns double value if the HV is a number and returns NaN if it's not.
This avoids repeated tag checking in `isNumber()` and `getNumber()`
for HV32 mode.

On the given benchmark it's ~12X faster on both HV32 and HV64 mode.

Differential Revision: D58201521


miallo:
Nice!
For the reference: this fixes https://github.com/facebook/hermes/issues/1270
Related to https://github.com/facebook/react-native/issues/41631 and https://github.com/facebook/react-native/issues/36296

Issue #1009, opened 2023-05-23
Missing property: `Promise.prototype[Symbol.toStringTag]`


gibson042:
## Bug Description

https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise.prototype-@@tostringtag requires `Promise.prototype` to have a Symbol.toStringTag property with value "Promise", but it is absent in Hermes.

- [x] I have ~~run `gradle clean` and~~ confirmed this bug does not occur with JSC

Hermes version: 0.11.0

## Steps To Reproduce

code example:
```js
Promise.prototype[Symbol.toStringTag]
```

## The Expected Behavior
"Promise"

tmikov:
Hi, thank you for reporting this. 

As described [here](https://github.com/facebook/hermes/blob/e2f48291e4635b230c1267a77c05ea8efe216725/doc/Features.md#miscellaneous-incompatibilities), Hermes implements Promise by using the Promise polyfill https://github.com/then/promise, which unfortunately isn't fully compliant. This was done to ensure compatibility with React Native, which was already using it before Hermes.

Of course this situation isn't ideal, but none of the short term fix options are particularly attractive:
- Develop and maintain a local patch on top of the polyfill.
- Submit a PR to improve the polyfill.
- Find and use a better polyfill.
- Implement a native Promise implementation from first principles.

We also need to confirm that RN does not depend on any non-spec extensions in the existing polyfill. I will reach out to the React Native team to clarify the requirements.

ljharb:
https://npmjs.com/es6-shim is fully compliant, if you wanted to consider an alternative.

tmikov:
@ljharb It is certainly worth looking into. Two quick questions if you don't mind:
- AFAICT, the Promise shim isn't separated. Do you think it would be feasible for us to split it out?
- Do you have an intuition about the performance? My experience with promise is from 9 years ago, and at that time, IIRC, polyfills had wildly different performance. I think Bluebird was the fastest at the time.

ljharb:
You're right that it's not separated; I would be happy to split it out if that helps the hermes team.

Bluebird remains the fastest by violating the spec in a number of small ways; I'm not sure about the performance at this point tbh, but I'm also open to performance improvements as long as they don't violate the spec.

tmikov:
[...]

PR #1452, opened 2024-06-28
Implement ES2024 `Promise.withResolvers`

robik:
## Summary

This PR implements ES2024 [`Promise.withResolvers`](https://tc39.es/ecma262/#sec-promise.withResolvers) function.

## Test Plan

Added new test

### DIFF
--- a/lib/InternalBytecode/01-Promise.js
+++ b/lib/InternalBytecode/01-Promise.js
@@ -365,6 +365,17 @@
     return error;
   }
 
+  function withResolvers() {
+    var resolve, reject;
+    var promise = new this(function(doResolve, doReject) {
+      resolve = doResolve;
+      reject = doReject;
+    });
+    return { resolve, reject, promise };
+  }
+
+  core.withResolvers = withResolvers;
+
   core.any = function promiseAny(values) {
     return new core(function(resolve, reject) {
       var promises = iterableToArray(values);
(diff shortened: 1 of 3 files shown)


tmikov:
Hmm, we have to figure out how to deal with this. The problem is that 01-Promise.js is automatically generated by https://github.com/facebook/hermes/blob/main/utils/gen-promise-internal-bc.sh from . 

However, at the same time, we can't be stuck forever with an outdated Promise implementation. The discussion in https://github.com/facebook/hermes/issues/1009 is also relevant here. 

So, the question is, should we:
1. Essentially fork https://github.com/then/promise, which hasn't been updated in some time and start updating it?
2. Extract the promise from https://www.npmjs.com/package/es6-shim.

I think that at least in the short term 1) is the only practical option. Then the question becomes, how to actually modify it? Should we just tweak the built result? (I am on vacation, so I may be slow to respond, but I am curious what others think too)

neildhar:
Forking it does seem like the most straightforward. I suppose the best way to do it comes down to how difficult it is to build the promise implementation. If we could easily copy the original source to our `external` directory, and point `utils/promise` to it, then we can freely modify the original input while maintaining its structure. That has some value if we find ourselves digging into the blame for the implementation in the future, or if there is some movement in the repo upstream.

ljharb:
@tmikov still happy to extract the Promise implementation to a separate package, if that'd be helpful - i could include finally and withResolvers in it, since those are already authored as standalone packages. (i can also help you do an optimal build to eliminate parts of the graph that aren't needed in hermes)

retyui:
@robik I think pr can be closed https://x.com/tell_me_mur/status/2066833376674283715?s=20

PR #1454, opened 2024-06-28
Implement ES2024 `Map.groupBy`

robik:
## Summary

This PR implements EcmaScript 2024 [`Map.groupBy`](https://tc39.es/ecma262/#sec-map.groupby)

## Test Plan

Added tests.

```
>> Map.groupBy([1, 2, 3, 4, 5, 6, 7], function(key) { return key % 2; })
Map { "1" => [ 1, 3, 5, 7, [length]: 4 ], "0" => [ 2, 4, 6, [length]: 3 ] }
```


lib/VM/JSLib/Map.cpp
```diff
@@ -345,5 +354,116 @@ mapIteratorPrototypeNext(void *, Runtime &runtime, NativeArgs args) {
...
+    // compute key for current element
+    auto keyRes = Callable::executeCall2(grouperFunc, runtime, callbackThis, valueHandle.getHermesValue(), HermesValue::encodeTrustedNumberValue(k));
+    if (LLVM_UNLIKELY(keyRes == ExecutionStatus::EXCEPTION)) {
+      return iteratorCloseAndRethrow(runtime, iteratorRecord.iterator);
+    }
+
+    groupKey = runtime.makeHandle(std::move(*keyRes));
+
+    // make it a property key
+    auto propertyKeyRes = toPropertyKey(runtime, groupKey);
```
avp, line 430: Looks like we're supposed to call `GroupBy` with "COLLECTION" mode, so we shouldn't be converting to PropertyKey here.

test/hermes/map.js
```diff
@@ -363,3 +363,49 @@ testClear(o1, o2, o3);
...
+  {name: 'John', age: 22},
+];
+
+var map = Map.groupBy(team, function(p) { return p.age < 18 ? 'underage' : 'adult'; });
+print(map.size);
+//CHECK-NEXT: 2
+print(JSON.stringify(map.get('underage')))
+//CHECK-NEXT: [{"name":"Peter","age":15}]
+print(JSON.stringify(map.get('adult')))
+//CHECK-NEXT: [{"name":"Mike","age":20},{"name":"John","age":22}]
```
avp, line 411: We should also have a couple tests for non-string/int keys.

avp (changes requested):
Thanks for adding this!

Issue #1138, opened 2023-10-01
Android: Intl.NumberFormat style percent not working properly


Bekaxp:
## Bug Description
When trying to format a number as a percent with the following code:

```
Intl.NumberFormat('en', {style: 'percent'}).format(0.1091);
```

the output on Android is incorrect. Instead of `%` sign a `+` sign is displayed. This works correctly for the web and iOS. Please have a look at the repro bellow.

React Native version (if any): 0.72.4 (Expo SDK 49)
OS version (if any): Android API level 34

## Steps To Reproduce
The following code:

```
import { Text, SafeAreaView } from 'react-native';

export default function App() {
  return (
    <SafeAreaView style={{flex: 1, justifyContent: 'center', alignContent: 'center'}}>
      <Text style={{textAlign: 'center', fontSize: 24}}>
        {Intl.NumberFormat('en', {
          style: 'percent',
          minimumFractionDigits: 2,
          signDisplay: 'exceptZero'
        }).format(0.1091)}
      </Text>
    </SafeAreaView>
  );
}
```

will produce these results:

**Android**
<img width="322" alt="Screenshot 2023-10-01 at 11 11 22" src="https://github.com/facebook/hermes/assets/13329079/168c6c42-e332-44cc-bf16-31f2d022c6ee">

**iOS**
<img width="325" alt="Screenshot 2023-10-01 at 11 11 45" src="https://github.com/facebook/hermes/assets/13329079/e24ed91b-9098-4478-8317-b65ec9a0adb4">

As you can see, on Android the`%` sign is replaced by `+`, which is incorrect.

### Snack example
[SNACK LINK](https://snack.expo.dev/PkoMc6ByS)

## The Expected Behavior
Android should display the `%` sign correctly when formatting a number with `Intl.NumberFormat` and `style: 'percent'`.

tamlyn:
Related to – or possibly duplicate of – https://github.com/facebook/hermes/issues/789

Pre-Hermes we were using the [Intl polyfills](https://formatjs.io/docs/polyfills/) which worked nicely. But now it seems the polyfills are deferring to the native broken implementation. Any way to force the polyfill to be used?

In the meantime it's best not to use `signDisplay`. Luckily it's quite easy to implement the behaviour yourself.

shubhamguptadream11:
I had fixed this issue here: https://github.com/facebook/hermes/pull/1483. Please try it once

Bekaxp:
[...]

Issue #789, opened 2022-07-29
Intl.NumberFormat on Android with style: 'currency' and signDisplay: 'always' truncates currency sign for positive values


stefan-schweiger:
## Bug Description

When you use Intl.NumberFormat number with `style: 'currency'` and `signDisplay: 'always'` it works well enough for negative numbers, but as soon as you try it with a positive number it just removes the currency sign. See the code below.

This problem only occures on Android

This is not relevant as Intl doesn't work with JSC:
- [ ] I have run `gradle clean` and confirmed this bug does not occur with JSC

Hermes version: hermes-engine@0.11.0
React Native version: 0.69.3
Platform: arm64-v8a

## Steps To Reproduce

```
const App = () => {
  return (
    <View>
      <Text>
        {new Intl.NumberFormat('en-US', {style: 'currency', currency: 'EUR', signDisplay: 'always'}).format(100.32)}
        {'\n'}
        {new Intl.NumberFormat('en-US', {style: 'currency', currency: 'EUR', signDisplay: 'always'}).format(-100.32)}
        {'\n'}
        {new Intl.NumberFormat('en-US', {style: 'currency', currency: 'EUR', signDisplay: 'auto'}).format(100.32)}
        {'\n'}
        {new Intl.NumberFormat('en-US', {style: 'currency', currency: 'EUR', signDisplay: 'auto'}).format(-100.32)}
        {'\n'}

        {!!global.HermesInternal ? 'Hermes' : 'JSC'}
      </Text>
    </View>
  );
};
export default App;
```

Produces the following output on Android:
```
+100.32
-€100.32
€100.32
-€100.32
```

For iOS and Web it produces the correct output:
```
+€100.32
-€100.32
€100.32
-€100.32
```

## The Expected Behavior

Intl.NumberFormat should produce the same output on each platform.

dwatring:
Similarly, for android this configuration is also not consistent:
```
const App = () => {
  return (
    <View>
      <Text>
        {new Intl.NumberFormat('en-US', {style: 'percent', signDisplay: 'always'}).format(.32)}
        {'\n'}
        {new Intl.NumberFormat('en-US', {style: 'percent', signDisplay: 'always'}).format(-.32)}
        {'\n'}
        {new Intl.NumberFormat('en-US', {style: 'percent', signDisplay: 'auto'}).format(.32)}
        {'\n'}
[...]

PR #1483, opened 2024-08-17
fix: signDisplay plus sign behaviour fixed for NumberFormat

shubhamguptadream11:
## Summary

Fixes following issue: 
- https://github.com/facebook/hermes/issues/1466
- https://github.com/facebook/hermes/issues/1138
- https://github.com/facebook/hermes/issues/789

**Why this change is made?**

There are few cases where signDisplay is not being handled correctly.
Examples:
```
1.
new Intl.NumberFormat('de-DE', {
 style: 'currency',
 currency: 'EUR',
 signDisplay: 'exceptZero',
}).format(8537.71)

Output: +8,537,71+
Expected: +8.537,71 €

2.
new Intl.NumberFormat('ja-JP', {
 style: 'currency', 
 currency: 'JPY', 
 signDisplay: 'exceptZero' 
}).format(123456.789)

Output: +123,457
Expected: +￥123,457
```

**ChangeLog**

This PR updates the implementation and testing of the signDisplay functionality in the DecimalFormat class within the Hermes engine, specifically for Android API level 31 and above.

**Key Changes:**

Implementation:
- Integrated the **setSignAlwaysShown** method of DecimalFormat for API level 31 and above to control the display of the sign (+ or -) based on the signDisplay option. [For more detail about setSignAlwaysShown check [here](https://developer.android.com/reference/android/icu/text/DecimalFormat#setSignAlwaysShown(boolean))]
- For API levels below 31, maintained the existing logic for handling sign display, ensuring backward compatibility.

## Test Plan

- Added a comprehensive set of test cases in **HermesIntlAndroidTest.java** to validate the behaviour of the signDisplay functionality, specifically for API level 31 and above.
- Test cases cover scenarios for signDisplay: `NEVER | ALWAYS | EXCEPTZERO`

### DIFF
--- a/android/intltest/java/com/facebook/hermes/test/HermesIntlAndroidTest.java
+++ b/android/intltest/java/com/facebook/hermes/test/HermesIntlAndroidTest.java
@@ -95,7 +95,7 @@ public class HermesIntlAndroidTest {
             "var result = nf.format(8537.71);");
 
         String result = rt.getGlobalStringProperty("result");
-        assertThat(result).isEqualTo("+8.537,71 €");
+        assertThat(result).isEqualTo("+8.537,71 €");
       }
     }
   }
@@ -109,7 +109,7 @@ public class HermesIntlAndroidTest {
             "var result = nf.format(8537.71);");
 
         String result = rt.getGlobalStringProperty("result");
-        assertThat(result).isEqualTo("8.537,71 €");
+        assertThat(result).isEqualTo("8.537,71 €");
       }
     }
   }
@@ -123,7 +123,7 @@ public class HermesIntlAndroidTest {
             "var result = nf.format(8537.71);");
 
         String result = rt.getGlobalStringProperty("result");
-        assertThat(result).isEqualTo("+8.537,71 €");
+        assertThat(result).isEqualTo("+8.537,71 €");
       }
     }
   }
--- a/android/intltest/java/com/facebook/hermes/test/HermesIntlAndroidTest.java
+++ b/android/intltest/java/com/facebook/hermes/test/HermesIntlAndroidTest.java
@@ -9,19 +9,24 @@ package com.facebook.hermes.test;
 
 import static org.assertj.core.api.Java6Assertions.assertThat;
 
+import android.content.Context;
 import android.content.res.AssetManager;
-import android.test.InstrumentationTestCase;
+import androidx.test.platform.app.InstrumentationRegistry;
+import androidx.test.ext.junit.runners.AndroidJUnit4;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.stream.Collectors;
 import org.junit.Test;
+import org.junit.runner.RunWith;
 
-public class HermesIntlAndroidTest extends InstrumentationTestCase {
+@RunWith(AndroidJUnit4.class)
+public class HermesIntlAndroidTest {
+
   @Test
   public void testIntlFromAsset() throws IOException {
-    AssetManager assets = getInstrumentation().getContext().getAssets();
+    AssetManager assets = InstrumentationRegistry.getInstrumentation().getTargetContext().getAssets();
     InputStream is = assets.open("intl.js");
     String script =
         new BufferedReader(new InputStreamReader(is)).lines().collect(Collectors.joining("\n"));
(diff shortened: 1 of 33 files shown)


shubhamguptadream11:
@tmikov Whenever you have time, can you please check this PR? It solves the `signDisplay` issue reported twice on the platform.

neildhar:
Hey @shubhamguptadream11, thank you for looking into this and submitting this PR!

Could you describe the exact change here in more detail? It looks like there is more going on with this change, with regards to whether the positive/negative strings are checked and whether things are added to the suffix or prefix.

Alternatively, it looks like Android API level 31 adds [native support](https://developer.android.com/reference/android/icu/text/DecimalFormat#setSignAlwaysShown(boolean)) for this functionality. I think it would be preferable to add an API version check and use the newer API when possible, which should be available on most devices.

neildhar:
Looks great!

It looks like CI is failing, likely because the SDK version specified in our `build.gradle` is too low. Could you bump that, and add a test for the desired behaviour in `HermesIntlAndroidTest.java`?

shubhamguptadream11:
Hi @neildhar,

I attempted to upgrade the `compileSdkVersion` to 31, but I encountered several issues due to deprecated APIs. For example, **InstrumentationTestCase** is deprecated as of API level 24 and needs to be replaced with **InstrumentationRegistry**. Given the scope of these changes, it would be more appropriate to handle this upgrade in a separate PR.

Could someone from the Hermes team take on the task of upgrading the compile/target version to the latest supported? or help me in upgrading that. Once that's done, I'll be able to raise my PR.

Thanks!

shubhamguptadream11:
@neildhar Finally all the check are passed now.
Please review it once.

neildhar:
@shubhamguptadream11, thanks for bumping the SDK version and fixing up the deprecated APIs. Everything looks good, please add a test and update the summary, and I'll merge this.

shubhamguptadream11:
@neildhar Added test cases and updated PR summary as well. Please check and let me know if anything else needed from my end.
Thanks again!

shubhamguptadream11:
Hi @neildhar 

Just wanted to follow up on this PR. I’ve added the test cases and updated the summary as discussed. Could you please review it when you get a chance? Let me know if there’s anything else you need from my side.

Thanks again for your time!

android/intltest/java/com/facebook/hermes/test/HermesIntlAndroidTest.java
```diff
@@ -80,4 +85,46 @@ public void testDateTimeFormatCaseInsensitivity() {
...
+      }
+    }
+  }
+
+  @Test
+  public void testSignDisplayExceptZeroForApiLevelAbove31() {
+    if (Build.VERSION.SDK_INT >= 31) {
+      try (JSRuntime rt = JSRuntime.makeHermesRuntime()) {
+        rt.evaluateJavaScript(
+            "var nf = new Intl.NumberFormat('de-DE', { exceptZero: 'never', currency: 'EUR' });\n" +
```
neildhar, line 122: This test seems odd, is it correct? I'm not aware of `exceptZero` being an option to `NumberFormat` (as opposed to `signDisplay: "exceptZero"`), and I don't believe the currency is printed unless you set `style: "currency"`.

If this is incorrect, we should understand why this test didn't fail.

shubhamguptadream11: I've corrected the test case to properly reflect the intended behaviour by using signDisplay: 'exceptZero' and including the style: 'currency' option.

shubhamguptadream11:
Hi @neildhar ,

Could we please check and review this PR when you have a moment?

neildhar:
@shubhamguptadream11 Could you clarify if the test is actually running or if you ran it locally? It seems odd that the previous version of the test was working.

shubhamguptadream11:
@neildhar 
I haven't found documentation specifically for running Android test cases like HermesIntlAndroidTest.java. I was under the assumption that these tests are automatically run on the CI after raising a PR. It's surprising to me as well that the previous tests passed without issues. Could you provide any guidance on how to properly run these tests locally or confirm if they are indeed executed on the CI?

shubhamguptadream11:
@neildhar Please check the PR once. I had shared few screenshots above about test cases results.

shubhamguptadream11:
Hi @neildhar 
I hope you're doing well. I wanted to follow up on the status of PR. If there's anything you'd like me to adjust, just let me know.

Thank you for your time and support!

shubhamguptadream11:
Hi @neildhar, I hope you are doing well. Can you please take a moment to review this PR.
Thanks!

PR #1489, opened 2024-08-21
Makes HermesRuntimeImpl::isArray proxy-compatible

dorentus:
When using hermes engine, [`dynamicFromValue` in React Native](https://github.com/facebook/react-native/blob/main/packages/react-native/ReactCommon/jsi/jsi/JSIDynamic.cpp#L166) calls `isArray` here to check if the input object is an array.

Consider this piece of code:
```js
const list = ['a', 'b', 'c', 'd'];
const proxyed_list = new Proxy(list, {});
```

The `proxyed_list` behaves exactly the same as list in Javascript. But when been passed to native and converted to `folly::dynamic` by `dynamicFromValue`, `proxyed_list` will be converted to `{"0": "a", "1": "b", "2": "c", "3": "d"}` but not the expected `["a", "b", "c", "d"]`.

This patch implements similar routines in commit [26840ed441d614a07809c612521074ca9544086c](https://github.com/facebook/hermes/commit/26840ed441d614a07809c612521074ca9544086c#diff-059b8f2fcb0235b35582efc591e065bd8caa898c3ea2ef2a3e453bff97584e4dR1575) and makes `isArray` proxy-compatible.

### DIFF
--- a/API/hermes/hermes.cpp
+++ b/API/hermes/hermes.cpp
@@ -2030,9 +2030,7 @@ void HermesRuntimeImpl::setPropertyValue(
 
 bool HermesRuntimeImpl::isArray(const jsi::Object &obj) const {
   auto res = llvh::isArray(runtime_, llvh::dyn_vmcast<vm::JSObject>(phv(obj)));
-  if (LLVM_UNLIKELY(res == llvh::ExecutionStatus::EXCEPTION)) {
-    return false;
-  }
+  const_cast<HermesRuntimeImpl *>(this)->checkStatus(res.getStatus());
   return *res;
 }
 
--- a/API/hermes/hermes.cpp
+++ b/API/hermes/hermes.cpp
@@ -29,8 +29,6 @@
 #include "hermes/VM/JSArrayBuffer.h"
 #include "hermes/VM/JSLib.h"
 #include "hermes/VM/JSLib/RuntimeJSONUtils.h"
-#include "hermes/VM/JSObject.h"
-#include "hermes/VM/JSProxy.h"
 #include "hermes/VM/NativeState.h"
 #include "hermes/VM/Operations.h"
 #include "hermes/VM/Profiler/CodeCoverageProfiler.h"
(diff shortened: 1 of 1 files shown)


API/hermes/hermes.cpp
```diff
@@ -1996,7 +1998,20 @@ void HermesRuntimeImpl::setPropertyValue(
 }
 
 bool HermesRuntimeImpl::isArray(const jsi::Object &obj) const {
-  return vm::vmisa<vm::JSArray>(phv(obj));
+  vm::JSObject *jsobj = static_cast<vm::JSObject *>(phv(obj).getPointer());
```
neildhar, line 2001: We have an `isArray` in `Operations.h` that you can call into that will provide the right functionality here.

neildhar (changes requested):
To support this properly, we should fix all users of `jsi::Array` in this file to no longer imply that the underlying object is a `vm::JSArray`. In most cases, this should just be a matter of using `handle()` instead of `arrayHandle()`. But for length we will need to have a fast path that follows the current behaviour, and a slow path that does a regular property lookup.

API/hermes/hermes.cpp
dorentus: Code updated.

Question: Should I call `runtime_.clearThrownValue()` on `ExecutionStatus::EXCEPTION` before returning `false`?

neildhar: You should use `checkStatus` to rethrow the exception as a C++ exception

dorentus: `checkStatus` isn't `const` thus cannot be used here without modification...

dorentus: It should be safe to remove the constness and call `checkStatus` here. Code updated.

dorentus:
https://github.com/facebook/react-native/blame/321f7dbcadb78dede9048500ab8abe86af863061/packages/react-native/ReactCommon/jsc/JSCRuntime.cpp#L993

JSC used to implement `isArray` by calling `Array.isArray` when `_JSC_FAST_IS_ARRAY` is defined, but now it only calls `JSValueIsArray`.

And `JSValueIsArray` behaves the same as `HermesRuntimeImpl::isArray` and returns `false` for `new Proxy([], {})`.

So yes, changes suggested by this PR might cause more problems than expected.

----

As for the [dynamicFromValue](https://github.com/facebook/react-native/blob/main/packages/react-native/ReactCommon/jsi/jsi/JSIDynamic.cpp#L166) problem, I'm trying the `Array.isArray` approach in my local copy of code to see if is has any performance glitches.

tmikov:
Closing for now, since no progress. Please feel free to re-open when there are updates.

Issue #1496, opened 2024-08-27
Many libraries can cause "Error.stack getter called with an invalid receiver"


thomasttvo:
## Bug Description

When an error is thrown while rendering, RN crashes of course as expected. However, for some errors, when `ExceptionsManager` calls `error.stack`, it throws new errors, masking the original errors, which makes it impossible to know what the original error was for users in production (using tools like sentry). 

This happens when libraries using a technique like this to create their error classes:

```ts
    const MyError = function(message) {
      this.message = message;
    };
    MyError.prototype = new Error;
    MyError.prototype.name = 'MyError';
```

I can find at least 5 popular libraries that use this technique, and if it works in classic JS, I think it's supposed to work with hermes?

- [x] The issue is reproducible with the latest version of React Native.

React Native version: 0.73.5
OS: iOS, Android

## Steps To Reproduce

```ts
 useEffect(() => {
    const MyError = function(message) {
      this.message = message;
    };
    MyError.prototype = new Error;
    MyError.prototype.name = 'MyError';

    throw new MyError('1234')
  }, []);

```

## The Expected Behavior

Calling `new MyError().stack` doesn't throw `Error.stack getter called with an invalid receiver`

tmikov:
@thomasttvo hi, `Error.stack` is not part of the ECMAScript spec, so any behavior is valid. However, the example you have shared doesn't seem to work in any engine. I just tried the following:

```js
const MyError = function(message) {
      this.message = message;
    };
    MyError.prototype = new Error;
    MyError.prototype.name = 'MyError';


function foo() {
    throw new MyError('1234')
}
function bar() {
    foo();
}


try {
    bar();
} catch (e) {
    print(e.stack);
}
```

If you run it with v8 or jsc, you get a useless stack trace:
```
v8 test.js
MyError
    at test.js:4:25
```

So, why is this a useful?

thomasttvo:
@tmikov Even though the error doesn't report the stack trace that points to where it is actually thrown, it's useful enough to find out which library it comes from. We have no control over what libraries do with their errors, some throw plain objects, some throw strings. In any of those cases, **_some_** info is good enough, whereas the stack trace of `error.stack` provides no useful info
![image](https://github.com/user-attachments/assets/a9a38edb-8615-4d72-abad-e1d0341ee06a)

tmikov:
[...]

PR #1497, opened 2024-08-27
JSError: look for stack data in the prototype chain

tmikov:
Summary:
Look for stack data in the entire prototype chain in order to accommodate
usage like the one in the test.

See https://github.com/facebook/hermes/issues/1496

Differential Revision: D61870728


thomasttvo:
Thank you so much for the fix 🙏 🙏 🙏 🙏 🙏 
I have one small question. Would there be a way we could avoid raising the type error at the end of `getErrorFromStackTarget`?

tmikov:
I don't think we can avoid throwing an error from the getter when there is nothing to get from.

PR #1520, opened 2024-09-14
typed: exact object properties were stored twice

tmikov:
Summary: There was a missing `return` in IRGen.

Reviewed By: avp

Differential Revision: D62676445


tmikov:
Merged

PR #1523, opened 2024-09-23
RFC[SH]: implement double to int32 truncation without UB

tmikov:
Summary:
On the platforms we support, implement double truncation either with the
available intrinsics or with inline assembly.

Differential Revision: D63240632


matthargett:
Is there a high-level benchmark case this should help with? I'm poking around to see if/when the threeJS benchmark issue will be resolved in time for RN 0.76 or 0.77.

tmikov:
> Is there a high-level benchmark case this should help with? I'm poking around to see if/when the threeJS benchmark issue will be resolved in time for RN 0.76 or 0.77.

@matthargett this change is mostly about reliability. The previous UB behavior was correct, AFAICT, but risky.

IIRC, the slowdown in threeJS is completely unrelated - it was caused by `Array.push()`. We are still working on that.

tmikov:
This has been merged,

Issue #1455, opened 2024-07-09
Different behaviour between Hermes and other JS engines (V8/JSC/QuickJS)


trossimel-sc:
## Bug Description
I noticed a different behaviour between Hermes and other JS engines (tried with QuickJS, JSC and V8) when running the following code

```
if(true) {
    function test(a) {
        if(a == "a") {
            test("b");
            return;
        }
        print ('Error');
    }

    callback1 = test;
}
if(true) {

    function test(a) {
        if(a == "a") {
            test("b");
            return;
        }
        print ('Error1');
    }

    callback2 = test;
}

callback1("a");
callback2("a");
```

## The Expected Behavior
In Hermes the output is `Error1; Error1,` while in QuickJS the output is `Error; Error1`.
Is this behaviour expected?

avp:
Thanks for the report. Looks like we're resolving the `test()` call in the first `test()` function incorrectly. We'll figure out how to solve this.

tmikov:
I hope this isn't code that is actually used anywhere. It isn't even valid in strict mode, which is hopefully the default everywhere.

trossimel-sc:
@tmikov it's just being used for testing this behaviour

PR #1547, opened 2024-10-21
Fix function hoisting behaviour (non-strict mode)

trossimel-sc:
## Summary

Addresses #1455 

Hermes currently has divergent function hoisting behavior compared to QuickJS, V8, and JSC in non-strict mode. The following code prints `FAIL` instead of `SUCCESS`

```
{
    function promotedFunction(endRecursion) {
        if(endRecursion) {
            promotedFunction(true);
            return;
        }
        print ('SUCCESS');
    }

    callback = promotedFunction;
}

{
    function promotedFunction(endRecursion) {
        if(endRecursion) {
            promotedFunction(true);
            return;
        }
        print ('FAIL');
    }
}

callback(false);
```

This happens as scoped function promoted to global scope are treated as global functions, like for `var`. 
To fix this behaviour:

- Two declarations are created for a promoted identifier instead of one: a global and a scoped declaration, stored in `sideIdentifierDeclarationDecl_` and `promotedFunctionDecls_`
- If the identifier is promoted, we call `emitStore` on both pointers associated to the declarations

## Test Plan

- Added a unit test with the failing code


trossimel-sc:
Seems like the failing errors are due to auto generated checks

avp:
@trossimel-sc The `update-lit` build target will update the tests that are autogenerated, then `check-hermes` to run them. [`LIT_FILTER`](https://llvm.org/docs/CommandGuide/lit.html#cmdoption-lit-filter) will let you narrow down what exactly gets updated or run as well. Thanks for submitting this fix!

tmikov:
Hi, we really need the CLA before we can review it internally. Thanks!

trossimel-sc:
The CLA is signed and we can proceed with the review process 😄 .
We can start with this PR and `for await of` PR.
I also noticed the build is failing on `static_h`. Is this happening to me only?

lib/Sema/SemanticResolver.cpp
```diff
@@ -2166,6 +2166,14 @@ void SemanticResolver::validateAndDeclareIdentifier(
             "Can't create duplicate variable that shadows a global property: '") +
             ident->_name->str() + "'");
   }
+    
+  if (semCtx_.getDeclarationDecl(ident) &&
+      functionContext()->promotedFuncDecls.count(ident->_name)) {
+    decl = semCtx_.newDeclInScope(ident->_name, kind, curScope_);
+    bindingTable_.put(ident->_name, Binding{decl, ident});
```
avp, line 2173: This needs a comment explaining what the special case is doing.

e.g. Why is it using `put` to override instead of `try_emplace` as below?

lib/IRGen/ESTreeIRGen.h
```diff
@@ -1305,6 +1305,13 @@ class ESTreeIRGen {
     return getDeclData(getIDDecl(id));
   }
 
+  /// Resolve the identifier node associated with the promoted function
+  /// to the corresponding global variable in expression context.
+  Value *resolveIdentifierPromoted(ESTree::IdentifierNode *id) {
+    auto declPromoted = getIDDeclPromoted(id);
```
avp, line 1311:
```suggestion
    auto *declPromoted = getIDDeclPromoted(id);
```

avp (changes requested):
Thanks!

Overall this seems like a reasonable solution to me, just one question/clarification on maintaining the binding table.

avp:
> I also noticed the build is failing on static_h.

That should be fixed with ae186905e9b2d6c30e7015d3b418fba7aea51d11

trossimel-sc:
Thank you @avp !
Addressed :)

avp:
@trossimel-sc The internal failure is due to a bad interaction with lazy compilation.  Test case:
```
(function() {
  { function* x() {} }
})();

(function() {
  function* y() { yield 1; }
})();
```

Run that with
```
bin/hermes ~/test.js --lazy
```

and there'll be an assertion failure:
```
Assertion failed: (llvh::isa<CreateScopeInst>(RSI) && "variable must be initialized in its own function"), function emitStore, file ESTreeIRGen.cpp, line 1208.
fish: Job 1, 'bin/hermes ~/test.js --lazy' terminated by signal SIGABRT (Abort)
```

Some documentation about how lazy compilation works: https://github.com/facebook/hermes/blob/static_h/doc/LazyEvalCompilation.md

I haven't gone through and debugged but there's potentially some extra work that has to be done in the lazy compilation path due to the new information you've added to SemContext.

trossimel-sc:
Thank you, I am trying to understand the problem and find a way to fix it.
Seems like the address of identifier `x` and `y` are the same at some point, and even when a promoted decl is not set for `y` in `validateAndDeclareIdentifier`, the function `resolveIdentifierPromoted` does not return nullptr for `y`. Quite weird

lib/Sema/SemanticResolver.cpp
```diff
@@ -93,6 +93,8 @@ bool SemanticResolver::runLazy(
     if (astContext_.getDebugInfoSetting() != DebugInfoSetting::ALL) {
       semInfo->bindingTableScope.reset();
     }
+
+    semCtx_.clearPromotedDecls();
```
avp, line 97: I think this is the right fix, good catch.

Let's also put a comment so we don't forget why: the `promotedDecls` is used immediately, and keyed off the `IdentifierNode`, which is freed between invocations of lazy compilation. We need to remove references in `SemContext` when freeing the AST, which happens at the end of each invocation of lazy compilation.

avp:
@trossimel-sc We have one more issue with lazy compilation.

Test case:
```
function foo() {
  var result;
  {
    let x = 0;
    result = ret;
    function ret() {
      return x;
    }
  }
  return result;
}

print(typeof foo());
```

This should print `function` but it prints `undefined` in lazy mode.

lib/BCGen/HBC/HBC.cpp
```diff
@@ -195,6 +195,12 @@ static void compileLazyFunctionWorker(void *argPtr) {
     return;
   }
 
+  // promotedDecls is used immediately and relies on IdentifierNode,
+  // which is deallocated between invocations of lazy compilation.
+  // Therefore, it is necessary to clear promotedDecls when freeing the
+  // AST in order to avoid dangling pointers.
+  semCtx->clearPromotedDecls();
```
trossimel-sc, line 202: I noticed that I was clearing the promoted declarations after the AST generation and before the IR generation. Now I am doing it after the IR generation. Let's see if it passes the tests. Also, is there a more appropriate place to put this line of code?

avp:
d1cd9925c1b9deebfe7ca1c18dc5ed046b050c20

Issue #1549, opened 2024-10-25
`@babel/plugin-proposal-decorators` is not supported in `hermes-parser` - was: Broken babel preset setup with @babel/plugin-proposal-decorators and React Native 0.76.0


lucaswitch:
### Since i upgraded from react-native 0.75.4 to 0.76 and the babel plugin "@babel/plugin-proposal-decorators" stopped working.

What we know:

I was not able to fix the issue by changing babel or @babel/plugin-proposal-decorators versions so probably is not something with babel setup.
I can confirm that brand new react-native 0.76.0 can not work with @babel/plugin-proposal-decorators as we can see on the reproducer.

The reproducer is a standard react-native 0.76.0 with    "@babel/plugin-proposal-decorators": "^7.25.9"  and a babel.config.js that i believe most of the users configure.
The reproducer can be found on:
https://github.com/lucaswitch/broken-babel-preset

**I understand that it's not a problem of react-native project** but is something that we can probably can be adjusted on the official babel preset. 
Actually this plugin is used by a lot of developers including me and a lot of libraries need that configuration to be able to be used on a react-native project. 

### Steps to reproduce

I tested this issue only on IOS but i think it's not related to any platform since is a babel related problem.

I'm not a babel expert but i suspect there is something on @react-native/babel-preset that broke this configuration.
Created a little repro with a brand new bare react-native project and an small minimal decorator setup just to prove that is not transforming js syntax the right way on react-native.

![Screenshot 2024-10-25 at 16 26 09](https://github.com/user-attachments/assets/914f761e-37a7-4d30-9577-2cb23c703bc3)

### React Native Version

0.76.0

### Affected Platforms

Runtime - iOS

### Output of `npx react-native info`

```text
System:
  OS: macOS 14.0
  CPU: (24) x64 Intel(R) Xeon(R) CPU E5-2670 v3 @ 2.30GHz
  Memory: 30.60 MB / 16.00 GB
  Shell:
    version: "5.9"
    path: /bin/zsh
Binaries:
  Node:
    version: 22.10.0
    path: /usr/local/bin/node
  Yarn:
    version: 1.22.19
    path: /usr/local/bin/yarn
  npm:
    version: 10.9.0
    path: /usr/local/bin/npm
  Watchman:
    version: 2024.10.21.00
    path: /usr/local/bin/watchman
Managers:
  CocoaPods: Not Found
SDKs:
  iOS SDK:
    Platforms:
      - DriverKit 23.2
[...]

PR #1556, opened 2024-10-31
Skip plugin for non-Flow JS code

huntie:
Summary:
This is a mitigation for https://github.com/facebook/hermes/issues/1549.

Updates `babel-plugin-syntax-hermes-parser` to include a new `parseLangTypes` option, which we will configure in React Native's Babel preset to abort when the file contents do not include `@flow`.

**Context: React Native**

Originally changed in https://github.com/facebook/react-native/pull/46696, as our internal Flow support had diverged from `babel/plugin-syntax-flow` (https://github.com/facebook/react-native/issues/46601).

We effectively have three flavours of JavaScript in support:
- Flow@latest for the `react-native` package, shipped as source — uses `hermes-parser`.
- TypeScript for product code (community template, Expo) — uses `babel/plugin-syntax-typescript`.
- Plain JavaScript/JSX in product code, *which may be extended with additional user Babel plugins and needs lenient parsing* — uses base `babel/parser` (**this change**).

I'd love to simplify this 😅.

Differential Revision: D65272155

### DIFF
--- a/tools/hermes-parser/js/babel-plugin-syntax-hermes-parser/__tests__/babel-plugin-syntax-hermes-parser-test.js
+++ b/tools/hermes-parser/js/babel-plugin-syntax-hermes-parser/__tests__/babel-plugin-syntax-hermes-parser-test.js
@@ -16,22 +16,44 @@ import {transformSync} from '@babel/core';
 import hermesParserPlugin from '../src';
 import * as HermesParser from 'hermes-parser';
 
-const MODULE_PREAMBLE = '"use strict";\n\n';
+const MODULE_PREAMBLE = '// @flow\n\n"use strict";\n\n';
+const NON_FLOW_MODULE_PREAMBLE = '"use strict";\n\n';
 
 describe('babel-plugin-syntax-hermes-parser', () => {
-  test('test basic parsing', () => {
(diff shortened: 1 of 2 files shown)


jbroma:
Doesn’t `hermes-parser` handle this internally? I'm curious—how does this differ from the flow: 'detect' option in hermes-parser?

Reference: https://github.com/facebook/hermes/blob/8e238330effbbea6d2934f429839a74d1131a1a4/tools/hermes-parser/js/hermes-parser/src/index.js#L26

huntie:
@jbroma There's `hermes-parser` (directly) and `babel-plugin-syntax-hermes-parser` (this integration). Looks like this plugin is the point correctly to eject out of hermes-parser handling (to implement the `flow` arg) under Babel.

jbroma:
> @jbroma There's `hermes-parser` (directly) and `babel-plugin-syntax-hermes-parser` (this integration). Looks like this plugin is the point correctly to eject out of hermes-parser handling (to implement the `flow` arg) under Babel.
> 
> The test plan for this change requires [`transformer.hermesParser`](https://metrobundler.dev/docs/configuration#hermesparser) to be set `false` in Metro.

thanks for clearing that up! makes sense 👍

PR #1559, opened 2024-11-04
[WIP] Add for await of

trossimel-sc:
## Summary

This PR introduces support for the `for await of` syntax through IR.

In detail:
- The helper function `_makeAsyncIterator` is used to create an async iterator for a given iterable. It checks if an async iterator method is available; if not, it creates an async iterator from a synchronous iterator by wrapping its results in promises.
- Added a `genAsyncForOfStatement` which handles the for await of loop by iterating over the async iterator object. It follows the same rules described [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of)

## Test Plan
- Unit tests
- Test262

### DIFF
--- a/include/hermes/VM/RuntimeModule.h
+++ b/include/hermes/VM/RuntimeModule.h
@@ -51,6 +51,11 @@ union RuntimeModuleFlags {
     /// Whether this runtime module's epilogue should be hidden in
     /// runtime.getEpilogues().
     bool hidesEpilogue : 1;
+
+    /// Whether we want to ignore ES6 promise checks.
+    /// This is used to force compiling modules with ES6 promises even if the
+    /// runtime is not configured to support them.
+    bool ignoreES6PromiseChecks : 1;
   };
   uint8_t flags;
   RuntimeModuleFlags() : flags(0) {}
--- a/lib/IRGen/ESTreeIRGen.cpp
+++ b/lib/IRGen/ESTreeIRGen.cpp
@@ -480,12 +480,40 @@ Value *ESTreeIRGen::emitIteratorSymbol() {
       "iterator");
 }
 
+Value *ESTreeIRGen::emitAsyncIteratorSymbol() {
+  return Builder.createLoadPropertyInst(
+      Builder.createGetBuiltinClosureInst(BuiltinMethod::globalThis_Symbol),
+      "asyncIterator");
+}
+
 ESTreeIRGen::IteratorRecordSlow ESTreeIRGen::emitGetIteratorSlow(Value *obj) {
   auto *method = Builder.createLoadPropertyInst(obj, emitIteratorSymbol());
   auto *iterator = Builder.createCallInst(
       method, /* newTarget */ Builder.getLiteralUndefined(), obj, {});
 
   emitEnsureObject(iterator, "iterator is not an object");
+  auto *nextMethod = Builder.createLoadPropertyInst(iterator, "next");
+
+  return {iterator, nextMethod};
+}
+
+ESTreeIRGen::IteratorRecordSlow ESTreeIRGen::emitGetAsyncIteratorSlow(
+    Value *obj) {
+  auto *asyncIteratorMethod =
+      Builder.createLoadPropertyInst(obj, emitAsyncIteratorSymbol());
+  auto *syncIteratorMethod =
+      Builder.createLoadPropertyInst(obj, emitIteratorSymbol());
+
+  auto *wrapper = Builder.createLoadPropertyInst(
+      Builder.getGlobalObject(), "HermesAsyncIteratorsInternal");
+  wrapper = Builder.createLoadPropertyInst(
+      wrapper, "_makeAsyncIterator");
+
+  auto *iterator = Builder.createCallInst(
+      wrapper,
+      Builder.getLiteralUndefined(),
+      Builder.getLiteralUndefined(),
+      {obj, asyncIteratorMethod, syncIteratorMethod});
   auto *nextMethod = Builder.createLoadPropertyInst(iterator, "next");
 
   return {iterator, nextMethod};
--- a/lib/InternalJavaScript/04-AsyncIterator.js
+++ b/lib/InternalJavaScript/04-AsyncIterator.js
@@ -1,0 +1,38 @@
+var initAsyncIterators = function() {
+    function _makeAsyncIterator(iterable, asyncIterMethod, syncIterMethod) {
+        if (asyncIterMethod) {
+            return asyncIterMethod.call(iterable);
+        }
+
+        let syncIterator = syncIterMethod.call(iterable);
+        async function handleResult(result) {
+            if (result.done) {
+                return result;
+            }
+            const value = result.value;
+            const resolvedValue = value instanceof Promise ? await value : value;
+            return { done: false, value: resolvedValue };
+        }
+
+        return {
+            async next() {
+                const result = syncIterator.next();
+                return handleResult(result);
+            },
+            async return(value) {
+                const result = typeof syncIterator.return === 'function' ? syncIterator.return(value) : { done: true, value };
+                return handleResult(result);
+            },
+            async throw(error) {
+                const result = typeof syncIterator.throw === 'function' ? syncIterator.throw(error) : { done: true, value: error };
+                return handleResult(result);
+            }
+        };
+    }
+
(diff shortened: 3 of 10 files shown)


include/hermes/VM/RuntimeModule.h
```diff
@@ -51,6 +51,11 @@ union RuntimeModuleFlags {
     /// Whether this runtime module's epilogue should be hidden in
     /// runtime.getEpilogues().
     bool hidesEpilogue : 1;
+
+    /// Whether we want to ignore ES6 promise checks.
+    /// This is used to force compiling modules with ES6 promises even if the
+    /// runtime is not configured to support them.
+    bool ignoreES6PromiseChecks : 1;
```
tmikov, line 58: This seems counterintuitive. The correct behavior is to fail if code requiring promises is executed with promises disabled. I think you should delete this flag.

FWIW, we are removing the option to run without ES6 promise anyway.

lib/InternalJavaScript/04-AsyncIterator.js
```diff
@@ -0,0 +1,42 @@
...
+                return handleResult(result);
+            }
+        };
+    }
+
+    var HermesAsyncIteratorsInternal = {
+        _makeAsyncIterator
+    };
+    Object.freeze(HermesAsyncIteratorsInternal);
+    globalThis.HermesAsyncIteratorsInternal = HermesAsyncIteratorsInternal;
```
avp, line 37: We have a `JS_BUILTIN` mechanism in `Builtins.def` that's used for exposing these JS functions to the CallBuiltin mechanism, makes calling it much easier/faster. See the `spawnAsync` function for an example of how we do this for async functions. It should simplify things.

include/hermes/VM/RuntimeModule.h
trossimel-sc: Yes you are right, and I need to check again this PR as we don't have this in our codebase either. Seems like an older commit. Thanks for pointing out!

lib/InternalJavaScript/04-AsyncIterator.js
trossimel-sc: Thank you, addressed!
However, It does not compile using `JS_BUILTIN` but it does with `PRIVATE_BUILTIN`. 
Can `JS_BUILTIN` already be used more than once?

include/hermes/VM/RuntimeModule.h
tmikov: Good point. @avp has a commit to remove the option to run without promises. That should land very soon and then you should be good to go.

lib/InternalJavaScript/04-AsyncIterator.js
tmikov: I see the problem. Fix incoming.

include/hermes/VM/RuntimeModule.h
avp: https://github.com/facebook/hermes/commit/27a5377e9fa393572e56708d6b58fde623f3d962

lib/InternalJavaScript/04-AsyncIterator.js
tmikov: The fix is landing, but that may take a few hours. Meanwhile, if you are blocked, you can temporarily append a comma here:
https://github.com/facebook/hermes/blob/27a5377e9fa393572e56708d6b58fde623f3d962/lib/VM/Runtime.cpp#L1815

Here is a more complete example of how to add a JS builtin:
https://github.com/facebook/hermes/pull/1649

tmikov: Fix landed in https://github.com/facebook/hermes/commit/874b778fbf16c0242d4f0bac6530497de7ed05fb

include/hermes/VM/RuntimeModule.h
trossimel-sc: Thank you for the help! Addressed all suggestions

lib/InternalJavaScript/04-AsyncIterator.js
```diff
@@ -0,0 +1,45 @@
+function initAsyncIterators() {
+    function _makeAsyncIterator(iterable, asyncIterMethod, syncIterMethod) {
+        if (asyncIterMethod) {
+            return asyncIterMethod.call(iterable);
+        }
+
+        let syncIterator = syncIterMethod.call(iterable);
```
avp, line 7: Proper block scoping is still behind a flag in SH, so let's avoid the behavior changing here unless necessary and just use `var` for now (everywhere in this file)

lib/InternalJavaScript/04-AsyncIterator.js
```diff
@@ -0,0 +1,45 @@
...
+        async function handleResult(result) {
+            if (result.done) {
+                return result;
+            }
+            const value = result.value;
+            const resolvedValue = value instanceof Promise ? await value : value;
```
avp, line 13: Let's capture `Promise` as a local and use it, to avoid having to do a global lookup here as well as avoiding potential user changes to `Promise`.

avp (commented):
Thanks, a few comments on the internal JS.

lib/InternalJavaScript/04-AsyncIterator.js
trossimel-sc: Thanks, addressed all 👍

trossimel-sc:
@avp I just noticed we also need to define `Symbol.asyncIterator` as it's undefined in hermes

avp:
Hi, we've been reviewing this carefully and there's two issues we've identified that need to be handled before this can land.

1. This needs to handle block scoping (with TDZ, etc). Synchronous for-of handles this, so you should just need to transplant those changes. Here's a diff that adds that as well as a test that it works: https://gist.github.com/avp/798a795f798fb589eb48ae58bd5e51de

2. Closing the async iterator requires an `await`. Looks like when an async for-of ends, it needs to invoke the [AsyncIteratorClose](https://tc39.es/ecma262/#sec-asynciteratorclose) operation, which in step 4.d. awaits its result. We need to make sure the Promise given back by the `return` method gets awaited before continuing to execute any code after the loop.

Thanks!

trossimel-sc:
Thanks @avp,
One last thing: do you know how to address IR errors that occur after adding `genYieldOrAwaitExpr` as part of the try/catch block in `_emitIteratorCloseImpl` ? I didn't expect it to fail

avp:
`genYieldOrAwait` results in returns when special generator instructions are lowered out, so it's possible the problem is that there's issues with not ending the `Try` before returning.

avp:
@trossimel-sc Looks like your IRVerifier problems are the result of a slight oversight in `emitTryCatchScaffolding`'s body callback in `_emitIteratorCloseImpl`.

The old version of the code didn't instantiate a `SurroundingTry` even though it's technically required to. This just wasn't an observable problem in any way because there was nothing inside to read the `SurroundingTry`. Now that this PR adds more logic inside that body callback, it may be read so it must be instantiated.

So, much like for your other call to `emitTryCatchScaffolding`, you must instantiate a `SurroundingTry` at the top of the body callback and that should fix your IRVerifier errors.

Thanks!

trossimel-sc:
@avp Thank you!

lib/IRGen/ESTreeIRGen.cpp
```diff
@@ -619,12 +642,25 @@ void ESTreeIRGen::emitIteratorCloseSlow(
...
+          SurroundingTry thisTry{
+            curFunction(),
+            astNode,
+            catchBlock,
+            {},
+            [this, &iteratorRecord, isAsyncIterator, returnMethod](
+                ESTree::Node *,
+                ControlFlowChange cfc,
+                BasicBlock *continueTarget) {
+              auto *callResult = Builder.createCallInst(
```
avp, line 655: Why was this moved into the finalizer instead of left in the body block like it used to be?

avp (changes requested):
I think this doesn't handle the "throw in a loop" case properly. We need more tests:

1. `for-await-of-return.js` test which doesn't `break` inside the loop but instead throws an Error. That should also perform the cleanup in the correct order.
2. `for-await-of-return.js` test which throws an Error in the loop along with a `return()` function that _also_ throws an error.
3. Similar test which `break`s out of the loop, but `return()` still throws.

Issue #1056, opened 2023-07-17
Feature Request: `with` statement for SES (Secure EcmaScript) support


leotm:
## Problem

Continuing our [convo](https://github.com/facebook/hermes/issues/957#issuecomment-1561914135)
- https://github.com/facebook/hermes/issues/957

> FWIW, I think we might be able to add support for `with`. It feels wrong to completely ignore a language feature that has been with us for many years. It also feels good to pass most of test262. Plus, the implementation is interesting and will not slow everything down. But I can't promise it is high priority

in order to introduce [SES](https://github.com/endojs/endo/tree/master/packages/ses) to our RN ecosystem (starting with [metamask-mobile](https://github.com/MetaMask/metamask-mobile))

we can get by without _local eval_, but `with` statements are [required](https://github.com/facebook/hermes/issues/957#issuecomment-1560267961)

## Solution

`with` statements support 🙏

Alternatives considered
- _facebook/hermes_ community contribution: `with` statements
- more detailed in https://github.com/facebook/hermes/issues/957
- https://github.com/endojs/endo/issues/1561

## Additional Context

https://github.com/facebook/hermes/blob/main/doc/Features.md#excluded-from-support
> Excluded From Support
> - `Symbol.unscopables` (Hermes does not support `with`)
> - `with` statements

hermesengine.dev/playground

```console
# with ([1, 2, 3]) {
#   console.log(toString()); // 1,2,3
# }

/tmp/hermes-input.js:1:1: error: invalid statement encountered.
with ([1, 2, 3]) {
^~~~~~~~~~~~~~~~~~
Emitted 1 errors. exiting.
```

leotm:
@tmikov noting this won't slow everything down, but not promised as high priority
do you have an update on the timeframe? and would you be happy to accept a community contribution?

keen to know more on the interesting implementation and any pointers (like given for eval in the [past](https://github.com/facebook/hermes/issues/102#issuecomment-528489267)) to give it a try

tmikov:
@leotm unfortunately the answer is tricky. Ordinarily we are very happy to accept community contributions (although historically we have gotten very few, and the bar for changing the compiler is quite high).

The problem is that the current version of Hermes is frozen, except for bug fixes, while the next version, which is developed in a branch, has diverged significantly, particularly in the area of variable resolution and code generation, which is precisely where `with` would be handled. Worse, the next version is changing very rapidly at the moment, so it is not a very suitable target.

[...]

PR #1571, opened 2024-11-27
[WIP] Add with statement support

trossimel-sc:
## Summary
This PR introduces support for the with statement via IR

- Identifiers within a with statement are transformed into a series of conditional chain. Each condition checks the presence of a key in the object properties, and accesses it if present. This is done through `createConditionalChainImpl`
- We need to keep track of each with statement depth. Hermes uses a task queue to lazily compile functions, and for this reason, we also need to copy the with depths when adding the declaration to the queue.

## Test Plan
- Added a unit test covering most common scenarios
- Test262 result on with statements tests: 92.98%.


trossimel-sc:
While the output is generally correct, I've noticed that when running in debug mode, it occasionally fails with the following message:

```
Assertion failed: (I->hasOutput() && "Instruction has no output"), function encodeValue, file ISel.cpp, line 264.
```

And in this case the Intermediate Representation (IR) seems to be incorrectly generated. This issue arises specifically when a store operation is present within the `with` statement. Here's an example:

```javascript
let e = 20;
with (obj) {
    e = 30;
}
```

While the output is correct, I am trying to understand what am I doing wrong. Is there a straightforward way to handle this case? I'm not very familiar with Hermes IR, so I'm unsure why this failure occurs. Thank you for your help!

Please feel free to respond once my CLA is signed, if needed. —I hope this process won't take much longer. 🫤

tmikov:
@trossimel-sc I haven't reviewed the PR, but I compiled it and looked at the IR output of your example:

```
  %10 = CallInst (:any) %8: any, empty: any, false: boolean, empty: any, undefined: undefined, undefined: undefined, %9: any, "e": string
  %11 = CondBranchInst %10: any, %BB2, %BB1
%BB1:
  %12 = StoreFrameInst %0: environment, 30: number, [%VS0.e]: any
  %13 = BranchInst %BB3
%BB2:
  %14 = LoadFrameInst (:any) %0: environment, [%VS0.with1]: any
  %15 = StorePropertyLooseInst 30: number, %14: any, "e": string
  %16 = BranchInst %BB3
%BB3:
  %17 = PhiInst (:notype) %15: notype, %BB2, %12: notype, %BB1
  ```
  
You are calling a helper method to check if `e` exists in the object and branching to either a property store `%15` or a variable store `%12`. The problem is that you are treating these stores as if they have a result and in instruction `%17` you are reading this result. Stores have no result, they just have side effects.

AFAICT, instruction `%17` is incorrect and more importantly, unnecessary. You don't need to combine the stores with ` Phi`, they have already done their side effect.

BTW, I think that we probably want a new bytecode instruction for this instead of calling a helper, but we can do that incrementally. Also, the helper should just be a "builtin", which is safer and easier to call. But these are details that we can address when we start the review.

(We don't want to start reviewing before we know that we can accept the PR, I hope you understand)

trossimel-sc:
Thanks for the response and the assistance @tmikov 
Just wanted to update, I'm still waiting for my request to be approved by the company. Unfortunately, it's taking longer than expected, and there's not much I can do except wait for now 🫤

leotm:
great work 🎉 adding _Fixes: https://github.com/facebook/hermes/issues/1056_ to description should link the issue to this PR (auto-close on merge)

dream2333:
Any progress here?

PR #1580, opened 2024-12-09
Add line breaks for better meaning

zhuyongqi9:
## Summary

Add line breaks for better meaning

## Test Plan

No need， just update README


tmikov:
Hi, can you please submit this against the static_h branch. The main branch is effectively frozen.

PR #1604, opened 2025-01-29
- Add conversion of the iterator from Flow to TS

coado:
Summary:
`@iterator` and `@asyncIterator` are valid syntaxes for Flow but not for TS which accepts [Symbol.iterator] and [Symbol.asyncIterator]. According to flow lexer these are the only valid syntaxes in this form so checking for other cases prefixed with `@@` should not be required.

Changelog:
[flow-api-translator] - Added conversion of the iterator from Flow to TS

Differential Revision: D68776171


tmikov:
@coado is there a reason to export these to GitHub before they land?

robhogan:
> @coado is there a reason to export these to GitHub before they land?

Maybe(?) no good reason in the Hermes repo, but for context we enforce export before land in RN to ensure changes don't break OSS CI.

coado:
> @coado is there a reason to export these to GitHub before they land?

My bad, I thought that there is the same procedure with exporting to GH as with react-native.

PR #1657, opened 2025-04-03
Add async generators support via AST transformation

trossimel-sc:
## Summary
This PR introduces support for async generators through AST transformations. In detail:

- Added AsyncGenerator.cpp to transform asynchronous generators into standard generators using AST transformations. This implementation mirrors the logic found in [Babel ](https://babeljs.io/docs/babel-plugin-transform-async-generator-functions).
-  Introduced an EnableAsyncGenerators flag to manage this functionality.

## Test Plan
- Added unit tests


lib/Sema/SemResolve.cpp
```diff
@@ -162,6 +163,10 @@ bool resolveAST(
     flow::FlowContext *flowContext,
     ESTree::ProgramNode *root,
     const DeclarationFileListTy &ambientDecls) {
+  if (astContext.getEnableAsyncGenerators()) {
+    transformAsyncGenerators(astContext, root);
```
trossimel-sc, line 167: Is there a way to add AST transformation in lazy mode? I tried by running `transformAsyncGenerators` in resolveASTLazy, but it does not let me change the tree.

avp: I haven't really started to review this PR yet (working on getting for-await-of committed right now) but I've put up a PR for a new compiler stage that should work with lazy and eager compilation. It does the transformation in a single place so that things can be kept track of in a principled manner.

https://github.com/facebook/hermes/pull/1658

Can you try it out and see if it works for your use case here? You just need to fill out `TransformAST.cpp` so that it returns a node of the same type as was passed in (it can be the same node). If it works, I'll commit it.

trossimel-sc: It works, but not In lazy compilation:

`Assertion failed: (saveNode == node && "node must not be replaced in-place with a no-replace visitor"), function visitESTreeNodeNoReplace, file RecursiveVisitor.h, line 568.`

trossimel-sc: I disabled the test in lazy compilation for now.
I am not sure of the issue, may it be related to `transformASTForCompilation` being called twice just in lazy compilation? 🤔

trossimel-sc: Addressed and enabled unit tests in lazy compilation, thanks!

PR #1676, opened 2025-04-22
[WIP] - Fix Empty Type Parameters

nickhudkins:
## Summary
While working on a project that utilizes `metro`, I found that some of the type definitions were incorrect:
https://github.com/facebook/metro/issues/1491

## Test Plan
Flow's empty type parameter syntax should properly translate to empty TypeScript type parameters. Currently `flow-api-translator` will happily translate:

`type Foo = Bar<>` to `type Foo= Bar<>` which is invalid TypeScript.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,105 +10,105 @@ jobs:
     env:
       HERMES_WS_DIR: ${{ github.workspace }}
     steps:
-    - name: Set up workspace and install dependencies
-      run: |-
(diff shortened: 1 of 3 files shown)


tmikov:
Not sure about the rest, but `build.yml` has a ton of spurious changes that would have to be cleaned up

nickhudkins:
Holy smokes, this was a push from a new machine with holy moly extensions enabled I guess. I believe I likely should be working off `static_h` instead of `main` as well.

PR #1677, opened 2025-04-22
tool/node-hermes/third-party/libuv/src/idna: fix OOB read in punycode decoder

npt-1707:
## Summary
**Description**
This PR fixes a potential vulnerability of an out-of-bounds read in uv__idna_toascii() that was cloned from libuv but did not receive the security patch. The original issue was reported and fixed under https://github.com/libuv/libuv/commit/b7466e31e4bee160d82a68fca11b1f61d46debae.
This PR applies the same patch to eliminate the vulnerability.

**References**
https://nvd.nist.gov/vuln/detail/cve-2021-22918
https://github.com/libuv/libuv/commit/b7466e31e4bee160d82a68fca11b1f61d46debae

### DIFF
--- a/tools/node-hermes/third-party/libuv/src/idna.c
+++ b/tools/node-hermes/third-party/libuv/src/idna.c
@@ -19,6 +19,7 @@
 
 #include "uv.h"
 #include "idna.h"
+#include <assert.h>
 #include <string.h>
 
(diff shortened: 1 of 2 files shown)


tmikov:
Hi, thanks for this PR! Unfortunately node-hermes is not supported and is in the process of being deleted, so we can't accept PRs for it.

PR #1693, opened 2025-04-25
fix: reset cache for timezone changes

### LINKED ISSUE #1121: Daylight savings time not handled properly in America/Santigo.
## Bug Description

Dates set to midnight of the day of a DST transition in Chile get moved to the prior day at 11 vs the intended day at 1am. 

This is inconsistent with the behavior of other runtimes - there may be further similar issues with other timezones.

- [x] I have run `gradle clean` and confirmed this bug does not occur with JSC

Hermes version: current
React Native version (if any): 0.71.8
OS version (if any): os x, ios, android
Platform (most likely one of arm64-v8a, armeabi-v7a, x86, x86_64): arm64-v8a

## Steps To Reproduce

1. Set your machine timezone to `America/Santiago`
2. Go to the hermes playground
3. Paste in this code
```
const first = new Date(2023,8,1);
const second = new Date(2023,8,2);
const third = new Date(2023,8,3);
const fourth = new Date(2023,8,4);

print(first);
print(second);
print(third);
print(fourth);
```
This will output:

```
Fri Sep 01 2023 00:00:00 GMT-0400
Sat Sep 02 2023 00:00:00 GMT-0400
Sat Sep 02 2023 23:00:00 GMT-0400
Mon Sep 04 2023 00:00:00 GMT-0300
```

## Expected behaviour

On v8 / jsc you see the following:

```
Fri Sep 01 2023 00:00:00 GMT-0400 (Chile Standard Time)
Sat Sep 02 2023 00:00:00 GMT-0400 (Chile Standard Time)
Sun Sep 03 2023 01:00:00 GMT-0300 (Chile Summer Time)
Mon Sep 04 2023 00:00:00 GMT-0300 (Chile Summer Time)
```

As is this causes the popular date-fns library to spin into an infinite loop in some cases. 
See this function for an example: https://github.com/date-fns/date-fns/blob/main/src/eachDayOfInterval/index.ts

If you're in `America/Santiago` the following will loop forever:

```
const days = eachDayOfInterval({
  start: parseISO("2023-09-01"),
  end: parseISO("2023-09-04")
});
```

troZee:
## Summary
A new native function resetTimezoneCache has been added to HermesRuntime object. This function handles resetting the timezone cache, while the timezone is changed

This fix has been implemented based on the @tmikov recommendation:
 
```
We propose adding a JSI method to HermesRuntime for reseting the TZ cache. The integrator (React Native) can then call this method after detecting that the TZ has changed, using whatever approach is suitable for the platform.
```

## Test Plan 

1. Add the below code to `include/hermes/VM/JSLib/DateCache.h`
```
  void reset() {
    std::cout << "Calling reset" << std::endl;
```    

2. Build Hermes 
3. Run the unit tests and check if this method is called 

## How to reproduce it?

1. `bunx @react-native-community/cli@latest init ReactNativePlayground769 --version 0.76.9`
2.  Replace the App.tsx with the below code 
```
/**
 * Sample React Native App
 * https://github.com/facebook/react-native
 *
 * @format
 */

import React, {useEffect, useState} from 'react';
import type {PropsWithChildren} from 'react';
import {
  AppState,
  SafeAreaView,
  ScrollView,
  StatusBar,
  StyleSheet,
  Text,
  useColorScheme,
  View,
} from 'react-native';

import {
  Colors,
  DebugInstructions,
  Header,
  LearnMoreLinks,
  ReloadInstructions,
} from 'react-native/Libraries/NewAppScreen';

import * as RNLocalize from 'react-native-localize';

const useTimezoneChange = () => {
  const [timezone, setTimezone] = useState(RNLocalize.getTimeZone());

  useEffect(() => {
    // Set up listener for timezone changes
    const subscription = AppState.addEventListener('change', nextAppState => {
      if (nextAppState === 'active') {
        const newTimezone = RNLocalize.getTimeZone();
        if (newTimezone !== timezone) {
          setTimezone(newTimezone);
          // Handle timezone change here
          console.log('Timezone changed to:', newTimezone);
        }
      }
    });

    // Clean up listener on unmount
    return () => {
      subscription.remove();
    };
  }, [timezone]);

  return timezone;
};

const TimezoneAwareComponent = () => {
  const currentTimezone = useTimezoneChange();

  const date = new Date(
    'Mon Apr 28 2025 22:05:36 GMT+0900 (Japan Standard Time)',
  );
  date.getTimezoneOffset();
  return (
    <View>
      <Text>Current timezone: {currentTimezone}</Text>
      <Text>
        Date: {date.getHours()}:{date.getMinutes()}
      </Text>
    </View>
  );
};

const DateComponent = () => {
  const date = new Date(
    'Mon Apr 28 2025 22:05:36 GMT+0900 (Japan Standard Time)',
  );
  return (
    <View>
      <Text>
        Fixed Date: {date.getHours()}:{date.getMinutes()}
      </Text>
    </View>
  );
};

function App(): React.JSX.Element {
  const isDarkMode = useColorScheme() === 'dark';

  const backgroundStyle = {
    backgroundColor: isDarkMode ? Colors.darker : Colors.lighter,
  };

  return (
    <SafeAreaView style={backgroundStyle}>
      <StatusBar
        barStyle={isDarkMode ? 'light-content' : 'dark-content'}
        backgroundColor={backgroundStyle.backgroundColor}
      />
      <ScrollView
        contentInsetAdjustmentBehavior="automatic"
        style={backgroundStyle}>
        <Header />
        <View
          style={{
            backgroundColor: isDarkMode ? Colors.black : Colors.white,
          }}>
          <TimezoneAwareComponent />
          <DateComponent />
        </View>
      </ScrollView>
    </SafeAreaView>
  );
}

const styles = StyleSheet.create({
  sectionContainer: {
    marginTop: 32,
    paddingHorizontal: 24,
  },
  sectionTitle: {
    fontSize: 24,
    fontWeight: '600',
  },
  sectionDescription: {
    marginTop: 8,
    fontSize: 18,
    fontWeight: '400',
  },
  highlight: {
    fontWeight: '700',
  },
});

export default App;
```

## Video

https://github.com/user-attachments/assets/68266155-02c2-4d3d-bd6f-60a07a0d3b6d

https://github.com/user-attachments/assets/290d114d-ed55-49dd-80e1-fc98d18bfae6


tmikov:
Thank you for submitting this PR. However I don't really understand this change. It seems it is exposing a new non-standard method `resetCache()`. Why is that desirable? It doesn't seem this method is ever called, so how does it address the problem?

lavenzg:
I can't reproduce your result locally. This is the result I'm getting:
```
TZ=America/Santiago ~/hermes_build/build/bin/hermes gh1693.js
Fri Sep 01 2023 00:00:00 GMT-0400
Sat Sep 02 2023 00:00:00 GMT-0400
Sun Sep 03 2023 01:00:00 GMT-0300
Mon Sep 04 2023 00:00:00 GMT-0300
```

We only need to reset cache if the timezone changes when hermes is running (e.g., set the TZ environment variable). Is the "hermes playground" an App that is already running before you change the timezone in step 1?

troZee:
@tmikov 
```
It seems it is exposing a new non-standard method resetCache() to JS
```

Yes correct. I forgot to mention that in the PR description, but when we call that function for every calendar event, which comes from the API all, it works and those timezone changes are reflected on the UI. I will update the PR description accordingly 

`Why is that desirable?`
In our app, we might have a situation where the timezone is changed (eg. someone is traveling from Tokyo to Sydney). When the timezone is changed, it should display a correct offset and calculate that based on the changed timezone 

`It doesn't seem this method is ever called, so how does it address the problem?`

I should have mentioned that in the PR description, that we call that method on the React Native side. 

`Also, what is the "hermes playground"?`
I meant hermes cli from https://github.com/facebook/hermes/releases/tag/v0.13.0 

@lavenzg 
I answered on all question above, but I will put more comment to your questions 

```
 Is the "hermes playground" an App that is already running before you change the timezone in step 1?
 ```
 Those changes were tested inside the RN app and we called that exposed function on the react-native side

**Comment by mdjastrzebski** [comment]
Together with @troZee  we are looking for a way to make Date work correctly in case underlying TZ changes. It seems that the caching has been introduced around 0.76 in this [PR](https://github.com/facebook/hermes/commit/59ea64c52d78ecd8e281e576fab16c208f87d4ae).

I see two ways that could be done:
1. Hermes will detect TZ change and invalidate the cache itself
2. Hermes will expose method to invalidate the cache manually.

@tmikov wdyt?

**Comment by tmikov** [comment]
So, the idea is one of:
a) to introduce a non-standard JS API to reset the TZ cache
b) introduce a native API for it
c) have Hermes correctly detect that the TZ has changed.

If we decide to go with a), it would likely be a method on HermesInternal, not on `Date`.

In your experiments, when are you calling `resetCache()`?

**Comment by tmikov** [comment]
We discussed it internally and concluded that the best way to handle this is to detect a timezone change using platform-specific APIs and then reset the cache. At the same time, we don't think hooking into platform APIs should be done by Hermes - it may even be impossible.

We propose adding a JSI method to HermesRuntime for reseting the TZ cache. The integrator (React Native) can then call this method after detecting that the TZ has changed, using whatever approach is suitable for the platform.

**Comment by troZee** [comment]
@tmikov @lavenzg 
Thank you for checking that issue 🙏 

```
Do you want to submit a PR adding it? It just need to call localTimeOffsetCache.reset().
```

Should I submit a new PR or apply changes to this one?

**Comment by tmikov** [comment]
> Should I submit a new PR or apply changes to this one?

Makes no difference either way.

**Comment by troZee** [comment]
@tmikov @lavenzg 
I have updated the code. Please let me know if I should test additional scenarios. 

I did not add any unit tests, since I am not sure if those are needed for the exposed methods to JS

**Comment by troZee** [comment]
@tmikov @lavenzg 

We noticed the following comment https://github.com/facebook/hermes/pull/1693/files#diff-cd87104dd09a2271bbedfe3f5aab799f6efa6c1583c8dab9f1abdc1a57297b4cR202 I changed the implementation based on this comment, so instead of exposing the `reset` method, I created a new method `needsToReset()` for changing `needsToReset_` flag to make this function thread-safe

**Comment by tmikov** [comment]
Hi, I am sorry, as I said earlier, this is what we are proposing:

> We propose adding a JSI method to HermesRuntime for reseting the TZ cache. The integrator (React Native) can then call this method after detecting that the TZ has changed, using whatever approach is suitable for the platform.

**Comment by troZee** [comment]
@tmikov reverted

**Comment by troZee** [comment]
@tmikov @lavenzg 
I added more unit tests for that PR. I run the below command to check if all unit tests are passing 

```
hermes git:(fix/1121-tz-change) ✗ cmake --build ../build --target check-hermes
[0/1] Running the Hermes regression tests
Testing Time: 38.50s
  Expected Passes    : 2238
  Unsupported Tests: 44
```

Could you also run the CI for this PR? I would like to be sure everything has been addressed in this PR
Screenshot 2025-05-08 at 22 54 43

Please let me know if something is still missing in this PR

**Comment by troZee** [comment]
@tmikov @lavenzg 
I moved this implementation from HermesInternal to HermesRuntime. Please let me know if it is implemented correctly. 

I need help with the unit tests here https://github.com/facebook/hermes/pull/1693/files#diff-dac96edcb6b8bf138649480205106b295f2fbc8768bdd85541b8624d070527d6R107 . For some reason, there is a linker error, and those files are in the CMAKE.txt file. I am unsure why those files are not linked correctly when running unit tests. 

If this comment part is not needed, then I can remove that

**Comment by troZee** [comment]
@tsaichien 
In this commit https://github.com/facebook/hermes/commit/2bba5f049b6b9c3e959f92391cb0f682c70d668b
the method getVMRuntimeUnsafe changed the returned value from`::hermes::vm::Runtime` to `void`:

```
- ::hermes::vm::Runtime *getVMRuntimeUnsafe() const override;
+ void *getVMRuntimeUnsafe() const override;
 ```
 
 Could you advise me on what I should do to resolve the conflicts for this PR?

troZee:
@tmikov @lavenzg 
The conflicts are resolved. The PR is ready for review

troZee:
@tmikov @lavenzg
I've addressed all the points raised in the initial review and believe it's ready for final approval.

Could you please provide an estimated timeline for when I can expect the review to be completed? This will help with planning.

Thanks!

API/hermes/hermes.h
```diff
@@ -179,6 +179,9 @@ class HERMES_EXPORT IHermes : public jsi::ICast {
   /// be used by third parties.
   virtual sampling_profiler::Profile dumpSampledTraceToProfile() = 0;
 
+  /// Reset timezone cache
```
tmikov, line 182: Can you please add a much more detailed comment explaining what purpose this call serves, uses cases when it should be invoked, etc.

troZee:
@tmikov @lavenzg 
Is there a possibility to merge this PR this week or next week? I checked the pipeline, and on the main branch, there are the same errors as in this PR, hence it is ready to be merged.

tmikov:
Landing changes to JSI is a complex and time consuming process, unfortunately. It will take us some time.

tmikov:
Merged in https://github.com/facebook/hermes/commit/1b4de35e4e42c807393b35f26892a4e531255b39

troZee:
@tmikov or @lavenzg 
Do you know how to call it inside a turbomodule? I tried something like this
```
    RCTCxxBridge *cxxBridge = (RCTCxxBridge *)self.bridge;
      facebook::jsi::Runtime *jsiRuntime = (facebook::jsi::Runtime *)cxxBridge.runtime;
     
facebook::hermes::HermesRuntime *hermesRuntime =
      dynamic_cast<facebook::hermes::HermesRuntime*>(jsiRuntime);
```
      
 but hermesRuntime is null after the casting

tmikov:
@troZee no idea. Turbo Modules are an RN invention - we are not familiar with them... 

Note you shouldn't use `dynamic_cast` anymore. You should use `jsi::castInterface()` (though I don't think it will help in this case).

PR #1701, opened 2025-05-07
Fixes `flow-api-translator` default type parameter generation

nickhudkins:
## Summary
While working on Metro, I noticed that many  of the typescript type definitions were defined with `<void>` as type parameters. This is *incorrect*.

See Discussion Here:
https://github.com/facebook/metro/issues/1491#issuecomment-2824661024

## Test Plan
Added tests to the FlowDef -> TSDef tests to ensure empty type params (which imply default type params) properly omit type parameters when generating TSDef.

### DIFF
--- a/tools/hermes-parser/js/flow-api-translator/__tests__/flowDefToTSDef/fixtures/type-parameters/default/translate-result.shot
+++ b/tools/hermes-parser/js/flow-api-translator/__tests__/flowDefToTSDef/fixtures/type-parameters/default/translate-result.shot
@@ -12,5 +12,6 @@ exports[`flowDefToTSDef type-parameters/default 1`] = `
  */
 
 type T1<A = B> = A;
+type T2 = T1;
 "
 `;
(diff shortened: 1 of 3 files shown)


tools/hermes-parser/js/flow-api-translator/__tests__/flowDefToTSDef/fixtures/type-parameters/default/translate-result.shot
```diff
@@ -12,5 +12,6 @@ exports[`flowDefToTSDef type-parameters/default 1`] = `
  */
 
 type T1<A = B> = A;
+type T2 = T1;
```
nickhudkins, line 15: previously, this would generate `type T2 = T1<>` (invalid TS)

pieterv:
The changes look fine but can you split out the build changes and the translator changes into separate PR's?

pieterv:
A single PR for both those should be fine.

nickhudkins:
Build Fixes:
https://github.com/facebook/hermes/pull/1702

(This PR has been updated to drop the build changes)

pieterv:
@nickhudkins I landed this change here: https://github.com/facebook/hermes/commit/cb33e4bbfd9a215eb96df711e68e3fea76cd9c0e

I had to change the code a bit as your approach was creating an invalid AST structure and erroring Flow. Thanks for the fix!

PR #1762, opened 2025-08-20
Fix publish workflow and optimize `build_android`

coado:
Summary:

This diff fixes publish workflow and optimizes `build_android` by building hermesc once in the main `build.gradle.kts` and linking `TestUtilsPlugin` with built `ImportHermesc` instead of relying on a separate step that prepared hermesc before the gradle build.

Differential Revision: D80538437


cortinico (approved):

PR #1765, opened 2025-08-22
Publishing Static Hermes Artifacts to Maven

coado:
Summary: This diff configures publishing Static Hermes artifacts to Maven to `com.facebook.hermes-v2` coordinates. The logic is the same as in Hermes publishing workflow (realeses, commitlies).

Differential Revision: D80450523


cortinico (changes requested):

PR #1771, opened 2025-08-26
Publish HermesC from static_h to npm

coado:
Summary:
This diff provides a setup for publishing hermesc artifacts from CI.

## Motivation

Build and publish hermesc artifacts on npm from the Hermes repo to be referenced and used in React Native projects.

Differential Revision: D80798577


cortinico (approved):

PR #1783, opened 2025-09-15
Set gha-npm-token before publish workflow

coado:
Summary: Configures `gha-npm-token` auth token in `.npmrc` before the publish npm workflow and uploads npm logs after that. Removes the `otp` flag which is not used in Hermes repo.

Differential Revision: D82440778


cortinico (approved):

PR #1792, opened 2025-09-18
Set hermes release version

coado:
Summary: Configures build to pass `-DHERMES_RELEASE_VERSION=for RN ${hermesReleaseVersion}` where `hermesReleaseVersion` is a version read from the `hermes-compiler/package.json`.

Differential Revision: D82733502


cortinico (approved):

PR #1819, opened 2025-10-28
Bump Macos runners to 15

cipolleschi:
Summary:
GH asked us to bump the macos runners to 15.

Unfortunately, with macos-15, Apple decided that Apple Clang was not supporting the `x86_64-arm64-apple-ios14.0-macabi` compiler flag anymore.

So we now have to build the two Catalyst architecture separately and then merge them together calling `lipo` manually

Differential Revision: D85662422


cipolleschi:
This pr has been committed

PR #1821, opened 2025-10-29
Reapply "[SH] Sync JSI changes"

cipolleschi:
## Summary
Next week we are going to cut a new version of React Native. 
during these months, JSI.cpp has been modified, so we need to sync HermesV1's `jsi.cpp` with the React Native's `jsi.cpp`

This PR reapply the commits 358bbab27 and 129351a09

## Test Plan
GHA


cipolleschi:
Yeah, they updated the clang formatter and run it against the whole repo. We probably don't need those commits here.

PR #1834, opened 2025-11-04
Cherry Pick: Speed up fixupEnvironmentIDs in BytecodeGenerator.cpp

coado:
Summary:
`fixupEnvironmentIDs()` calls `DominanceInfo::properlyDominates() for
almost every instruction in a function. The cost ends up being
significant. A 3000-line function of the following form:
```
function sum() {
  let a = 0;
  a = a + 1 + 2 + 3 + 4 + 5 + 6 + 7 + 8 + 9 + 10;
  ...
  a = a + 1 + 2 + 3 + 4 + 5 + 6 + 7 + 8 + 9 + 10;

  return a;
}
```
took 50ms to compile with Hermes, but takes 10s to compile with Static
Hermes, on an M4 MBP.

Fix it by using a DFS dominator tree visitor with a ScopedHashTable of
the "active" environments. Employ additional single element cache of the
last successfully matched dominating environment within a BB (idea by
fbmal7).

### DIFF
--- a/lib/BCGen/HBC/BytecodeGenerator.cpp
+++ b/lib/BCGen/HBC/BytecodeGenerator.cpp
@@ -8,6 +8,7 @@
 #include "BytecodeGenerator.h"
 
 #include "LoweringPipelines.h"
+#include "hermes/ADT/ScopedHashTable.h"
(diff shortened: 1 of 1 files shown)


coado:
This was merged in 40dd6c6

PR #1854, opened 2025-12-04
feat: disable dynamic code execution for function constructors

kewde:
## Summary

Disables dynamic code execution through function constructors. 

Q: is this acceptable?
Q: behind enableEval flag as well?

## Test Plan

- [ ] TODO if concept is acknowledged


avp:
All these code paths do go through `evalInEnvironment` eventually, which is the one place where we have the check for the `enableEval` flag. Things appear to be working as expected in the CLI when I pass `-enable-eval=false`.

Can you please explain why this PR is needed?

kewde:
@avp Ahah! You're right, we totally don't need this, closing.

(context: I've been carrying over a set of patches to hermes written by other people, I'm looking through them to see if there's anything worth upstreaming. Most patches are towards getting reproducible builds, but this one was a security "improvement" which turned out to be redundant)

PR #1858, opened 2025-12-11
Migrate TextEncoder from VM APIs to JSI APIs

tmikov:
Summary:
Move TextEncoder implementation from lib/VM/JSLib/ (using Hermes private
VM APIs) to API/hermes/extensions/ (using portable JSI APIs). This is
the first step toward enabling "extensions" - JS built-ins that can be
implemented using JSI rather than internal VM APIs.

The new implementation:
- Creates a JS constructor via evaluateJavaScript() since JSI host
  functions cannot be used as constructors
- Uses jsi::NativeState on regular JS objects for instance validation
- Uses jsi::String::getStringData() for efficient ASCII optimization
- Installs TextEncoder globally in makeHermesRuntime()

Note: This is an intermediate state. Evaluating JS source at runtime is
not acceptable for production - it has performance overhead and prevents
the code from being precompiled. A proper solution would precompile the
JS setup code and include it in the binary. Additionally, extensions are
currently only registered when JSI is initialized via makeHermesRuntime(),
which is not always the case (e.g., the main Hermes CLI).

Differential Revision: D88606329


kewdex:
edit: marking this offtopic myself since i misconstrued what was being said.

> This is an intermediate state. Evaluating JS source at runtime is
not acceptable for production - it has performance overhead and prevents
the code from being precompiled. A proper solution would precompile the
JS setup code and include it in the binary.

Interesting,  if we were to measure performance using JS at runtime, would we have any of that overhead?

At the moment, our benchmark evaluates the performance at runtime inside the JS. We always assumed the following things:
1. Hermes (CLI) compiles the JS to HBC
2. our single bundle of JS gets (lazy) compiled to HBC in one go, no partial lazy compilation
3. we measure at runtime directly in JS

> Additionally, extensions are
currently only registered when JSI is initialized via makeHermesRuntime(),
which is not always the case (e.g., the main Hermes CLI).

Is this something desirable to add to the main Hermes CLI? It would make testing these extensions a lot easier. I've seen the JSI test runner example, I suppose that could be used as well.

tmikov:
@kewdex I am sorry, I don't understand your question.

kewdex:
@tmikov 

I'm sorry, I misconstrued what you were saying.

If someone wanted to benchmark an extension, say `TextEncoder`, how should they approach it in this new extension system?

> Additionally, extensions are
currently only registered when JSI is initialized via makeHermesRuntime(),
which is not always the case (e.g., the main Hermes CLI).

Our current benchmarking tool relies on the Hermes CLI - so I'm curious how we can adapt it to make it work with the "built-in" extensions.

edit: happy to take this into a GitHub Discussions, this PR may not be the right place for it.

tmikov:
@kewdex you don't need to do anything special. It just works.

Issue #1883, opened 2026-01-24
Hermes 0.14 artifact breaks Mac Catalyst build


robertying:
## Bug Description

There are two issues with macOS (Catalyst) build due to the hermes artifact.

1. bundle format is ambiguous (could be app or framework) when building against Catalyst
2. found architecture 'arm64', required architecture 'x86_64' when building against Catalyst with universal (arm64+x86_64)

### 1

There was a similar issue a while ago: https://github.com/facebook/react-native/issues/46213 and it was caused by GitHub Action's handling of symlinks.

The issues have come back and it might be because the change of moving hermes build from the react-native repo to the hermes repo: https://github.com/facebook/react-native/pull/53901

[The previous fix](https://github.com/facebook/react-native/pull/46301) added manual zip in RN repo's CI steps. However, in the hermes repo's CI configurations, it is currently missing.

Digging deeper, we can see the unzipped artifact's directory structure:

```
drwxr-xr-x@ 5 rying  staff       160 Jan 20 02:49 .
drwxr-xr-x@ 3 rying  staff        96 Jan 20 02:49 ..
drwxr-xr-x@ 3 rying  staff        96 Jan 20 02:49 Resources -> Versions/Current/Resources
drwxr-xr-x@ 4 rying  staff       128 Jan 20 02:49 Versions
-rw-r--r--@ 1 rying  staff  13564883 Jan 20 02:49 hermesvm
```

The correct structure should look like this instead:

```
drwxr-xr-x@ 5 rying  staff       160 Jan 20 02:49 .
drwxr-xr-x@ 3 rying  staff        96 Jan 20 02:49 ..
drwxr-xr-x@ 3 rying  staff        96 Jan 20 02:49 Resources -> Versions/Current/Resources
drwxr-xr-x@ 4 rying  staff       128 Jan 20 02:49 Versions
-rw-r--r--@ 1 rying  staff  13564883 Jan 20 02:49 hermesvm -> Versions/Current/hermesvm
```

### 2

It also seems that hermes was not built for x86_64 correctly. The bundled hermesvm file doesn't contain x86_64 binary:

> Non-fat file: ~/Downloads/hermes-ios-0.14.0-debug/destroot/Library/Frameworks/universal/hermesvm.xcframework/ios-arm64_x86_64-maccatalyst/hermesvm.framework/hermesvm is architecture: arm64

---

- [ ] ~I have run `gradle clean` and confirmed this bug does not occur with JSC~ N/A
[...]

PR #1886, opened 2026-01-27
[Catalyst][main] Fix target triple for multi-arch builds

cipolleschi:
## Summary
Fixes: #1883

When both `-target` and `-arch` flags are passed to Apple clang, the `-arch` flag overrides only the architecture portion of the target triple while preserving the OS and environment.

Use a single `-target arm64-apple-ios$VERSION-macabi` flag, and let CMake's multi-arch handling add `-arch x86_64` and `-arch arm64` flags which correctly produce the right target triple for each architecture.

Example:
```
clang -target arm64-apple-ios15.1-macabi -arch x86_64 ...
→ Effective target: x86_64-apple-ios15.1-macabi
```

## Test Plan
Tested via GitHub Actions. The resulting Hermes binary is properly structured and contains all required architectures.


cipolleschi:
this has been committed already.

PR #1902, opened 2026-02-13
Fix typo in error message when no case is matched in a match statement

NicolasRoelandt:
## Summary

I saw the comment about avoiding a pull request for a single typo, but this exception message is user-facing text, and is likely used as an assertion in tests


gkz:
Thanks for the report.
I plan on making different changes to this error message in the future.
Because of the process of rolling out such changes, I prefer to only make such a change once rather than twice.
If I never end up making the change I have in mind, I will come back to this and merge this.

PR #1914, opened 2026-02-23
Fix Iterator.prototype.reduce callable check order

JakeChampion:
## Summary

The callable check was wrapped in IteratorClose with a fake iterated record created before GetIteratorDirect. This caused .return() to be called on the generator, closing it prematurely. Per the spec, the callable check is a plain throw before GetIteratorDirect, since the iterator hasn't been obtained yet.

## Test Plan

Added test and also ran Test262 test `test262/test/built-ins/Iterator/prototype/reduce/non-callable-reducer.js`

### DIFF
--- a/lib/InternalJavaScript/15-IteratorPrototypeReduce.js
+++ b/lib/InternalJavaScript/15-IteratorPrototypeReduce.js
@@ -39,16 +39,11 @@ var IteratorPrototypeReduce = function reduce(reducer) {
 		throw new TypeError('`this` value must be an Object'); // step 2
 	}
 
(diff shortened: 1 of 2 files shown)


tsaichien:
Hi! Thanks for bringing this to our attention. 

The polyfill follows the [spec](https://tc39.es/ecma262/#sec-iterator.prototype.reduce), where `IteratorClose` is indeed called. The changes you are describing here is following the proposal, which is now outdated. 

The reason why our CI tests are failing is we didn't update our test262 version, so the `non-callable-reducer.js` test reflects the outdated behavior. We'll fix this on our end.

Thanks again for bringing this up!

PR #1922, opened 2026-03-02
(DRAFT) Fix JSProxy::ownPropertyKeys index key comparison

vmoroz:
## Summary

`JSProxy::ownPropertyKeys` throws a spurious `TypeError` when the proxy's
`ownKeys` trap returns string-valued array indices (e.g. `"0"`, `"1"`) and
the target object is non-extensible, sealed, or frozen.

## The Bug

`JSObject::getOwnPropertyKeys` stores indexed properties as **numbers** in the
result array (via `HermesValue::encodeTrustedNumberValue`). However, the
`ownKeys` trap result processes values from the trap's return array as-is,
keeping string keys like `"0"` as strings.

Later, the ES spec invariant checks (steps 19 and 21 of
`[[OwnPropertyKeys]]`) compare `targetKeys` entries against `trapResult`
entries using `isSameValue`. Since `isSameValue` checks the HermesValue tag
first, `isSameValue(Number(0), String("0"))` returns `false` — the types
don't match — even though they represent the same property key.

This causes the invariant check to incorrectly throw:
- `"ownKeys target key is non-configurable but not present in trap result"`
  (for sealed/frozen targets)
- `"ownKeys target is non-extensible but key is missing from trap result"`
  (for non-extensible targets)

## The Fix

In `lib/VM/JSProxy.cpp`, inside the `createListFromArrayLike_RJS` callback
in `ownPropertyKeys`, convert string values that are valid array indices to
their numeric representation before storing them in `trapResult`. This uses
`toArrayIndex()` to detect numeric strings and
`HermesValue::encodeTrustedNumberValue()` to convert, matching the format
used by `getOwnPropertyKeys`.

```cpp
if (valHandle->isString()) {
  Handle<StringPrimitive> str =
      Handle<StringPrimitive>::vmcast(valHandle);
  OptValue<uint32_t> strAsIndexOpt = toArrayIndex(*str);
  // Convert index keys to numbers to match the format
  // returned by getOwnPropertyKeys.
  if (strAsIndexOpt) {
    HermesValue strAsIndexValue =
        HermesValue::encodeTrustedNumberValue(
            static_cast<double>(strAsIndexOpt.getValue()));
    return JSArray::setElementAt(
        trapResult,
        runtime,
        index,
        runtime.makeHandle(strAsIndexValue));
  }
}
```

## Reproduction

Any of these patterns trigger the bug without the fix:

```javascript
// Non-extensible object with numeric-string keys
var obj = {"0": "a", "1": "b"};
Object.preventExtensions(obj);
var proxy = new Proxy(obj, {
  ownKeys(target) { return ["0", "1"]; }
});
Object.getOwnPropertyNames(proxy); // TypeError!

// Non-extensible array
var arr = [10, 20, 30];
Object.preventExtensions(arr);
var proxy2 = new Proxy(arr, {
  ownKeys(target) { return ["0", "1", "2", "length"]; }
});
Object.getOwnPropertyNames(proxy2); // TypeError!

// Reflect.ownKeys on a non-extensible array
var arr2 = ["a", "b"];
Object.preventExtensions(arr2);
var proxy3 = new Proxy(arr2, {
  ownKeys(target) { return Reflect.ownKeys(target); }
});
Reflect.ownKeys(proxy3); // TypeError!
```

Note: the bug does **not** manifest for extensible targets because the
`isSameValue` invariant checks (steps 19 and 21) are only reached for
non-extensible targets.

## Test

Added `test/hermes/regress-proxy-ownkeys-index.js` covering five cases:
1. Non-extensible plain object with numeric-string keys
2. Non-extensible array
3. Sealed array (non-configurable + non-extensible)
4. Frozen object with mixed numeric and named keys
5. `Reflect.ownKeys` on a non-extensible array

All five fail with `TypeError` without the fix and pass with it.

### DIFF
--- a/lib/VM/JSProxy.cpp
+++ b/lib/VM/JSProxy.cpp
@@ -1411,6 +1411,23 @@ CallResult<PseudoHandle<JSArray>> JSProxy::ownPropertyKeys(
                         JSMap::insert(lv.map, runtime, valHandle, valHandle) ==
                         ExecutionStatus::EXCEPTION))
                   return ExecutionStatus::EXCEPTION;
+                if (valHandle->isString()) {
+                  Handle<StringPrimitive> str =
+                      Handle<StringPrimitive>::vmcast(valHandle);
+                  OptValue<uint32_t> strAsIndexOpt = toArrayIndex(*str);
+                  // Convert index keys to numbers to match the format
+                  // returned by getOwnPropertyKeys.
+                  if (strAsIndexOpt) {
+                    HermesValue strAsIndexValue =
+                        HermesValue::encodeTrustedNumberValue(
+                            static_cast<double>(strAsIndexOpt.getValue()));
+                    return JSArray::setElementAt(
+                        trapResult,
+                        runtime,
+                        index,
+                        runtime.makeHandle(strAsIndexValue));
+                  }
+                }
                 return JSArray::setElementAt(
                     trapResult, runtime, index, valHandle);
               }) == ExecutionStatus::EXCEPTION)) {
--- a/test/hermes/regress-proxy-ownkeys-index.js
+++ b/test/hermes/regress-proxy-ownkeys-index.js
@@ -1,0 +1,94 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// RUN: %hermes -Xes6-proxy -non-strict -O -target=HBC %s | %FileCheck --match-full-lines %s
+
+// Regression test: JSProxy::ownPropertyKeys must convert string-valued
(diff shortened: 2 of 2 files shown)


fbmal7:
There seems to be an error in the 262 testsuite: `test\built-ins\Object\assign\strings-and-symbol-order-proxy.js`

vmoroz:
> There seems to be an error in the 262 testsuite: `test\built-ins\Object\assign\strings-and-symbol-order-proxy.js`

Yep, I see it. It sounds that the fix must be a bit more complex. I will switch the PR to DRAFT mode and look for a new approach.
The issue that the PR is addressing is real. You can run the new tests from this PR without the fix and see them to pass in V8 (or any browser) and fail in Hermes.

vmoroz:
Closing the PR since it causes 262 tests to fail. I will submit another PR if I have better solution.

PR #1925, opened 2026-03-03
Make Math.sumPrecise reject multiple arguments per spec

coderyi:
## Summary
Make Math.sumPrecise throw a TypeError when called with zero or more than one argument, matching the TC39 proposal and test262 expectations.


tsaichien:
Hi! I believe the test comment is wrong and the current behavior is fine. The ECMA spec doesn't throw error if more than 1 argument is passed into `sumPrecise` AFAICT. Please let me know if I missed something in the spec. We will update the comment on our side. Sorry for the confusion.

PR #1651, opened 2025-03-26
chore: fix repository url not displayed on npm

JounQin:
## Summary

Takes [`hermes-eslint`](https://www.[npmjs.com/package/hermes-eslint](https://www.npmjs.com/package/hermes-eslint)) as an example

image

The git protocol repository url won't be displayed on npm now.

PR #1926, opened 2026-03-04
Add ESLint v10 support to hermes-eslint

JasonWeinzierl:
## Summary

Add support to `hermes-eslint` for ESLint v10.

1. Added support for `ScopeManager.addGlobals()`.
    - Since this scope manager is heavily based on typescript-eslint's scope manager, I followed their changes in https://github.com/typescript-eslint/typescript-eslint/pull/11914 to ensure it was correct.
    - Also followed their unit test changes.
2. Added `eslint` to `hermes-eslint`'s peer dependencies as optional, for documenting supported versions.
    - Everything works as far back as ESLint v4.  I created a demo repository that tests compatibility: https://github.com/JasonWeinzierl/hermes-eslint-compatibility-test
    - Also moved `esrecurse` to `devDependencies` since it's not required by the published code.
3. Migrated the internal eslint config to the flat config instead of eslintrc.
    - Several incompatible plugins were swapped out for their updated forks (ft-flow and eslint-comments).
    - Upgraded the internal eslint version to v9.  Cannot go to v10 because some plugins use APIs not available in v10 (e.g. `context.getSourceCode()`).  For example, https://github.com/flow-typed/eslint-plugin-ft-flow/pull/71 recursively needs `hermes-eslint` to be upgraded first.
    - Installed `@eslint/js` as a defensive measure: eslint v10 will require this to be installed.
4. Updated README of `hermes-eslint` with flat config example.

Note: it would be nice to see https://github.com/facebook/hermes/pull/1651 merged, since these packages published on npm do not display links to this repository, so it took me a minute to find this.

## Test Plan

Ran the following commands:

```sh
# (first I set up the repository and built the wasm, then I did the following)
cd tools/hermes-parser/js
yarn install
yarn build ../../../build-wasm/bin/hermes-parser-wasm.js ../../../include
yarn test
yarn lint
yarn format
```

Build, test, and lint passes.

Demonstration of peer dependency compatibility: https://github.com/JasonWeinzierl/hermes-eslint-compatibility-test


gkz:
Hi @JasonWeinzierl, thanks for your work! We will take a look at this, though we will need to confirm that nothing breaks internally.

JasonWeinzierl:
Rebased onto `static_h`

Edit: rebased again

gkz:
> Several incompatible plugins were swapped out for their updated forks (ft-flow and eslint-comments)

Is it possible to just update `ft-flow` plugin instead of forking it? If there needs to be some PR approved I can look into getting that done.

> Migrated the internal eslint config to the flat config instead of eslintrc

To be clear, a flat config is not required now with these updates, eslintrc is still reported? It is just also supported? Or is this a backwards incompatible change

JasonWeinzierl:
> > Several incompatible plugins were swapped out for their updated forks (ft-flow and eslint-comments)
> 
> Is it possible to just update `ft-flow` plugin instead of forking it? If there needs to be some PR approved I can look into getting that done.

I might be misunderstanding, but yes this PR is pulling in the most updated version of `ft-flow`.  By "updated forks" I just meant that `ft-flow` itself is a fork of the old `eslint-plugin-flowtype`.  `ft-flow` does need changes to be fully compatible with ESLint v9 and v10, which I've submitted here https://github.com/flow-typed/eslint-plugin-ft-flow/pull/73 , but those aren't strictly necessary for this PR.

> > Migrated the internal eslint config to the flat config instead of eslintrc
> 
> To be clear, a flat config is not required now with these updates, eslintrc is still reported? It is just also supported? Or is this a backwards incompatible change

Correct, both formats are supported.  There shouldn't be any incompatible changes here; the "hermes-eslint-compatibility-test" repo I linked demonstrates that the parser is compatible all the way back to ESLint v4.

gkz:
Thanks @JasonWeinzierl for your contributions!
Some small changes were made to the PR during review - if you grab the latest make sure it still works for your use-case

PR #1932, opened 2026-03-11
Fix freeze/seal on lazy function objects

Kenaz123:
## Summary

Fix `Object.freeze()` / `Object.seal()` on lazy function objects.

Before this change, `JSObject::freeze()` and `JSObject::seal()` could mark a lazy function object as frozen/sealed before its lazy properties were materialized. Later, lazy initialization would still add properties like `length` and `name`, leaving them configurable on an object that already reported itself as frozen.

The problematic code path was that `seal()` / `freeze()` did not initialize lazy objects first:

```cpp
ExecutionStatus JSObject::seal(Handle<JSObject> selfHandle, Runtime &runtime) {
  CallResult<bool> statusRes = JSObject::preventExtensions(
      selfHandle, runtime, PropOpFlags().plusThrowOnError());
  ...
}

ExecutionStatus JSObject::freeze(
    Handle<JSObject> selfHandle,
    Runtime &runtime) {
  CallResult<bool> statusRes = JSObject::preventExtensions(
      selfHandle, runtime, PropOpFlags().plusThrowOnError());
  ...
}
```

This change fixes that by forcing lazy initialization before the object is marked non-extensible and before the hidden class flags are updated.

[poc.js](https://github.com/user-attachments/files/25900056/poc.js)

Original PoC reproduction:

```bash
./build/bin/hermes ./poc.js
```

Original output before the fix:

```text
isFrozen: true
length.configurable: true
BUG CONFIRMED: delete foo.length succeeded on a frozen function object.
  foo.length after delete: 0
  Object.isFrozen(foo): true
```

## Test Plan

```bash
cmake --build ./build --target hermes
./build/bin/hermes test/hermes/regress-lazy-freeze-seal-function.js
./build/bin/hermes -emit-binary -out /tmp/regress-lazy-freeze-seal-function.hbc \
  test/hermes/regress-lazy-freeze-seal-function.js
./build/bin/hermes /tmp/regress-lazy-freeze-seal-function.hbc
cmake --build ./build --target check-hermes
```

`check-hermes` result:

```text
Testing Time: 25.44s
  Expected Passes    : 2761
  Expected Failures  : 5
  Unsupported Tests  : 159
```

### DIFF
--- a/lib/VM/JSObject.cpp
+++ b/lib/VM/JSObject.cpp
@@ -35,12 +35,11 @@ const ObjectVTable JSObject::vt{
         nullptr
 #ifdef HERMES_MEMORY_INSTRUMENTATION
         ,
-        VTable::HeapSnapshotMetadata{
-            HeapSnapshot::NodeType::Object,
-            JSObject::_snapshotNameImpl,
-            JSObject::_snapshotAddEdgesImpl,
-            nullptr,
-            JSObject::_snapshotAddLocationsImpl}
+        VTable::HeapSnapshotMetadata {
+          HeapSnapshot::NodeType::Object, JSObject::_snapshotNameImpl,
+              JSObject::_snapshotAddEdgesImpl, nullptr,
+              JSObject::_snapshotAddLocationsImpl
+        }
 #endif
         ),
     JSObject::_getOwnIndexedRangeImpl,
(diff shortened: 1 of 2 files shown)


tmikov:
Hi, thanks for this. We do not accept PRs for `main`, which has been frozen for a couple of years. Can you please resubmit against the `static_h` branch, where all current development is going on.

PR #1935, opened 2026-03-12
Support @@dispose/@@asyncDispose and fix @@iterator/@@asyncIterator in object/interface types (TS declarations)

zoontek:
## Summary

The flow-api-translator is not translating `@@iterator` / `@@asyncIterator` in object types and interfaces (only class members are handled). This caused raw `@@iterator` syntax to appear in `react-native` `0.85-rc.0` TypeScript definitions, where an `@@iterator` is present in the generated `Read` TS interface.

This PR extends the `DeclareClass` and `ObjectTypeProperty` transforms to handle all well-known symbols (`@@iterator`, `@@asyncIterator`, `@@dispose`, `@@asyncDispose`), converting them to their `[Symbol.*]` TypeScript equivalents.

## Test Plan

Added snapshot tests for all three contexts (class, interface, object type) covering both iterator and dispose symbols. All tests pass:

```sh
NODE_OPTIONS="--experimental-vm-modules" npx jest flow-api-translator/__tests__/flowDefToTSDef -t "iterator|dispose"
```

### DIFF
--- a/tools/hermes-parser/js/flow-api-translator/__tests__/flowDefToTSDef/fixtures/class/members/method/dispose/spec.js
+++ b/tools/hermes-parser/js/flow-api-translator/__tests__/flowDefToTSDef/fixtures/class/members/method/dispose/spec.js
@@ -1,0 +1,14 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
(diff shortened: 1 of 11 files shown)


gkz:
0.34.0 with this change is out now

PR #1946, opened 2026-03-24
Guard gtest build on HERMES_ENABLE_TEST_SUITE

CedricGuillemet:
## Summary

The gtest and gtest_main targets in external/llvh/utils/unittest were built unconditionally, even when HERMES_ENABLE_TEST_SUITE was set to OFF. This adds a conditional guard so the unittest subdirectory is only included when the test suite is enabled, matching the existing guard on the unittests/ directory in the top-level CMakeLists.txt.

### DIFF
--- a/external/llvh/CMakeLists.txt
+++ b/external/llvh/CMakeLists.txt
@@ -32,7 +32,9 @@ add_subdirectory(utils/FileCheck)
 add_subdirectory(utils/TableGen)
 add_subdirectory(utils/not)
 add_subdirectory(utils/count)
-add_subdirectory(utils/unittest)
+if(HERMES_ENABLE_TEST_SUITE)
+  add_subdirectory(utils/unittest)
+endif()
 
 set(LLVH_LIT_ARGS "-sv")
 set(LLVH_LIT_TEST_PARAMS


tmikov:
Thanks for the contribution! However, we'd prefer not to merge this because:

1. The build time savings are negligible — gtest is just a couple of source files.
2. This introduces a dependency from `external/llvh/` (a vendored LLVM library) on `HERMES_ENABLE_TEST_SUITE`, which is a Hermes-specific CMake option. External libraries shouldn't have knowledge of Hermes configuration.

If there were a stronger motivation, the right approach would be to guard this from the Hermes side rather than modifying the external library.

CedricGuillemet:
Hi @tmikov 
I want to integrate Hermes in my App but LLVH GoogleTest is conflicting with my version. What would be the good practice to disable GTest in that case?

Thanks

tmikov:
Thanks for reporting this issue! We agree that the `gtest`/`gtest_main` target names are too generic and can conflict when Hermes is embedded via `add_subdirectory()`.

We've opened #1947 to rename them to `llvh-gtest`/`llvh-gtest_main`, which avoids the conflict without needing to disable tests. This should fix your use case while keeping the test suite buildable in embedded configurations.

tmikov:
https://github.com/facebook/hermes/pull/1947 landed. Can you try now?

CedricGuillemet:
It works for me! Thank you!

PR #1958, opened 2026-03-26
fix: include macOS slice in universal hermes xcframework

Saadnajmi:
## Summary

Mirror of https://github.com/facebook/react-native/pull/56235

Currently, Hermes is compiled into 2 xcframeworks: a "macosx" one with just the macOS slice, and a "universal" one for the rest of the Apple platforms (iOS, visionOS, Mac Catalyst, tvOS, etc). This means React Native macOS needs extra diffs and patches to use the right Hermes framework. However, xcframeworks are multiplatform — there's no reason we can't just add the macOS slice to the "universal" one, making it truly universal.

Including the macOS slice in the universal xcframework allows react-native-macos to use the same prebuilt Hermes artifacts as react-native for iOS, visionOS, and tvOS — without platform-specific patching.

### Changes

- Consolidate the three separate Hermes Apple build scripts (`build-apple-framework.sh`, `build-ios-framework.sh`, `build-mac-framework.sh`) into a single `build-apple-framework.sh` that builds all platforms — including macOS — into one universal `hermesvm.xcframework`
- Update the `hermes-engine` podspec so the Pre-built subspec uses the universal xcframework for macOS (matching iOS, tvOS, and visionOS)
- Update CI workflows to use `build-apple-framework.sh` uniformly for all slices (no more macOS special-casing)
- Remove standalone macOS framework validation from hermes-utils.js

Simpler variants of this change targeting `main` and `250829098.0.0-stable` just add `"macosx"` to the existing build scripts without the full consolidation, since those are legacy/release branches.

## Test plan

CI should pass. The `build_apple_framework "macosx"` codepath is already exercised by the existing `build-mac-framework.sh` — this PR just routes it through the same path as every other platform.


utils/build-apple-framework.sh
```diff
@@ -62,6 +72,35 @@ function get_release_version {
...
+      echo "arm64"
+    elif [[ $1 == "iphonesimulator" || $1 == "xrsimulator" ]]; then
+      echo "x86_64;arm64"
+    elif [[ $1 == "appletvos" ]]; then
+      echo "arm64"
+    elif [[ $1 == "appletvsimulator" ]]; then
+      echo "x86_64;arm64"
+    elif [[ $1 == "catalyst" || $1 == "macosx" ]]; then
+      echo "x86_64;arm64"
+    else
```
cipolleschi: this can be simplified in two branches

cipolleschi (approved):
I really like the cleanup!

cipolleschi:
CI is failing to build hermesc:
```
  clang: error: no such file or directory: 'Missing'
  clang: error: no such file or directory: 'MAC_DEPLOYMENT_TARGET'
  clang: error: no such file or directory: 'environment'
  clang: error: no such file or directory: 'variable'
  clang: error: invalid version number in '-mmacosx-version-min=error:'
  clang: error: invalid version number in '-mmacosx-version-min=error:'
```
we might have broken some file refs...

cipolleschi (changes requested):
CI needs to be fixed.

cipolleschi (approved):

cortinico (approved):

PR #1961, opened 2026-03-27
Fix Number.prototype.toFixed comment for fraction digit range

cuiweixie:
## Summary
Align the inline comment above `numberPrototypeToFixed` with the actual validation: fraction digits are constrained to 0–100 (per spec step 3 and the RangeError message), not 0–20.

## Test plan
- Comment-only change; no runtime behavior change.


tmikov:
I am sorry, we no longer accept spelling, typo and grammar fixes.

PR #1970, opened 2026-04-03
fix: include macOS slice in universal Hermes xcframework

Saadnajmi:
## Summary

Mirror of https://github.com/facebook/hermes/pull/1958 (targeting `static_h`), adapted for the legacy `main` branch.

Unlike the `static_h` PR which consolidates all Apple build scripts and removes `build-mac-framework.sh`, this PR takes a minimal approach: it adds macOS support to the existing `build-ios-framework.sh` functions (`get_architecture`, `get_deployment_target`, `create_framework`, `build_universal_framework`) so the macOS slice is included in the universal xcframework.

`build-mac-framework.sh` is left in place for backward compatibility on this legacy branch.

### Motivation

Including the macOS slice in the universal xcframework allows [react-native-macos](https://github.com/microsoft/react-native-macos) to use the same prebuilt Hermes artifacts as react-native for iOS, visionOS, and tvOS — without platform-specific patching.

## Test plan

CI should pass. The `build_apple_framework "macosx"` call is already exercised by `build-mac-framework.sh` — this PR just routes it through the same path as every other platform.

### DIFF
--- a/utils/build-ios-framework.sh
+++ b/utils/build-ios-framework.sh
@@ -12,15 +12,9 @@ set -e
 # Given a specific target, retrieve the right architecture for it
 # $1 the target you want to build. Allowed values: iphoneos, iphonesimulator, catalyst, xros, xrsimulator
 function get_architecture {
-    if [[ $1 == "iphoneos" || $1 == "xros" ]]; then
+    if [[ $1 == "iphoneos" || $1 == "appletvos" || $1 == "xros" ]]; then
       echo "arm64"
-    elif [[ $1 == "iphonesimulator" || $1 == "xrsimulator" ]]; then
-      echo "x86_64;arm64"
-    elif [[ $1 == "appletvos" ]]; then
-      echo "arm64"
-    elif [[ $1 == "appletvsimulator" ]]; then
-      echo "x86_64;arm64"
-    elif [[ $1 == "catalyst" ]]; then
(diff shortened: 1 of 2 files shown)


utils/build-ios-framework.sh
```diff
@@ -20,7 +20,7 @@ function get_architecture {
       echo "arm64"
     elif [[ $1 == "appletvsimulator" ]]; then
       echo "x86_64;arm64"
-    elif [[ $1 == "catalyst" ]]; then
+    elif [[ $1 == "catalyst" || $1 == "macosx" ]]; then
       echo "x86_64;arm64"
     else
```
cipolleschi: we can probably refactor this, at this point...

cipolleschi (changes requested):
This is great for unifying the build. However, I feel that we need to update the ci also, because, IIRC, github builds slices in isolation and it might not be calling build-ios-framework for mac os.

So basically we never hit these changes in CI in this way

Saadnajmi:
> This is great for unifying the build. However, I feel that we need to update the ci also, because, IIRC, github builds slices in isolation and it might not be calling build-ios-framework for mac os.
> 
> So basically we never hit these changes in CI in this way

I think before my change we always had a macos slice and job in CI, and after this change we make sure that that slice is _also_ in the universal xcframework by adding macos to build_universal_framework()

cipolleschi (approved):

cortinico (approved):

PR #1971, opened 2026-04-03
fix: include macOS slice in universal Hermes xcframework

Saadnajmi:
## Summary

Mirror of https://github.com/facebook/hermes/pull/1958 (targeting `static_h`), adapted for the `250829098.0.0-stable` release branch.

Unlike the `static_h` PR which consolidates all Apple build scripts and removes `build-mac-framework.sh`, this PR takes a minimal approach: it adds `build_apple_framework "macosx"` and includes `"macosx"` in the `create_universal_framework` call so the macOS slice is part of the universal xcframework.

`build-mac-framework.sh` is left in place for backward compatibility on this release branch.

### Motivation

Including the macOS slice in the universal xcframework allows [react-native-macos](https://github.com/microsoft/react-native-macos) to use the same prebuilt Hermes artifacts as react-native for iOS, visionOS, and tvOS — without platform-specific patching.

## Test plan

CI should pass. The `build_apple_framework "macosx"` call is already exercised by `build-mac-framework.sh` — this PR just routes it through the same path as every other platform.

### DIFF
--- a/.github/workflows/build-hermes-macos.yml
+++ b/.github/workflows/build-hermes-macos.yml
@@ -57,7 +57,7 @@ jobs:
           chmod +x ./utils/build-apple-framework-rn.sh
           source ./utils/build-apple-framework-rn.sh
           prepare_dest_root_for_ci
-      - name: Create fat framework for iOS
+      - name: Create universal framework
         shell: bash
         run: |
           echo "[HERMES] Creating the universal framework"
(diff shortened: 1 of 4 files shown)


cipolleschi (approved):

cipolleschi (changes requested):
this change the build-ios-framework script, but shouldn't we also update the CI?

Otherwise CI will still call the other script as we build the slices in their own action.

Saadnajmi:
> this change the build-ios-framework script, but shouldn't we also update the CI?
> 
> Otherwise CI will still call the other script as we build the slices in their own action.

Similar to the comment for the main branch PR, I think before my change we always had a macos slice and job in CI, and after this change we make sure that that slice is _also_ in the universal xcframework by adding macos to build_universal_framework()

Saadnajmi:
I see the ci failures.. I'll take another look

cipolleschi (approved):

Issue #1853, opened 2025-12-02
When is Sampler::platformSuspendVMAndWalkStack called?


coolsoftwaretyler:
## Problem

Hey folks, I'm investigating a crash on Android that provides me with the following error:

```sh
invalid pthread_t 0x<sanitized> passed to pthread_kill
```

Comes with a stack trace like this:

```sh
*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 0, tid: 5131 >>> com.my.ap <<<

backtrace:
  #00  pc 0x000000000007123c  /apex/com.android.runtime/lib64/bionic/libc.so (abort+160)
  #01  pc 0x0000000000082c28  /apex/com.android.runtime/lib64/bionic/libc.so (__pthread_internal_find(long, char const*)+196)
  #02  pc 0x0000000000082b44  /apex/com.android.runtime/lib64/bionic/libc.so (__pthread_internal_gettid(long, char const*)+12)
  #03  pc 0x0000000000083948  /apex/com.android.runtime/lib64/bionic/libc.so (pthread_kill+52)
  #04  pc 0x00000000001105e0  /data/app/~~GVAFjiv3MRHi6wN_jyFNGQ==/com.my.app-dfsyqHFTuzXmYnuve2e92g==/split_config.arm64_v8a.apk!libhermes.so (BuildId: b06c77a49801680608345e3c05bd59aba90e9f19)
  #05  pc 0x0000000000110a60  /data/app/~~GVAFjiv3MRHi6wN_jyFNGQ==/com.my.app-dfsyqHFTuzXmYnuve2e92g==/split_config.arm64_v8a.apk!libhermes.so (BuildId: b06c77a49801680608345e3c05bd59aba90e9f19)
  #06  pc 0x000000000011095c  /data/app/~~GVAFjiv3MRHi6wN_jyFNGQ==/com.my.app-dfsyqHFTuzXmYnuve2e92g==/split_config.arm64_v8a.apk!libhermes.so (BuildId: b06c77a49801680608345e3c05bd59aba90e9f19)
  #07  pc 0x0000000000110d78  /data/app/~~GVAFjiv3MRHi6wN_jyFNGQ==/com.my.app-dfsyqHFTuzXmYnuve2e92g==/split_config.arm64_v8a.apk!libhermes.so (BuildId: b06c77a49801680608345e3c05bd59aba90e9f19)
[...]

PR #1995, opened 2026-04-23
Fix sampling profiler race when registered thread exits

antonis:
## Summary

Fix a crash in the sampling profiler where `pthread_kill` is invoked on a thread that has exited but whose corresponding `SamplingProfiler` is still registered. On Android bionic this aborts the process with `invalid pthread_t 0x... passed to pthread_kill`; on other POSIX platforms the sampler silently signals a stale thread ID.

Root cause: thread termination was never synchronized with the sampling path. If a JS thread exited while a host held the Runtime alive, the sampler would still attempt to sample it.

Fix: register each profiler with a thread-local `ThreadDeathGuard` backed by `pthread_key_create`. Its destructor, which runs before the `pthread_t` is reaped, invalidates the registered thread under the existing `runtimeDataLock_`. A `shared_ptr`-based `ProfilerHandle` keeps the synchronization state alive across either-end destruction, so there is no UAF if the profiler is destroyed on a different thread than the one that registered it. The sampler now skips any profiler whose registered thread has exited, and also returns cleanly on a non-fatal `ESRCH` from `pthread_kill` on platforms where that window is observable.

Related: #1853 and getsentry/sentry-react-native#5441.

## Test Plan

- New unit test `SamplingProfilerTest.SamplingAfterRegisteredThreadExitDoesNotCrash` reproduces the scenario deterministically on all POSIX platforms and asserts that the registered thread is invalidated after `std::thread::join()`.
- Existing `SamplingProfilerTest` suite continues to pass.
- Local build and test run on macOS arm64 (Xcode 16.4, ninja) against `static_h`:
  ```
  $ ninja -C build HermesVMRuntimeTests
  [484/484] Linking CXX executable unittests/VMRuntime/HermesVMRuntimeTests

  $ ./build/unittests/VMRuntime/HermesVMRuntimeTests --gtest_filter='SamplingProfilerTest.*'
  [==========] Running 4 tests from 1 test case.
  [----------] 4 tests from SamplingProfilerTest
  [ RUN      ] SamplingProfilerTest.Invariants
  [       OK ] SamplingProfilerTest.Invariants (96 ms)
  [ RUN      ] SamplingProfilerTest.RegisterDifferentThread
  [       OK ] SamplingProfilerTest.RegisterDifferentThread (29 ms)
  [ RUN      ] SamplingProfilerTest.RegisterIdenticalThread
  [       OK ] SamplingProfilerTest.RegisterIdenticalThread (30 ms)
  [ RUN      ] SamplingProfilerTest.SamplingAfterRegisteredThreadExitDoesNotCrash
  [       OK ] SamplingProfilerTest.SamplingAfterRegisteredThreadExitDoesNotCrash (85 ms)
  [==========] 4 tests from 1 test case ran. (240 ms total)
  [  PASSED  ] 4 tests.

  $ ./build/unittests/VMRuntime/HermesVMRuntimeTests
  [==========] 351 tests from 79 test cases ran.
  [  PASSED  ] 351 tests.
  ```

### DIFF
--- a/lib/VM/Profiler/SamplingProfilerPosix.cpp
+++ b/lib/VM/Profiler/SamplingProfilerPosix.cpp
@@ -9,6 +9,7 @@
 
 #if defined(HERMESVM_SAMPLING_PROFILER_POSIX)
 
+#include "hermes/Support/ErrorHandling.h"
 #include "hermes/Support/Semaphore.h"
 #include "hermes/VM/Callable.h"
 #include "hermes/VM/HostModel.h"
@@ -28,8 +29,11 @@
 #include <chrono>
 #include <cmath>
 #include <csignal>
+#include <memory>
+#include <mutex>
 #include <random>
 #include <thread>
+#include <vector>
 
 namespace hermes {
 namespace vm {
@@ -39,14 +43,32 @@ namespace {
 /// Name of the semaphore.
 constexpr char kSamplingDoneSemaphoreName[] = "/samplingDoneSem";
 
+struct SamplingProfilerPosix;
+
+/// Shared state between a SamplingProfilerPosix and any per-thread guards
+/// that reference it. Outlives whichever of the two is destroyed first, so
+/// thread-death handlers can safely observe "profiler is gone" without a
+/// use-after-free.
+struct ProfilerHandle {
+  std::mutex mu;
+  /// Back-pointer to the owning profiler. Cleared by the profiler's
+  /// destructor, at which point thread-death handlers skip this entry.
+  SamplingProfilerPosix *profiler{nullptr};
+};
+
 struct SamplingProfilerPosix : SamplingProfiler {
   SamplingProfilerPosix(Runtime &rt);
   ~SamplingProfilerPosix() override;
 
-  /// Thread that this profiler instance represents. This can be updated as the
-  /// runtime is invoked on different threads. Must only be accessed while
-  /// holding the runtimeDataLock_.
+  /// Thread that this profiler instance represents. A value of 0 means the
+  /// registered thread has exited or no thread is currently registered.
+  /// Must only be accessed while holding the runtimeDataLock_.
   pthread_t currentThread_;
+
+  /// Shared state with per-thread death guards referencing this profiler.
+  /// Set once at construction, never reassigned. The underlying object
+  /// outlives this profiler if any thread guard still references it.
+  std::shared_ptr<ProfilerHandle> handle_;
 };
 
 struct SamplerPosix : Sampler {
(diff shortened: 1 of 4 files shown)


tmikov:
Hi, the `main` branch is not actively maintained, so no guarantees about when this can be reviewed or merged. Does the problem exist in V1?

antonis:
Thanks for the heads-up. Confirmed the same code path exists in `static_h` (same `pthread_kill(posixProfiler->currentThread_, SIGPROF)` call, no thread-death synchronization), so I've retargeted this PR to `static_h` and force-pushed 🙇

antonis:
Hello 👋 
I was wondering if there is any feedback on this PR or guidance on how to get it merged 🙇

lavenzg:
> Hello 👋 I was wondering if there is any feedback on this PR or guidance on how to get it merged 🙇

Thanks for the update! We will take a look.

yonitou:
Hello 👋  ! 
Any news on this one ? We are experiencing the same issue described [here](https://github.com/getsentry/sentry-react-native/issues/5441)

lib/VM/Profiler/SamplingProfilerPosix.cpp
```diff
@@ -213,14 +300,32 @@ void Sampler::platformPostSampleStack(SamplingProfiler *localProfiler) {}
...
   // held by the caller, ensuring the runtime won't start to be used on
   // another thread before sampling begins.
-  pthread_kill(posixProfiler->currentThread_, SIGPROF);
+  int result = pthread_kill(posixProfiler->currentThread_, SIGPROF);
+  if (result != 0) {
+    // On non-Android POSIX, pthread_kill may return ESRCH if the target
+    // terminated in the narrow window where its pthread_t is still in the
+    // thread list but the thread has exited.
+    self->profilerForSig_.store(nullptr, std::memory_order_release);
+    return false;
```
lavenzg: I think this is not the same as the below `samplingDoneSem_.wait()` case.. In this case we want to skip this profiler and keep the sampling loop running for other profilers.

lib/VM/Profiler/SamplingProfilerPosix.cpp
```diff
@@ -78,10 +100,69 @@ struct SamplerPosix : Sampler {
...
+/// the thread that will host the profiler.
+void attachHandleToCurrentThread(std::shared_ptr<ProfilerHandle> handle) {
+  ensureThreadDeathGuardKey();
+  auto *guard = static_cast<ThreadDeathGuard *>(
+      pthread_getspecific(g_threadDeathGuardKey));
+  if (!guard) {
+    guard = new ThreadDeathGuard();
+    pthread_setspecific(g_threadDeathGuardKey, guard);
+  }
+  guard->handles.push_back(std::move(handle));
```
lavenzg: We should remove stale handle. Maybe right here before adding the new handle.

lavenzg (changes requested):
Sorry for the late review!

lib/VM/Profiler/SamplingProfilerPosix.cpp
antonis: Thank you for your feedback @lavenzg 🙇 
I've changed it to use a tri-state result with 0d875d3bec20974c50dc8042344115ad4f8632a5

lib/VM/Profiler/SamplingProfilerPosix.cpp
antonis: Updated with cd5d56cc2f73f822afebd12e5ff2abdf4fe8c4e9

lib/VM/Profiler/SamplingProfilerPosix.cpp
lavenzg: > Thank you for your feedback @lavenzg 🙇 I've changed it to use a tri-state result with [0d875d3](https://github.com/facebook/hermes/commit/0d875d3bec20974c50dc8042344115ad4f8632a5)

Can you add a test to cover this case? Thanks!

antonis: Added a test with 7b157c132c0d5dd9ee900d735f59452b7b0c595f

antonis:
I saw a lint failure and pushed a fix https://github.com/facebook/hermes/pull/1995/commits/37bb36ac2891a332d9389dcccc93670a641b6550

I don't have access to the log of the internal linter so feel free to ignore if this is irrelevant.

PR #1998, opened 2026-04-26
FlowChecker: Index constraint per arrow parameter in contextual typing

lee-sihun:
## Summary

The contextual-typing loop in `FlowChecker::ExprVisitor::visit(ArrowFunctionExpressionNode *, ..., Type *constraint)` walks arrow parameters and assigns types from `constraintFnType->getParams()[i]`, but the index `i` was never incremented inside the range-based loop.

As a result, every arrow parameter was typed as the first constraint parameter, so property access on the second and later parameters of a multi-parameter callback failed to type-check:

```ts
type Entry   = {name: string};
type Counter = {value: number};

function combine(cb: (l: Entry, r: Counter) => string): string {
  return cb(({name: 'left'}: any), ({value: 7}: any));
}

combine((l, r) => l.name + r.value);
// Before this PR: `r` is typed as Entry (the first constraint param),
// so `r.value` fails with `property 'value' not defined in object`.
```

This PR increments `i` after pushing each parameter so the Nth arrow parameter receives the Nth constraint parameter type. The existing `i < constraintFnType->getParams().size()` bounds guard keeps the fallback to `any` when the arrow parameter count exceeds the constraint arity, so mismatched-arity cases are unchanged.

A focused lit test is added that fails before the fix (`property 'value' not defined in object`) and passes after.

## Test Plan

- `ninja -C build check-hermes` — full suite passes; 2897 expected passes, no regressions.
- The new test `test/hermes/flow/arrow-multi-param-contextual.js` fails on `static_h` without the fix and passes with the fix applied.
- Reproduces the original symptom on a minimal multi-parameter callback (`(left: Entry, right: Counter) => string`) and asserts that property access on the second parameter type-checks.

### DIFF
--- a/lib/Sema/FlowChecker-expr.cpp
+++ b/lib/Sema/FlowChecker-expr.cpp
@@ -300,6 +300,7 @@ class FlowChecker::ExprVisitor {
         }
 
         params.push_back(typedFnParam);
+        ++i;
       }
 
       Type *returnType;
--- a/test/hermes/flow/arrow-multi-param-contextual.js
+++ b/test/hermes/flow/arrow-multi-param-contextual.js
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ */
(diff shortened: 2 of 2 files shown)


tmikov:
Good catch. Will review with high priority!

PR #2018, opened 2026-05-12
Implement `Promise.try(func, arg1, arg2, /* …, */ argN)`

retyui:
## Summary

`Promise.try(fn, ...args)` :
- Specification: https://tc39.es/ecma262/#sec-promise.try
- Docs: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/try
- TS: [TypeScript/src/lib/esnext.promise.d.ts](https://github.com/microsoft/TypeScript/blob/66edca11c98ade9a5e2a2b019fdad7d58ee9d4ac/src/lib/esnext.promise.d.ts#L15)

The `Promise.try()` static method takes a callback of any kind (returns or throws, synchronously or asynchronously) and wraps its result in a [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise).

```tsx
Promise.try(() => func(arg1, arg2));
// or you can do this 
Promise.try(func, arg1, arg2);
```

## Test Plan

```bash
LIT_OPTS="-j1" LIT_FILTER="promise-try.js" cmake --build build --target check-hermes
```

### DIFF
--- a/lib/InternalJavaScript/01-Promise.js
+++ b/lib/InternalJavaScript/01-Promise.js
@@ -417,6 +417,13 @@
     return { promise: promise, resolve: resolve, reject: reject };
   };
 
+  core.try = function (callback) {
+    var args = arguments;
+    return new this(function (res) {
+      res(callback.apply(undefined, iterableToArray(args).slice(1)))
+    });
+  };
+
   core.prototype.finally = function (f) {
     return this.then(function (value) {
       return core.resolve(f()).then(function () {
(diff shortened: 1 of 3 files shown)


lavenzg:
@retyui We have landed the Promise stack to static_h yesterday. But we don't plan to cherry-pick it into Hermes V1, since this stack changes existing Promise methods and may break things. It will become available to RN only after next Hermes stable release. 

Your change here only adds a new method, so it's relatively safe. If you need to use it in RN, we encourage you to make a PR against the Hermes V1 branch (250829098.0.0-stable) instead.

PR #2022, opened 2026-05-15
ci: declare workflow-level `contents: read` on 2 workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on 2 workflows in `.github/workflows/` that don't call a GitHub API beyond the initial checkout.

The following files were left implicit because they reference `GITHUB_TOKEN` / use a write-scope action / trigger on `pull_request_target`. Those scopes are best declared by maintainers: `build-apple-slices-hermes.yml`, `build-hermes-macos.yml`, `build-hermesc-apple.yml`, `build-hermesc-linux.yml`, `build-hermesc-windows.yml`.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,10 @@ on:
   pull_request:
     branches:
       - static_h
+
+permissions:
+  contents: read
+
 jobs:
   macos:
     runs-on: macos-15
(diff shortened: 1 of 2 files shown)


cortinico (approved):

PR #2030, opened 2026-05-21
Fix async arrow non simple params bug

micleo2:
Summary:
Add missing logic in async arrow functions for non simple parameters. This logic is copied over from the existing codepath in `genBasicFunction`.

Reviewed By: avp

Differential Revision: D81596715

fbshipit-source-id: ac3accf710f3d32db24743674e590700beb18bb0

### DIFF
--- a/lib/IRGen/ESTreeIRGen-func.cpp
+++ b/lib/IRGen/ESTreeIRGen-func.cpp
@@ -251,13 +251,15 @@ NormalFunction *ESTreeIRGen::genCapturingFunction(
         capturedState);
     return newFunc;
   }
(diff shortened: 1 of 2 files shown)


hsource:
@micleo2 - I just encountered this bug while upgrading to React Native 0.86.2 (Expo 57). It looks like the companion fixes (#2045 and #2046) have already been merged.

Is there anything else we need to do to get this merged? We're using the Expo workaround (https://github.com/expo/expo/pull/45601) for now but it seems likely this can help still!

@robhogan since I think you wrote the original PRs

Issue #2035, opened 2026-05-26
`CreateBaseClass` result overwritten after `SpillRegisters` for named class expression


leegeunhyeok:
## Bug Description

Optimized HBC can miscompile a named class expression assigned to a variable under high register pressure.

The class binding becomes the class home object instead of the constructor function, so static methods are callable but constructing the class fails.

```txt
TypeError: Object cannot be used as a constructor.
```

- [x] I have run `gradle clean` and confirmed this bug does not occur with JSC
- [x] The issue is reproducible with the latest version of React Native.

Hermes git revision (if applicable): d8001980e852cd81b5f84927b8c14f6fa68d8345
React Native version:
OS:
Platform (most likely one of arm64-v8a, armeabi-v7a, x86, x86_64):

## Steps To Reproduce

> This reproduces on the `static_h` branch (d8001980e852cd81b5f84927b8c14f6fa68d8345) and `250829098.0.8`, `250829098.0.13` versions from RN releases

Gist: [repro.js](https://gist.github.com/leegeunhyeok/143f93da8d4d983240b19f02b2505277)

1. Build Hermes binaries
2. Run `./build/bin/hermes -O0 repro.js`

code example: (gist)

## The Expected Behavior

Expected:

```
typeof TestClass: function
has prototype: true
OK: 1234
```

Actual:

```
typeof TestClass: object
has prototype: false
Uncaught TypeError: Object cannot be used as a constructor.
    at createInstance (repro.js:517:26)
    at global (repro.js:1035:40)
```

NOTE: `-O0` (without optimization) works correctly.

### What Reproduces

This reproduces:

```js
var TestClass = class _TestClass {
  static createInstance(value) {
    return new _TestClass(value);
  }
  constructor(value) {
    this.value = value;
  }
};
```

This does not reproduce under the same register pressure:

```js
class TestClass {
  static createInstance(value) {
    return new TestClass(value);
  }
  constructor(value) {
    this.value = value;
  }
}
```

### Bytecode Observation

Problematic bytecode:

```js
CreateBaseClass   r0, r1, ...
MovLong           r265, r0
MovLong           r265, r1
...
StoreToEnvironment ..., r265
```

`CreateBaseClass` has two outputs:

1. Arg1 = output register for the closure/class constructor
2. Arg2 = output register for the home object

So the bytecode above effectively does:

```
r0 = class constructor
r1 = home object

r265 = r0
r265 = r1
```

The second `MovLong` overwrites the constructor with the home object.
[...]

PR #2036, opened 2026-05-28
Cherry-pick: Spill `CreateClassInst` output last

lavenzg:
## Summary

Cherry-pick of 5423909e212d9464ebf70633a02a61b9e43963df from `static_h` to the `250829098.0.0-stable` branch.

Preserves spill store ordering for instructions whose result can alias a writable operand output. `CreateClassInst` writes the home object through an operand and returns the constructor as the instruction output; when both are spilled to the same high register, storing the instruction output last preserves the bytecode interpreter's intended write order.

## Test plan

- Cherry-pick applied cleanly (auto-merge on `lib/BCGen/HBC/Passes.cpp` and `unittests/BCGen/HBC.cpp`)
- Test added: `test/hermes/regress-class-spill-register.js`
- Unit test added in `unittests/BCGen/HBC.cpp`


lavenzg:
Fix #2035 in OSS stable.

PR #2046, opened 2026-06-04
Cherry-pick: Fix class variable caching bugs when class is in finally block

ramonclaudio:
## Summary

Cherry-pick of `1e94fbe0ebb46d676ff656287e17c58839765e73` from `static_h` to the `250829098.0.0-stable` branch.

A `finally` block emits its body twice, once for normal flow and once for the exception path. On the second emission, legacy class compilation created fresh internal Variables (class and prototype vars, the instance-element init function, computed field keys, private brands) while cached functions reached via `findCompiledEntity` still referenced the originals, miscompiling classes declared in `finally` blocks. The fix caches these in a `LegacyClassVars` struct keyed by the class AST node (`legacyClassVars_`) and emits private-brand stores unconditionally.

This landed on `static_h` after the branch was cut (2025-08-29), so it is missing from the V1 line React Native ships (needed by facebook/react-native#56816). The stacked typed-class sibling `639e5d6af` is intentionally not included: it fixes the Static Hermes typed class path, which standard React Native JS does not exercise. Authored by @tmikov, reviewed internally by @fbmal7.

## Test plan

- Clean cherry-pick, no conflicts. Carries the upstream tests: new `test/IRGen/class-static-in-finally.js` and `test/hermes/class-in-finally.js`, a CHECK update to `test/IRGen/regress-class-recompiled.js` that drops the duplicate `?C1#1`, `?C1.prototype#1`, `<instElemInitFunc:C1>#1` Variables the bug produced, and a brand-store ordering update to `test/IRGen/private-properties.js`.
- Built `hermes` and `hermesc` from `250829098.0.0-stable` (`517edaff8`): the baseline fails `regress-class-recompiled.js` and the runtime `class-in-finally.js`. With this cherry-pick all four bundled tests pass.
- `LIT_FILTER="class-in-finally|regress-class-recompiled|class-static-in-finally" cmake --build build --target check-hermes` passes.

### DIFF
--- a/lib/IRGen/ESTreeIRGen-legacy-class.cpp
+++ b/lib/IRGen/ESTreeIRGen-legacy-class.cpp
@@ -62,43 +62,35 @@ namespace irgen {
 /// inspecting the kind of a `Decl`, it's possible to know which kind of element
 /// in the side table the customData field of the `Decl` points to.
 void ESTreeIRGen::emitPrivateNameDeclarations(
+    ESTree::ClassLikeNode *classNode,
     sema::LexicalScope *scope,
     Identifier className) {
-  Variable *staticBrand = nullptr;
-  Variable *instanceBrand = nullptr;
+  // Get the cached variables for this class. Brand Variables must be cached
+  // for recompilation (e.g., finally blocks) to ensure consistency.
+  auto &cachedVars = legacyClassVars_[classNode];
+
   /// Lazily create and return the correct private brand for the given static
-  /// level.
-  auto getPrivateBrand =
-      [this, &className, &staticBrand, &instanceBrand](bool isStatic) {
-        if (isStatic) {
-          if (!staticBrand) {
-            staticBrand = Builder.createVariable(
-                curFunction()->curScope()->getVariableScope(),
-                Twine("?static_brand_") + className.str(),
-                Type::createPrivateName(),
-                /* hidden */ true);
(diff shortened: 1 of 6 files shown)


ramonclaudio:
Hey @lavenzg @cipolleschi any chance these make it in before the next release?

@cortinico and @robhogan asked for cherry-picks onto `250829098.0.0-stable` instead of babel workarounds (this one and #2045). See: [react/react-native#56816](https://github.com/react/react-native/pull/56816).

lavenzg:
We have a new stable (260318099.0.0-stable) cut a few months ago and it includes this fix. And I believe this new stable will be used by next release of RN (starting from rc1). Correct me if I'm wrong @cipolleschi.

ramonclaudio:
@lavenzg yeah, the new stable has all three fixes, but nothing on the RN side uses it yet. `v0.87.0-rc.1` from July 15 still pins the old one. This branch is already bumped for `.0.17` though, so if this, #2045, and #2030 make it in, current RN versions are one pin bump away from the fix.

lavenzg:
Riccardo has confirmed that v0.87 will not use the new stable. So we will merge this. I'll let @micleo2 have a double check.

ramonclaudio:
Thank you! @lavenzg

micleo2:
Looks good to me

PR #2069, opened 2026-06-24
Support default perf report rows in agent perf tools

lin-hongkuan:
## Summary

`profile_to_json.py` and `runtime_call_profiler.py` only parsed perf rows with one percentage column, with or without a sample count. Default `perf report --stdio` output commonly uses `Children` and `Self` percentage columns before `Command`, `Shared Object`, and `Symbol`, so those rows were ignored.

This change adds a shared perf report row parser and reuses it from both tools. The parser now handles:

- existing single-percentage rows with samples
- existing single-percentage rows without samples
- default `Children` / `Self` rows, using `Self` as the per-symbol percentage
- bracketed modules such as `[kernel.kallsyms]`
- module names with spaces
- function names with bracketed suffixes such as `[abi:cxx11]`

## Test Plan

```powershell
python -m unittest agent-perf\tests\test_profile_to_json.py agent-perf\tests\test_runtime_call_profiler.py
git diff --check
```

All commands passed locally.

## Branch

`codex/agent-perf-default-perf-report`

## Commit

`0a00509e0a8546f7cb148dda0fab3f9efbb332bb`

### DIFF
--- a/agent-perf/tests/test_profile_to_json.py
+++ b/agent-perf/tests/test_profile_to_json.py
@@ -58,6 +58,48 @@ class TestParsePerfReport(unittest.TestCase):
         self.assertEqual(results[1]["samples"], 0)
         self.assertEqual(results[1]["percentage"], 5.67)
 
+    def test_parse_default_children_self_format(self):
+        """Test parsing perf's default Children/Self percentage columns."""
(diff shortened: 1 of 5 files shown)


lavenzg:

PR #2070, opened 2026-06-24
Keep skipped source lines in SH source maps

lin-hongkuan:
## Summary

`sh_source_mapper.py` intentionally skips empty source lines and bare braces when building mappings, but it also skipped advancing the current JS line number for those lines. That made every later mapping in the same `#line` region drift upward when the source contained blank lines or standalone braces.

This change keeps the existing behavior of not emitting mapping entries for empty lines and bare braces, while still accounting for those source lines before mapping subsequent code.

## Test Plan

```powershell
python -m unittest agent-perf\tests\test_sh_source_mapper.py
git diff --check
```

All commands passed locally.

## Branch

`codex/sh-source-mapper-line-accounting`

## Commit

`75f7d4b6d2a899cc6a81368fa35e31868284a5f9`


lavenzg:

PR #2071, opened 2026-06-24
Flush trailing generated-code call blocks

lin-hongkuan:
## Summary

`generated_code_antipatterns.py` reports excessive consecutive `_sh_*` runtime calls only when a non-call line terminates the block. If a long runtime-call block reaches the end of a generated function, function-boundary handling runs first and resets state, so the high-density call block is missed.

This change extracts call-block flushing into a helper and uses it both for ordinary non-call terminators and function-end boundaries. A regression test covers a function ending immediately after ten consecutive `_sh_*` calls.

## Test Plan

```powershell
python -m unittest agent-perf\tests\test_generated_code_antipatterns.py
git diff --check
```

All commands passed locally.

## Branch

`codex/generated-code-tail-call-blocks`

## Commit

`0b7f73e971327862e6e56a8d1648a369127e4d25`

### DIFF
--- a/agent-perf/tests/test_generated_code_antipatterns.py
+++ b/agent-perf/tests/test_generated_code_antipatterns.py
@@ -78,6 +78,26 @@ class TestGeneratedCodeAntipatterns(unittest.TestCase):
         finally:
             os.unlink(filepath)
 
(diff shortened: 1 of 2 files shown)


lavenzg:

PR #2072, opened 2026-06-24
Handle single-line generated C functions

lin-hongkuan:
## Summary

`generated_code_analyzer.py` did not close functions whose opening and closing braces are on the same line. A generated C function such as `static ... { return _sh_ljs_add(...); }` left parser state open, so the next generated function could be folded into the previous one.

This change factors function finalization into a helper and flushes immediately when a function header line has balanced braces. The existing multi-line function behavior is unchanged.

## Test Plan

```powershell
python -m unittest agent-perf\tests\test_generated_code_analyzer.py
git diff --check static_h..codex/generated-analyzer-single-line-functions
```


lavenzg:

PR #2073, opened 2026-06-24
Handle single-line functions in type coverage

lin-hongkuan:
## Summary

`type_coverage_analyzer.py` had the same one-line generated C function boundary issue in its per-function runtime-call extractor. A function that starts and ends on the same line was not finalized immediately, which could incorrectly attribute later calls to the wrong function.

This change adds a small finalization helper and flushes balanced one-line functions as soon as they are seen.

## Test Plan

```powershell
python -m unittest agent-perf\tests\test_type_coverage_analyzer.py
git diff --check static_h..codex/type-coverage-single-line-functions
```


lavenzg:

PR #2074, opened 2026-06-24
Parse generated function ids in antipattern scan

lin-hongkuan:
## Summary

`generated_code_antipatterns.py` already had a test comment for generated functions with `#id` suffixes, but its function-definition regex did not actually accept that suffix. Findings inside declarations such as `_2_bar#123(...)` could therefore lose the generated function attribution.

This change accepts an optional numeric `#id` suffix while preserving the base function name in reports.

## Test Plan

```powershell
python -m unittest agent-perf\tests\test_generated_code_antipatterns.py
git diff --check static_h..codex/generated-antipattern-function-ids
```


lavenzg:

PR #2075, opened 2026-06-24
Preserve nm symbol names with spaces

lin-hongkuan:
## Summary

`binary_size_analyzer.py` parsed `nm --print-size --radix=d` rows with an unrestricted `split()` and then kept only `parts[3]` as the symbol name. If `nm` output contains a symbol display name with spaces or suffix text, the analyzer truncates it before comparison and classification.

This change limits splitting to the first three fields and keeps the remainder of the row as the symbol name.

## Test Plan

```powershell
python -m unittest agent-perf\tests\test_binary_size_analyzer.py
git diff --check static_h..codex/binary-size-symbol-names-with-spaces
```


lavenzg:

PR #2076, opened 2026-06-24
Accept broader perf annotate file markers

lin-hongkuan:
## Summary

`annotate_source.py` only recognized file marker lines that looked like absolute Unix paths without spaces. That leaves relative paths, Windows paths, and paths with spaces as `<unknown>` in JSON and human output.

This change keeps the existing marker structure but accepts any non-empty path text between the marker colon and trailing colon.

## Test Plan

```powershell
python -m unittest agent-perf\tests\test_annotate_source.py
git diff --check static_h..codex/annotate-source-file-markers
```


lavenzg:

PR #2077, opened 2026-06-24
Treat missing PMU counters as unavailable

lin-hongkuan:
## Summary

`pmu_anomalies.py` used `data.get(..., 0)` for every raw counter. That makes a missing numerator indistinguishable from a real zero counter, so partially collected PMU JSON can produce misleading `0.0` rates instead of `N/A`.

This change returns `None` when either side of a metric is absent, preserving the existing divide-by-zero handling for present zero denominators.

## Test Plan

```powershell
python -m unittest agent-perf\tests\test_pmu_anomalies.py
git diff --check static_h..codex/pmu-missing-counters
```


lavenzg:

PR #2078, opened 2026-06-24
Recognize alternate RA spill slot forms

lin-hongkuan:
## Summary

`regalloc_report.py` counted spill instructions but only extracted spill slot IDs written as `%stack5`. Other common textual forms such as `%stack.5` or `%stack[5]` were counted as spills but omitted from `spill_slots`.

This change centralizes spill-slot extraction and recognizes all three forms while keeping the existing `%stack5` behavior.

## Test Plan

```powershell
python -m unittest agent-perf\tests\test_regalloc_report.py
git diff --check static_h..codex/regalloc-spill-slot-forms
```


lavenzg:

PR #2079, opened 2026-06-24
Match indented perf annotate source functions

lin-hongkuan:
Summary:
- Recognize indented function headers in perf annotate source output.
- Keep file marker lines from being treated as function headers.

Test Plan:
- python -m unittest agent-perf\tests\test_annotate_source.py
- git diff --check


lavenzg:

PR #2080, opened 2026-06-24
Skip single-line JS block comments

lin-hongkuan:
Summary:
- Skip single-line /* ... */ block comments before running JS anti-pattern checks.
- Prevent comment-only eval/slow-path examples from being reported as real findings.

Test Plan:
- python -m unittest agent-perf\tests\test_js_antipattern_scan.py
- git diff --check


lavenzg:

PR #2081, opened 2026-06-24
Skip C++ comments in antipattern scan

lin-hongkuan:
Summary:
- Skip whole-line // comments and block comments before applying C++ anti-pattern heuristics.
- Avoid false positives for patterns that appear only inside comments.

Test Plan:
- git diff --check


lavenzg:

PR #2082, opened 2026-06-24
Limit C++ single-line loop scope

lin-hongkuan:
Summary:
- Reset C++ loop tracking after scanning a single-line loop statement.
- Prevent string concatenation after such a loop from being reported as loop-local work.

Test Plan:
- git diff --check


lavenzg:

PR #2085, opened 2026-06-24
Filter single-file C++ scan inputs

lin-hongkuan:
Summary:
- Apply the C++ extension allowlist to single-file inputs as well as directory inputs.
- Return no files for non-C++ single-file paths.

Test Plan:
- git diff --check


lavenzg:

PR #2086, opened 2026-06-24
Accept 0x-prefixed annotate addresses

lin-hongkuan:
Summary:
- Accept perf annotate disassembly addresses that already include a 0x prefix.
- Keep normalized output in the existing 0x-prefixed address format.

Test Plan:
- python -m unittest agent-perf\tests\test_annotate_disasm.py
- git diff --check


lavenzg:

PR #2087, opened 2026-06-24
Parse quoted RA function names

lin-hongkuan:
Summary:
- Parse quoted shermes -dump-ra function display names that contain spaces.
- Preserve existing parsing for unquoted function names and #id suffixes.

Test Plan:
- python -m unittest agent-perf\tests\test_regalloc_report.py
- git diff --check


lavenzg:

PR #2088, opened 2026-06-24
Classify more binary-size arithmetic symbols

lin-hongkuan:
Summary:
- Classify additional _sh_ljs arithmetic runtime symbols in binary size reports.
- Cover div, mod, inc, dec, and negate alongside add, sub, and mul.

Test Plan:
- python -m unittest agent-perf\tests\test_binary_size_analyzer.py
- git diff --check


lavenzg:

PR #2089, opened 2026-06-25
Fix shermes Windows CC path splitting (#1994)

### LINKED ISSUE #1994: shermes native compilation fails on Windows (with Clang)
Compiling and executing richard.js: shermes richard.js =>
clang: error: no such file or directory: 'C'
clang: error: no such file or directory: 'C'
clang: error: no such file or directory: 'C'
C:\Program Files\LLVM\bin\clang.exe: execution failed

<!--
 
OS:
Windows 11

Shermes V1.0. Git pull last night

tangtaizong666:
Summary:
- Use the platform path-list separator for shermes CC include/library paths on Windows.
- Split generated CC path lists with `llvh::sys::EnvPathSeparator`.
- Add a Windows regression test for drive-letter paths in verbose native compilation output.

Test Plan:
- `cmake -U SHERMES_CC_INCLUDE_PATH -U SHERMES_CC_LIB_PATH -S work\hermes-git -B work\hermes-git\cmake-build-asan -G Ninja -DCMAKE_BUILD_TYPE=Debug -DHERMES_ENABLE_ADDRESS_SANITIZER=ON -DHERMES_ENABLE_NAPI=OFF -DHERMES_TEST_EXTRA_ARGS="--path C:/application/Git/usr/bin" -DCMAKE_MAKE_PROGRAM="C:\application\vs\Common7\IDE\CommonExtensions\Microsoft\CMake\Ninja\ninja.exe"`
- `cmake --build work\hermes-git\cmake-build-asan --target check-shermes --parallel 4` with `LIT_FILTER=windows-cc-path-separator`

Fixes #1994


tools/shermes/CMakeLists.txt
```diff
@@ -18,10 +18,24 @@ set(HERMES_ENABLE_EH OFF)
...
 set(SHERMES_CC_SYSCFLAGS "" CACHE STRING "C compiler system flags")
 set(SHERMES_CC_SYSLDFLAGS "" CACHE STRING "C compiler system linker flags")
-set(SHERMES_CC_INCLUDE_PATH
-  "${PROJECT_BINARY_DIR}/lib/config:${PROJECT_SOURCE_DIR}/include" CACHE STRING
+if(CMAKE_HOST_WIN32)
+  set(SHERMES_CC_PATH_SEPARATOR ";")
+else()
+  set(SHERMES_CC_PATH_SEPARATOR ":")
+endif()
+string(CONCAT SHERMES_CC_INCLUDE_PATH_DEFAULT
```
lavenzg: Use `JOIN` instead.

test/shermes/windows-cc-path-separator.js
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// REQUIRES: windows
+// RUN: (CC=cmd %shermes -v -o %t.exe %s || true) 2>&1 | %FileCheck %s
+
+// CHECK: cmd.exe
+// CHECK-NOT: {{(^| )-IC( |$)}}
```
lavenzg: There is no guarantee that `SHERMES_CC_INCLUDE_PATH` and `SHERMES_CC_LIB_PATH` will point to `C:` drive right?

test/shermes/windows-cc-path-separator.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// REQUIRES: windows
+// RUN: (CC=cmd %shermes -v -o %t.exe %s || true) 2>&1 | %FileCheck %s
```
lavenzg: I understand that we don't really need to compile the file, just use it to print out the include/lib paths. But it's still weird to pass `cmd` to CC. Could we just pass a  fake flag here so that we can still get the full failed command and do check on the paths?

lavenzg (commented):
Thanks for making the PR. Left some comments here.

tools/shermes/CMakeLists.txt
tangtaizong666: Done in 79a4ab3 — switched to `list(JOIN ...)`. The include and lib paths are now built from `*_ENTRIES` lists joined with `SHERMES_CC_PATH_SEPARATOR` (`;` on a Windows host, `:` otherwise), instead of hardcoding `:` inside a `string(CONCAT ...)`.

test/shermes/windows-cc-path-separator.js
tangtaizong666: Right — fixed in 79a4ab3. The CHECK-NOT patterns no longer assume the `C:` drive: they match any drive letter (`-I[A-Za-z]`, `-L[A-Za-z]`, `-Wl,-rpath [A-Za-z]`) and also assert the colon-joined `DRIVE:...:DRIVE:` shape (`{{[A-Za-z]:[^ ]*:[A-Za-z]:}}`) is absent, so the test catches the regression regardless of which drive the build tree lives on.

test/shermes/windows-cc-path-separator.js
tangtaizong666: Switched to a fake flag in 79a4ab3: `-Wc,--shermes-invalid-cc-option` is forwarded to CC, which fails, so we still get the full failed CC invocation to FileCheck against (`// CHECK: --shermes-invalid-cc-option`) while no longer overriding `CC=cmd`.

PR #2093, opened 2026-07-02
Add fast paths for Array.prototype shift/unshift/slice/splice/forEach/map/filter/every/some

zhelezkov:
Hey all,

I've been exploring the Hermes codebase to investigate some performance aspects of our React Native app, and I noticed a few potentially small but meaningful improvements around array fast-path handling.

**AI usage notice:** I used AI to help identify opportunities, implement the changes, review the code, and create benchmark tests. I also manually reviewed the code and validated the benchmark results to ensure the improvements are correct and meaningful.

Please let me know what I can improve/adjust to get this optimizations upstreamed into hermes!

## Description

Resolves the `TODO: Implement a fast path for actual arrays.` in several `Array.prototype` methods by using the existing `arrayFastPathCheck` to bypass the generic per-index property protocol for ordinary `JSArray`s:

- **`shift` / `unshift`**: move elements directly in the indexed storage under `NoAllocScope` and write the new length straight to the length slot. Guarded by `arrayFastPathCheck` against `runtime.arrayClass`, so accessors, index-like properties anywhere on the prototype chain, a read-only `length`, and frozen/sealed arrays all fall back to the generic path. The fast-path branches are wrapped in `NoLeakHandleScope` to ensure they allocate no handles.
- **`slice`**: copy elements straight from the source storage into the new array's storage. The check runs *after* argument coercion, so an array mutated by a `valueOf`/`toString` side effect fails the check (element count no longer matches the captured `len`) and falls back.
- **`forEach` / `map` / `filter` / `every` / `some`**: read each element through a new `ElementReader`, which tries the indexed storage first. The fast-path check is made once, on construction, rather than per element — that stays correct even though callbacks can mutate the array between reads: a non-empty slot in the indexed storage is always a plain own data property (converting an element to an accessor empties its slot first), `at()` re-checks the bounds against the *current* storage on every read, and `hasFastIndexProperties()` is re-checked so a slot can never be shadowed by an index-like named property. Holes fall back to the generic per-index read, which handles prototype-chain lookups (including Proxy prototypes installed mid-iteration).

Element moves and copies go through `GCHVType::copy()` / `copy_backward()` instead of per-element loops, which also required making `GCHVType` public in `ArrayStorageBase`.

Two helpers factor out the shared entry-point logic: `canReadElementsFast()` (is a direct read possible, including a `len <= UINT32_MAX` check so the fast paths can cast indices to `uint32_t`) and `canUseArrayFastPath()` / `ensureObjectAndLength()` (fetch the length once for both the fast and the slow path of `pop`/`shift`/`unshift`).

## Benchmarks

Apple M5 Max, macOS 26.5.2, `hermes` Release build with default options (HV64,
no JIT — plain interpreter, matching how Hermes ships in React Native).
Baseline is `static_h` at 10a40b5f1, built from a separate worktree so both
binaries come from identical configurations.

Methodology: each case auto-calibrates its iteration count to >= 100 ms per
timed run, reports the best of 5 in-process runs, and the two binaries are then
executed interleaved (A, B, A, B, ...) for 7 rounds, of which the median is
taken. Numbers are per call of the listed operation. A checksum script running
every benchmarked operation (dense, sparse and array-like receivers) produces
identical output on both binaries.

### Fast paths

| Case (10k dense elements unless noted) | Baseline | Fast paths | Speedup |
|---|---:|---:|---:|
| `shift()` + `push()`, 1000 ops on a 1k array | 8375 µs | 291 µs | **28.8×** |
| `unshift()` + `pop()`, 1000 ops on a 1k array | 8375 µs | 289 µs | **29.0×** |
| `slice()` (full copy) | 58.1 µs | 5.4 µs | **10.7×** |
| `slice(4000, 5000)` | 5.89 µs | 0.65 µs | **9.1×** |
| `every()` (full scan) | 137.7 µs | 97.7 µs | 1.41× |
| `some()` (full scan) | 137.7 µs | 97.7 µs | 1.41× |
| `splice(1000, 1)` + `splice(1000, 0, x)`, 100× on a 2k array | 230.5 µs | 168.0 µs | 1.37× |
| `forEach()` | 150.4 µs | 113.3 µs | 1.33× |
| `map()` | 152.3 µs | 115.2 µs | 1.32× |
| `filter()` | 182.6 µs | 142.6 µs | 1.28× |

`shift`/`unshift`/`slice` improve by an order of magnitude because the fast path
replaces a computed-descriptor lookup plus a property get/put per element with
raw indexed-storage moves; for `shift`/`unshift` the ratio grows with the array
length, since both paths are O(n) but with very different constants. The
callback-driven methods improve by ~1.3× because the per-element callback
invocation dominates in the interpreter. `splice` already had a fast path; the
gain there comes from replacing its per-element copy loop with
`GCHVType::copy()`.

### Regression guards

Untouched or already-fast operations, to confirm nothing regressed
(`indexOf`/`lastIndexOf`/`includes` already had a fast path and were only
refactored to share the new helper):

| Case | Baseline | Fast paths | Change |
|---|---:|---:|---:|
| `indexOf(-1)` (full scan) | 2.23 µs | 2.23 µs | +0.0% |
| `lastIndexOf(-1)` (full scan) | 1.92 µs | 1.89 µs | −1.6% |
| `includes(-1)` (full scan) | 2.23 µs | 2.24 µs | +0.7% |
| `join(",")` | 123.0 µs | 123.0 µs | +0.0% |
| `push()` + `pop()`, 1000 ops | 18.9 µs | 19.0 µs | +0.6% |
| `concat()` (10k + 10k) | 21.2 µs | 21.2 µs | +0.0% |
| `reverse()` | 4.46 µs | 4.21 µs | −5.5% (noisy, treat as parity) |

Cases that deliberately fail the fast-path checks, i.e. that pay for the added
guards without benefiting from them — sparse arrays, array-like receivers, and
an index-like property installed on `Object.prototype`:

| Case | Baseline | Fast paths | Change |
|---|---:|---:|---:|
| `forEach()` on a sparse 10k array (1 in 10 present) | 80.6 µs | 82.5 µs | +2.4% |
| `map()` on the same sparse array | 86.9 µs | 86.9 µs | +0.0% |
| `slice()` on the same sparse array | 74.2 µs | 73.2 µs | −1.3% |
| `indexOf(-1)` on the same sparse array | 73.2 µs | 74.7 µs | +2.0% |
| `forEach.call()` on a 10k array-like object | 406.3 µs | 410.2 µs | +1.0% |
| `map.call()` on a 10k array-like object | 410.2 µs | 418.0 µs | +1.9% |
| `slice.call()` on a 10k array-like object | 294.9 µs | 296.9 µs | +0.7% |
| `shift.call()` on an array-like object, 100 ops | 585.9 µs | 589.8 µs | +0.7% |
| `forEach()` on a dense 10k array, index-like property on the proto chain | 151.4 µs | 153.3 µs | +1.3% |
| `slice()` on the same | 58.6 µs | 58.6 µs | +0.0% |

## Test plan

- New lit test `test/hermes/array-fastpath.js` covering, for each method: holes, callback-driven mutation mid-iteration (shrink/grow), callbacks that throw or trigger a `gc()`, accessors installed by callbacks or by an argument coercion, Proxy prototypes installed mid-iteration, frozen/sealed arrays, read-only `length`, a length extended past the storage, storage that does not start at index 0, heap-allocated elements and boxed doubles moved across a GC, generic-path receivers (plain objects and `Array` subclasses), chained fast-path operations, and index-like properties on `Array.prototype`.
- Full `check-hermes` on a Debug build: all tests pass.
- Array lit tests also run on a `-DHERMESVM_HEAP_HV_MODE=HEAP_HV_BOXED` Debug build to force boxed doubles through the fast paths.

### DIFF
--- a/lib/VM/JSLib/Array.cpp
+++ b/lib/VM/JSLib/Array.cpp
@@ -2012,14 +2012,66 @@ CallResult<HermesValue> arrayPrototypeSort(void *, Runtime &runtime) {
   return lv.O.getHermesValue();
 }
 
+/// Read the element at index \p k of \p O inside an iteration loop that
+/// runs user callbacks (forEach/map/filter), storing it in \p kValue.
+/// If \p fastPath is true, first try reading the element directly out of
+/// the indexed storage. Correctness does not depend on the fast path check
+/// remaining true after a callback runs: a non-empty value in the indexed
+/// storage is always a plain own data property (converting an indexed
+/// property to an accessor deletes the indexed slot first), and at()
+/// re-checks the bounds against the current storage on every iteration.
+/// Everything else falls back to the generic per-index path.
+/// \param O the object being iterated. Must be a JSArray if \p fastPath is
+///   true.
+/// \param fastPath the result of arrayFastPathCheck on \p O, hoisted out of
+///   the loop.
+/// \param k the current index.
+/// \param kHandle a Handle containing \p k encoded as a number.
+/// \param[out] kValue set to the value of the element if it is present.
+/// \return true if the element is present, false if it is a hole.
+static CallResult<bool> readElementForCallback(
+    Runtime &runtime,
+    Handle<JSObject> O,
+    bool fastPath,
+    uint32_t k,
+    Handle<> kHandle,
+    PinnedValue<> &kValue) {
+  if (fastPath) {
+    SmallHermesValue elem = vmcast<JSArray>(O.get())->at(runtime, k);
+    if (LLVM_LIKELY(!elem.isEmpty())) {
+      // Fast path: the element is a plain own data property in the indexed
+      // storage, so no property lookup is needed.
+      kValue = elem.unboxToHV(runtime);
+      return true;
+    }
+  }
+  struct : Locals {
+    PinnedValue<JSObject> descObj;
+    PinnedValue<SymbolID> tmpPropNameStorage;
+  } lv;
+  LocalsRAII lraii{runtime, &lv};
+  ComputedPropertyDescWithSymStorage desc{lv.tmpPropNameStorage};
+  JSObject::getComputedPrimitiveDescriptor(
+      O, runtime, kHandle, lv.descObj, desc);
+  CallResult<PseudoHandle<>> propRes = JSObject::getComputedPropertyValue_RJS(
+      O, runtime, lv.descObj, desc.get(), kHandle);
+  if (LLVM_UNLIKELY(propRes == ExecutionStatus::EXCEPTION)) {
+    return ExecutionStatus::EXCEPTION;
+  }
+  if (LLVM_LIKELY(!(*propRes)->isEmpty())) {
+    kValue = std::move(*propRes);
+    return true;
+  }
+  return false;
+}
+
 inline CallResult<HermesValue> arrayPrototypeForEach(void *, Runtime &runtime) {
   NativeArgs args = runtime.getCurrentFrame().getNativeArgs();
   struct : Locals {
     PinnedValue<JSObject> O;
     PinnedValue<> lenProp;
     PinnedValue<> k;
-    PinnedValue<JSObject> descObj;
-    PinnedValue<SymbolID> tmpPropNameStorage;
+    PinnedValue<> kValue;
   } lv;
   LocalsRAII lraii{runtime, &lv};
 
@@ -2051,28 +2103,29 @@ inline CallResult<HermesValue> arrayPrototypeForEach(void *, Runtime &runtime) {
   // Index to execute the callback on.
   lv.k = HermesValue::encodeTrustedNumberValue(0);
 
+  // Whether elements can be read directly out of the indexed storage; see
+  // readElementForCallback() for why this can be hoisted out of the loop.
+  bool fastPath = false;
+  if (JSArray *arr = dyn_vmcast<JSArray>(lv.O.get()))
+    fastPath = arrayFastPathCheck(runtime, arr, nullptr, (uint32_t)len);
+
   // Loop through and execute the callback on all existing values.
-  // TODO: Implement a fast path for actual arrays.
   auto marker = gcScope.createMarker();
   while (lv.k->getDouble() < len) {
     gcScope.flushToMarker(marker);
 
-    ComputedPropertyDescWithSymStorage desc{lv.tmpPropNameStorage};
-    JSObject::getComputedPrimitiveDescriptor(
-        lv.O, runtime, lv.k, lv.descObj, desc);
-    CallResult<PseudoHandle<>> propRes = JSObject::getComputedPropertyValue_RJS(
-        lv.O, runtime, lv.descObj, desc.get(), lv.k);
-    if (LLVM_UNLIKELY(propRes == ExecutionStatus::EXCEPTION)) {
+    CallResult<bool> kPresent = readElementForCallback(
+        runtime, lv.O, fastPath, (uint32_t)lv.k->getDouble(), lv.k, lv.kValue);
+    if (LLVM_UNLIKELY(kPresent == ExecutionStatus::EXCEPTION)) {
       return ExecutionStatus::EXCEPTION;
     }
-    if (LLVM_LIKELY(!(*propRes)->isEmpty())) {
-      auto kValue = std::move(*propRes);
+    if (*kPresent) {
       if (LLVM_UNLIKELY(
               Callable::executeCall3(
                   callbackFn,
                   runtime,
                   args.getArgHandle(1),
-                  kValue.get(),
+                  lv.kValue.get(),
                   lv.k.get(),
                   lv.O.getHermesValue()) == ExecutionStatus::EXCEPTION)) {
         return ExecutionStatus::EXCEPTION;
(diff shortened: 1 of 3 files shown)


lib/VM/JSLib/Array.cpp
```diff
@@ -2012,14 +2012,66 @@ CallResult<HermesValue> arrayPrototypeSort(void *, Runtime &runtime) {
...
+///   the loop.
+/// \param k the current index.
+/// \param kHandle a Handle containing \p k encoded as a number.
+/// \param[out] kValue set to the value of the element if it is present.
+/// \return true if the element is present, false if it is a hole.
+static CallResult<bool> readElementForCallback(
+    Runtime &runtime,
+    Handle<JSObject> O,
+    bool fastPath,
+    uint32_t k,
```
lavenzg: This parameter seems unnecessary. You are using kHandle to derive it at the call site anyway. Just do it in this function.

lib/VM/JSLib/Array.cpp
```diff
@@ -2012,14 +2012,66 @@ CallResult<HermesValue> arrayPrototypeSort(void *, Runtime &runtime) {
   return lv.O.getHermesValue();
 }
 
+/// Read the element at index \p k of \p O inside an iteration loop that
```
lavenzg: This doc comment is confusing. We should just say this is to load value for property kHandle, and when fastPath is true, O must be a JSArray so we load it from the underlying ArrayStorage, otherwise, go through the normal property loading path. We can mention that this is currently used by methods forEach/map/filter. But this helper method is more general (you could remove `ForCallback` from the function name).

lib/VM/JSLib/Array.cpp
```diff
@@ -2457,13 +2510,33 @@ CallResult<HermesValue> arrayPrototypeSlice(void *, Runtime &runtime) {
...
+  // because the element count would no longer match \c len.
+  // Note that JSArray::create has already set the length of \c A, and holes
+  // (empty values) are copied like ordinary values, which is unobservable
+  // because the fast path check has ensured there are no index-like
+  // properties in the prototype chain.
+  if (JSArray *arr = dyn_vmcast<JSArray>(lv.O.get());
+      arr && arrayFastPathCheck(runtime, arr, nullptr, (uint32_t)len)) {
+    if (count32 > 0) {
+      NoAllocScope noAlloc{runtime};
+      uint32_t start = (uint32_t)k;
```
lavenzg: Coud `k` be out of range of uint32_t? It seems we don't have a check for `len` value range in this function.

lib/VM/JSLib/Array.cpp
```diff
@@ -2457,13 +2510,33 @@ CallResult<HermesValue> arrayPrototypeSlice(void *, Runtime &runtime) {
...
+      auto *aStorage = lv.A->getIndexedStorageUnsafe(runtime);
+      auto *oStorage = arr->getIndexedStorageUnsafe(runtime);
+      for (uint32_t j = 0; j < count32; ++j) {
```
lavenzg: Use GCHVType::copy()?

lib/VM/JSLib/Array.cpp
```diff
@@ -3221,24 +3338,49 @@ CallResult<HermesValue> arrayPrototypeShift(void *, Runtime &runtime) {
...
   LocalsRAII lraii{runtime, &lv};
 
-  GCScope gcScope(runtime);
-  auto objRes = toObject(runtime, args.getThisHandle());
-  if (LLVM_UNLIKELY(objRes == ExecutionStatus::EXCEPTION)) {
-    return ExecutionStatus::EXCEPTION;
-  }
-  lv.O.castAndSetHermesValue<JSObject>(*objRes);
+  // Ensure the fast path does not leak any handles.
+  NoLeakHandleScope noLeaks{runtime};
```
lavenzg: Don't you need to move this down into the fast path branch?

lib/VM/JSLib/Array.cpp
```diff
@@ -3590,6 +3730,66 @@ indexOfHelper(Runtime &runtime, NativeArgs args, const bool reverse) {
   return HermesValue::encodeTrustedNumberValue(-1);
 }
 
+/// Fast path for Array.prototype.unshift() when the array is a normal array.
+/// \pre \p arr is an array with fast index properties and there are no
```
lavenzg, line 3818: We usually put pre-condition after parameters doc comment.

lib/VM/JSLib/Array.cpp
```diff
@@ -3603,26 +3803,53 @@ CallResult<HermesValue> arrayPrototypeUnshift(void *, Runtime &runtime) {
...
   LocalsRAII lraii{runtime, &lv};
-  GCScope gcScope(runtime);
 
-  auto objRes = toObject(runtime, args.getThisHandle());
-  if (LLVM_UNLIKELY(objRes == ExecutionStatus::EXCEPTION)) {
-    return ExecutionStatus::EXCEPTION;
-  }
-  lv.O.castAndSetHermesValue<JSObject>(*objRes);
+  // Ensure the fast path does not leak any handles.
+  NoLeakHandleScope noLeaks{runtime};
```
lavenzg: Move into below fast path branch.

test/hermes/array-fastpath-mutation.js
```diff
@@ -0,0 +1,192 @@
...
+a = [1, 2, 3, 4, 5];
+s = a.slice({valueOf: function() { a.length = 2; return 0; }});
+print(s.length, s[0], s[1], 2 in s);
+// CHECK-NEXT: 5 1 2 false
+
+print('forEach');
+// CHECK-LABEL: forEach
+a = [10, 20, 30];
+var acc = [];
+a.forEach(function(v, i) { acc.push(v, i); });
```
lavenzg: We need to cover a few cases (unless you have them and I missed):
1. The callback throws.
2. The callback triggers GC (using gc() method).
3. The callback shrinks the array.

lavenzg (changes requested):
Thanks for the PR. Left a few comments.

zhelezkov:
hey @lavenzg, thanks for your review! I've addressed all your comments, short summary:

- `readElementForCallback` is now an `ElementReader`, the check happens once there
  instead of per element. tried to keep it clean and consistent with the rest of the
  codebase, curious if you have any feedback
- added a few asserts, and a `canReadElementsFast()` helper that also covers the `k`
  range you asked about
- switched to `GCHVType`, I needed to make it public, is it ok?
- fixed `NoLeakHandleScope` usage, it's inside the fast path branches now
- updated comments, hopefully I didn't miss any!
- tests: throwing callbacks, gc, shrinking and a few other edge cases. renamed the
  file to `array-fastpath.js`, it's not only about mutation anymore

also I added `every/some`, and `splice` got touched while switching to `GCHVType`

Hopefully it's not too much scope creep for this PR, lmk if you have any
questions/thoughts, happy to address any comments!

PR #2095, opened 2026-07-02
Fix Proxy [[OwnPropertyKeys]] check for integer-index keys (#1609)

### LINKED ISSUE #1609: TypeError: ownKeys target is non-extensible but key is missing from trap result error when using numeric string keys on Hermes
## Bug Description

When using Valtio with Hermes in React Native, an error is thrown if the proxy object contains numeric string keys. For example, using a key like '12345' results in the following error:

```vbnet
TypeError: ownKeys target is non-extensible but key is missing from trap result, js engine: hermes
```

I've confirmed that using non-numeric string keys (e.g. text) avoids the error.

- [ ] I have run `gradle clean` and confirmed this bug does not occur with JSC
- [ ] The issue is reproducible with the latest version of React Native.

Hermes git revision (if applicable):
React Native version: 0.76.5 (new arch is `false`)
OS: iOS, Android
Platform (most likely one of arm64-v8a, armeabi-v7a, x86, x86_64):  arm64

## Steps To Reproduce

1. Create a Valtio proxy object with a numeric string as key:
```tsx
import { useSnapshot, proxy } from 'valtio';
const state = proxy({ '12345': 'hello' });
```

2.In a ReactNative component, take a snapshot and log it when a button is pressed:
```tsx
export const MyScreen: React.FC = () => {
  const snap = useSnapshot(state);
  return (
    <Button
      onPress={() => {
        // Triggering iteration on the snapshot throws the error:
        Object.entries(snap).forEach(([key, value]) => {
          console.log(key, value);
        });
      }}
    >
      button for test
    </Button>
  );
}
```
3. Run the application on a device/emulator with Hermes enabled.
4. Press the button. The error occurs:
```vbnet
TypeError: ownKeys target is non-extensible but key is missing from trap result, js engine: hermes
```

## The Expected Behavior

Using a numeric string as a key in a Valtio proxy should not lead to an error when accessing the snapshot (e.g., via Object.entries or direct logging). The behavior should be consistent with proxy objects that use non-numeric keys.

Additional context:

- Replacing the key with a non-numeric string (for example, proxy({ text: 'hello' })) avoids the error.
- A patch from [this Valtio PR](https://github.com/pmndrs/valtio/pull/752/files#diff-b68a2cbc8606fe37f9c2c72867a87ff50a11c2b38a6d9de45a94281f02bf0a4eL149) appears to work around the issue, suggesting the problem may be related to how Hermes handles Proxy traps.
- A similar issue is discussed in the [Hermes issue #1063](https://github.com/facebook/hermes/issues/1063), though the root cause has not yet been identified.

Any guidance or fixes for this issue would be greatly appreciated. Thank you for your help!

tangtaizong666:
## Summary

`Reflect.ownKeys` / `Object.keys` / `Object.getOwnPropertyNames` on a `Proxy`
whose target is **non-extensible** and has an **integer-index** own key (e.g.
`"12345"`) throws

```
TypeError: ownKeys target is non-extensible but key is missing from trap result
```

even when the trap faithfully returns the target's own keys. This breaks common
state libraries such as Valtio (see #1609), where a frozen snapshot object with
numeric string keys can no longer be enumerated.

**Root cause.** `JSProxy::ownPropertyKeys` enforces the ES `[[OwnPropertyKeys]]`
invariant (steps 19 and 21) by checking that every own key of the target is
present in the trap result using `SameValue`. Hermes represents integer-index
own-property keys internally as **numbers** (`JSObject::getOwnPropertyKeys`
returns `12345` as a Number), whereas the trap result is validated to contain
only Strings and Symbols and therefore holds the key as the String `"12345"`.
`SameValue(12345, "12345")` is always `false`, so the invariant check fails and
throws. The comparison only runs for non-extensible targets, which is why the
bug only surfaces once the target is frozen / `preventExtensions`'d.

**Fix.** Before the invariant comparisons, normalize any numeric index keys in
`targetKeys` to their canonical string form — the same conversion that for-in
enumeration already performs for numeric array indices
(`Interpreter-slowpaths.cpp`, "We must return the property as a string"). The
returned key list is unchanged (it still comes from the trap result), so this
only removes the spurious invariant failure.

## Test Plan

New regression test `test/hermes/regress-proxy-ownkeys-numeric-key.js` covering:

- an integer-index key on a `preventExtensions` target (configurable-keys loop),
- integer-index keys on a `Object.freeze`d target (non-configurable-keys loop),
- mixed integer/string key ordering, and
- integer-index keys alongside symbols.

```
$ ./build-rel/bin/hermes -Xes6-proxy -non-strict -O -target=HBC \
    test/hermes/regress-proxy-ownkeys-numeric-key.js
case1: 12345
case2: 0,42
case3: 1,2,b,a
case4: 7,Symbol(s)
```

The test passes with this change and fails on the unpatched engine with the
reported `TypeError`. Existing `test/hermes/proxy.js` continues to pass.

Fixes #1609

### DIFF
--- a/lib/VM/JSProxy.cpp
+++ b/lib/VM/JSProxy.cpp
@@ -12,6 +12,7 @@
 #include "hermes/VM/JSArray.h"
 #include "hermes/VM/JSCallableProxy.h"
 #include "hermes/VM/JSMapImpl.h"
+#include "hermes/VM/Operations.h"
 #include "hermes/VM/PropertyAccessor.h"
 
 #include "llvh/ADT/SetVector.h"
@@ -1495,6 +1496,40 @@ CallResult<PseudoHandle<JSArray>> JSProxy::ownPropertyKeys(
   if (*extensibleRes && nonConfigurable.empty()) {
     //   a. Return trapResult.
     return filterKeys(selfHandle, trapResult, runtime, okFlags);
+  }
+  // Hermes represents integer-index own-property keys internally as numbers,
+  // but [[OwnPropertyKeys]] is specified to return only Strings and Symbols
+  // (see the assert on step 12 above), and the trap result was already
+  // validated to contain only Strings and Symbols. Convert any numeric index
+  // keys in targetKeys to their canonical string form so the SameValue
+  // comparisons below behave correctly; otherwise an integer-index key such as
+  // "12345" on a non-extensible target would never match its string form in
+  // the trap result. See https://github.com/facebook/hermes/issues/1609.
+  {
+    GCScopeMarkerRAII normalizeMarker{runtime};
+    for (uint32_t i = 0, len = JSArray::getLength(*targetKeys, runtime);
+         i < len;
+         ++i) {
+      HermesValue key = targetKeys->at(runtime, i).unboxToHV(runtime);
+      if (!key.isNumber()) {
+        continue;
+      }
+      normalizeMarker.flush();
+      CallResult<PseudoHandle<StringPrimitive>> keyStrRes =
+          numberToStringPrimitive(runtime, key.getNumber());
+      if (LLVM_UNLIKELY(keyStrRes == ExecutionStatus::EXCEPTION)) {
+        return ExecutionStatus::EXCEPTION;
+      }
+      if (LLVM_UNLIKELY(
+              JSArray::setElementAt(
+                  targetKeys,
+                  runtime,
+                  i,
+                  runtime.makeHandle(std::move(*keyStrRes))) ==
+              ExecutionStatus::EXCEPTION)) {
+        return ExecutionStatus::EXCEPTION;
+      }
+    }
   }
   // 18. Let uncheckedResultKeys be a new List which is a copy of trapResult.
   // 19. For each key that is an element of targetNonconfigurableKeys, do
--- a/test/hermes/regress-proxy-ownkeys-numeric-key.js
+++ b/test/hermes/regress-proxy-ownkeys-numeric-key.js
@@ -1,0 +1,58 @@
+/**
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// RUN: %hermes -Xes6-proxy -non-strict -O -target=HBC %s | %FileCheck --match-full-lines %s
+
+// Regression test for https://github.com/facebook/hermes/issues/1609
+//
+// A Proxy [[OwnPropertyKeys]] over a non-extensible target used to throw
+// "ownKeys target is non-extensible but key is missing from trap result"
+// whenever the target had an integer-index own key (e.g. "12345"). Hermes
+// represents integer-index keys internally as numbers, but the ownKeys trap
+// result holds them (per spec) as strings, so the SameValue comparison in the
+// non-extensible invariant check never matched.
+
+function ownKeysTrap(t) {
+  return Reflect.ownKeys(t);
+}
+
+// Case 1: preventExtensions with a configurable integer-index key.
+// Exercises the "target configurable keys" invariant loop.
+var t1 = {};
+Object.defineProperty(t1, '12345', {
+  value: 1,
+  enumerable: true,
+  configurable: true,
+  writable: true,
+});
+Object.preventExtensions(t1);
+var p1 = new Proxy(t1, {ownKeys: ownKeysTrap});
+print('case1:', Object.keys(p1).join(','));
+// CHECK: case1: 12345
+
+// Case 2: frozen (non-configurable) integer-index keys.
+// Exercises the "target non-configurable keys" invariant loop.
+var t2 = Object.freeze({0: 'a', 42: 'b'});
+var p2 = new Proxy(t2, {ownKeys: ownKeysTrap});
+print('case2:', Object.getOwnPropertyNames(p2).join(','));
+// CHECK-NEXT: case2: 0,42
+
+// Case 3: mixed integer-index and string keys, frozen; spec ordering is
+// ascending integer indices first, then string keys in insertion order.
+var t3 = Object.freeze({b: 1, 2: 2, a: 3, 1: 4});
+var p3 = new Proxy(t3, {ownKeys: ownKeysTrap});
+print('case3:', Reflect.ownKeys(p3).join(','));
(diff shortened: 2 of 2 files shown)


xhon-pelushi (commented):
Built `static_h` (`e9edc8b`) with and without this patch and checked it against V8. The diagnosis is right, the fix is right, and it's a bigger win than the description claims. Three suggestions below, none blocking.

## Root cause confirmed in source

- `JSObject.cpp:507` encodes integer-index own keys with `HermesValue::encodeTrustedNumberValue`, so `targetKeys` really does hold `12345` as a Number.
- `isSameValue` (`Operations.cpp:125`) returns `false` on the `getTag()` mismatch before it ever reaches the deep string comparison, so `SameValue(12345, "12345")` is false as described.
- `inTrapResult` (`JSProxy.cpp:1507`) is the only place in `JSProxy.cpp` that compares a *key*. The other three `isSameValue` calls — lines 201, 912, 1047 — compare descriptor and trap *values*, not keys. So this is the only affected comparison and the fix doesn't need a sibling.

Two things I checked because they'd have made the patch wrong:

- **Mutating `targetKeys` in place is safe.** `JSObject::getOwnPropertyKeys` allocates a fresh array per call, and `targetKeys` is never returned — every return path hands back `trapResult` or the separate step-6a array. The normalized list is only ever used for the comparisons.
- **The approach is already the in-repo idiom**, not a new invention. `getOwnPropertyKeysAsStrings` (`lib/VM/JSLib/Object.cpp:490`) is the same in-place number→string `JSArray::setElementAt` loop with the same GC-marker flushing, and the single-element version appears at `Interpreter-slowpaths.cpp:570` and `StaticH.cpp:2365`. Placing the block after the step-17 early return also keeps it off the common all-extensible path.

## It reproduces, and the fix holds

Unpatched, the new test fails exactly as claimed, and both invariant messages are reachable depending on the shape:

```
$ ./build-rel/bin/hermes -Xes6-proxy -non-strict -O -target=HBC test/hermes/regress-proxy-ownkeys-numeric-key.js
Uncaught TypeError: ownKeys target is non-extensible but key is missing from trap result
    at keys (native)
    at global (test/hermes/regress-proxy-ownkeys-numeric-key.js:34:28)

$ ./build-rel/bin/hermes -Xes6-proxy i1609.js          # Object.freeze({123:'x'}) behind a passthrough trap
Uncaught TypeError: ownKeys target key is non-configurable but not present in trap result
```

Patched, the test prints the expected four lines and the issue repro returns `["123"]`, matching V8.

**Regression sweep.** I ran all 498 `test/hermes/*.js` under one uniform invocation on both builds and diffed. Six files differed; five of them (`array-literal-large`, `date-constructor`, `narrow-trusted-regress`, `put-to-transient`, `regress-error-stack-native-stack-overflow`) also differ **base-against-base** on a rerun, so they're nondeterministic and not attributable. The only real change is `regress-proxy-ownkeys-numeric-key.js` going 1 → 0. `test/hermes/proxy.js` is byte-identical.

**Differential test against V8.** 27 cases covering index keys, ordering, symbols, boundary indices, non-canonical numeric-looking keys, the negative invariant cases, and the consumers (`Object.keys`, `getOwnPropertyNames`, `JSON.stringify`, spread, `for-in`, nested proxies), run on node 18 and both Hermes builds:

| | disagreements with V8 |
| --- | --- |
| Hermes before | **15 of 27** |
| Hermes after | **0 of 27** |

Importantly the five negative cases still throw — trap omitting the index, returning the wrong index, returning an extra key, returning a Number, and returning duplicates all still raise `TypeError`, same as V8. The fix removes the spurious failures without loosening the invariant.

**Stress.** 5000 integer-index keys on a frozen target, `Reflect.ownKeys` 50 times: no crash or handle-scope trouble, and the result matches V8 (`len=5000`, first/last both `string`).

## 1. The description under-sells the scope: it isn't only non-extensible targets

> The comparison only runs for non-extensible targets, which is why the bug only surfaces once the target is frozen / `preventExtensions`'d.

Step 19's loop runs whenever `targetNonconfigurableKeys` is non-empty, independent of extensibility — `if (*extensibleRes) return` is step 20, *after* it. So a perfectly extensible object with one non-configurable integer-index property also throws:

```js
var o = {};
Object.defineProperty(o, '5', {value: 1, configurable: false, enumerable: true});
Reflect.ownKeys(new Proxy(o, {ownKeys: t => Reflect.ownKeys(t)}));
// before: TypeError: ownKeys target key is non-configurable but not present in trap result
// after:  ["5"]     (V8: ["5"])
```

That's also the message @tmikov got in [the first repro on #1609](https://github.com/facebook/hermes/issues/1609#issuecomment-2632964757), so it's the path the original reporter hit. Worth fixing the wording since it makes the change look narrower than it is.

The other case I'd highlight is a **frozen array**, which I'd guess is at least as common in the wild as numeric string keys:

```js
Reflect.ownKeys(new Proxy(Object.freeze([1,2,3]), {ownKeys: t => Reflect.ownKeys(t)}));
// before: TypeError    after: ["0","1","2","length"]   (V8: same)
```

## 2. Two test cases worth adding

The four cases in the new test all have `extensibleTarget === false`, so they only ever exercise the step-21 loop plus step 19. Neither of the shapes above is covered:

- **extensible target with a non-configurable integer-index key** — the only case where step 19 fires and step 20 returns before step 21, i.e. the one path where the normalization matters but step 21 never runs.
- **`Object.freeze([1,2,3])` behind a proxy** — index keys alongside `"length"`, and the likely real-world shape.

## 3. Minor: the loop converts more keys than get compared

The normalization walks every numeric entry in `targetKeys`, but when `extensibleTarget` is true only the `nonConfigurable` entries are ever passed to `inTrapResult` — step 20 returns before the step-21 loop. So an extensible target with one non-configurable index key among a thousand index keys pays a thousand `numberToStringPrimitive` allocations to perform one comparison. Converting inside `inTrapResult`, or restricting the loop to the indices that will actually be compared, would avoid that. Very much a nit given Proxy's overall cost here, but it's nearly free to fix.

## What I did not verify

`FileCheck` isn't built in my tree, so the regression sweep used one uniform flag set across `test/hermes/*.js` rather than each test's own `RUN` line and compared base-vs-patched output rather than checking `CHECK:` directives — good for spotting differences caused by this patch, not a substitute for `check-hermes`. I also only exercised the `-target=HBC` CLI path; `JSProxy::ownPropertyKeys` is the single implementation (reached from `JSObject.cpp:432`), so the SH path should be covered too, but I didn't run it, and nothing was tested on-device under React Native.

PR #2099, opened 2026-07-07
Fix dry-run npm publish colliding with already-published version

cipolleschi:
## Context

The "RN Build Static Hermes" workflow runs on every push to `static_h`, but the `publish / publish` job fails at the **Publish to npm** step. We don't want to publish from `static_h` — we only want the dry-run to validate that the whole pipeline works.

## Findings

1. **The job *is* running in dry-run mode.** `set_release_type` forces `RELEASE_TYPE=dry-run` on `static_h`, and the npm log confirms it (`npm warn publish ... (dry-run)`). So the failure is **not** an accidental real release.

2. **The dry-run itself fails on a version collision:**
   ```
   npm error You cannot publish over the previously published versions: 250829098.0.0.
   ```
   `npm publish --dry-run` still performs a registry pre-check, and `250829098.0.0` (the raw `package.json` version) is already published.

3. **Root cause:** `publish-npm.js` runs `npm publish --dry-run` against `npm/hermes-compiler/package.json` **without ever setting the version**. `getVersion('dry-run')` produces a unique, per-commit version (`${mainVersion}-${shortCommit}`) precisely so dry-runs don't collide — but it was never written to `package.json` before publishing. The script even imported `updatePackageJsonVersion` from `version-utils`, but that function was **never defined or exported** there.

## Fix

- Add `updatePackageJsonVersion(version)` to `version-utils.js` (regex-replace on the version field, mirroring the existing `updateGradlePropertiesFile`, so formatting is preserved) and export it.
- In `publishNpm()`, **only in dry-run mode**, set the version to `getVersion('dry-run')` (the unique per-commit value) before `npm publish --dry-run`.

Result:
- **dry-run** → `package.json` gets `250829098.0.0-<sha>`, which is not published, so `npm publish --dry-run` validates packaging end-to-end and passes.
- **release** → `package.json` is left untouched; we publish exactly the version declared in the file. Behavior unchanged.

## Test Plan
- Validated the version-replacement produces valid JSON, preserves formatting, and changes only the version line.
- `node --check` passes on both edited files.
- CI on this PR / a subsequent push to `static_h` should show `publish / publish` completing the dry-run **Publish to npm** step successfully.

### DIFF
--- a/utils/scripts/hermes/publish-npm.js
+++ b/utils/scripts/hermes/publish-npm.js
@@ -70,16 +70,16 @@ async function publishNpm(
 
   if (buildType === 'dry-run') {
     tagFlag = ` --dry-run`;
+    // For a dry run, publish a unique per-commit version
+    // (`${mainVersion}-${shortCommit}`) so the `npm publish --dry-run`
+    // validation does not collide with an already-published version (npm
+    // performs this check even with `--dry-run`). For a real release we keep
+    // the version declared in package.json untouched.
+    const version = await getVersion(buildType);
+    await updatePackageJsonVersion(version);
   } else if (buildType === 'release') {
     tagFlag = ` --tag latest-v1`;
   }
-
-  // Set the version we are about to publish. For a dry-run this is
-  // `${mainVersion}-${shortCommit}`, which is unique per commit and therefore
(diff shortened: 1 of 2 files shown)


cortinico (approved):

PR #2102, opened 2026-07-10
Bump hermes-compiler version to 260318099.0.1

cipolleschi:
## Summary

We released 260318099.0.0 to npm and Maven central, so we now need to bump this version or future publishing for React Native will fail.

## Test Plan
N/A


cortinico (approved):

PR #2110, opened 2026-07-23
Bump hermes-compiler version to 260318099.0.2

cipolleschi:
## Summary
I'm releasing a new version (260318099.0.1) of the Hermes engine to have React Native on main run with this version.
The source code for Hermes should always track the next version, hence the version bump.

## Test Plan
N/A


cortinico (approved):

PR #2113, opened 2026-07-24
Rebalance external memory accounting in OrderedHashMapBase::clear()

### LINKED ISSUE #2111: OrderedHashMapBase::clear() strands external-memory credit; per-frame Set/Map clear()+refill OOMs long sessions
## Summary
`OrderedHashMapBase::clear()` does not rebalance the external-memory accounting that `initializeStorage()`/`rehash()` maintain for `hashTable_`. Every `Set`/`Map` `clear()` + refill cycle permanently strands external-memory credit, so long-running apps with a per-frame clear/refill pattern watch `externalBytes` grow without bound and eventually abort with `LLVM ERROR: OOM ... Max heap size was exceeded` while holding a small, constant live set.

## Version
Static Hermes line, observed on `260318099.0.0-stable` (also matches current sources). Embedded via JSI on macOS arm64, Hades GC, 1GB `maxHeapSize`.

## The imbalance (lib/VM/OrderedHashMap.cpp)
- `initializeStorage()` credits the initial table: `runtime.getHeap().creditExternalMemory(*self, kInitialCapacity * sizeof(uint32_t));`
- `rehash()` credits the new table and debits the old one (`debitExternalMemory(*self, oldSizeInBytes)`), so net credit tracks `capacity_ * sizeof(uint32_t)`.
- `clear()` does neither:

```cpp
// Clear the hash table.
self->hashTable_ = std::vector<uint32_t>();          // frees capacity_*4 bytes, no debit
// Resize the hash table to the initial size.
self->hashTable_.resize(kInitialCapacity, ...);      // reallocates 64 bytes, no credit
self->capacity_ = kInitialCapacity;
```

After `clear()`, the cell keeps `capacity_ * 4` bytes of credit while really holding 64 bytes. Refilling re-runs the rehash ladder, crediting the growth again. Each clear-at-capacity-C/refill cycle therefore strands `C*4 - kInitialCapacity*4` bytes of phantom credit (64-byte quanta for the common small-set sizes).

## How we hit it
A three.js WebGPU game loop under Hermes cleared and refilled a small per-object `Set` every frame (`UniformsGroup`). `HermesInternal.getInstrumentedStats().js_externalBytes` climbed ~360KB/s (~14KB per render call; deltas always exact 64-byte multiples, e.g. 14,208 = 222×64) with a flat object population, and the process aborted at the 1GB cap after ~35 minutes. Replacing the per-cycle `Set` with boolean flags on pooled records eliminated the growth (flat over a 65-minute soak). Young-GC pacing also degraded as the counter inflated, so frame times crept up well before the abort.

## Repro sketch
```js
const s = new Set();
const stats = () => HermesInternal.getInstrumentedStats().js_externalBytes;
const before = stats();
for (let n = 0; n < 100000; n++) {
  s.clear();
  for (let i = 0; i < 40; i++) s.add(i);   // capacity reaches 64
}
const after = stats();   // grows by ~19MB (192B per cycle) and never returns to baseline
```

## Suggested fix
In `clear()`, mirror the rehash bookkeeping:

```cpp
uint32_t oldSizeInBytes = self->hashTable_.size() * sizeof(uint32_t);
self->hashTable_ = std::vector<uint32_t>();
self->hashTable_.resize(kInitialCapacity, kHashTableElementUnused);
runtime.getHeap().debitExternalMemory(*self, oldSizeInBytes);
runtime.getHeap().creditExternalMemory(*self, kInitialCapacity * sizeof(uint32_t));
```

(If the finalizer debits only the current `hashTable_` size, the stranded credit also survives the cell's death, which matches the monotonic growth we observed across collections.)

vernon99:
## Summary
`OrderedHashMapBase::clear()` discards the grown hash table without debiting its external-memory credit and reinitializes to `kInitialCapacity` without crediting it. `initializeStorage()` and `rehash()` maintain the invariant that the cell's external credit equals `hashTable_.size() * sizeof(uint32_t)`, and `cleanUp()` debits exactly the current size — so the credit stranded by `clear()` is never returned, not even when the cell dies.

Every `Set`/`Map` `clear()` + refill cycle therefore strands `(capacity - kInitialCapacity) * sizeof(uint32_t)` bytes of phantom credit. Long-running embedders with per-frame clear/refill patterns (e.g. three.js render loops, which clear a per-object `Set` in every uniform upload cycle) watch `js_externalBytes` climb to the configured heap limit and abort with a spurious OOM. We measured ~360KB/s of phantom growth (~14KB per render call, always exact 64-byte multiples) in a WebGPU game loop on Hermes, with an abort at a 1GB heap cap after ~35 minutes, while the live set stayed constant.

Details and repro: #2111

## Fix
Mirror the `initializeStorage()`/`rehash()` bookkeeping in `clear()`: debit the discarded table's bytes, credit the reinitialized table's bytes.

## Validation
In our embedder (JSI, Hades GC), 20,000 cycles of `set.clear()` + refilling 48 entries grew `js_externalBytes` by ~3.8MB before this change and by ~0 after it; a previously-OOMing 60+ minute game session now holds external memory flat. Happy to add a VMRuntime unit test if you'd like one — pointer to the preferred test file welcome.

Fixes #2111

### DIFF
--- a/lib/VM/OrderedHashMap.cpp
+++ b/lib/VM/OrderedHashMap.cpp
@@ -215,8 +215,6 @@ ExecutionStatus OrderedHashMapBase<BucketType, Derived>::rehash(
   // Create a new hash table.
   std::vector<uint32_t> newHashTable;
   newHashTable.resize(*newCapacity, kHashTableElementUnused);
-  runtime.getHeap().creditExternalMemory(
-      *self, *newCapacity * sizeof(uint32_t));
 
   // Create a new data table.
   auto dataTableRes = StorageType::create(
@@ -275,9 +273,19 @@ ExecutionStatus OrderedHashMapBase<BucketType, Derived>::rehash(
   rawSelf->updateIteratorIndicesForRehash(runtime);
 
   rawSelf->deletedCount_ = 0;
-  uint32_t oldSizeInBytes = rawSelf->hashTable_.size() * sizeof(uint32_t);
+  // The external memory credit tracks hashTable_.size() * sizeof(uint32_t).
+  // The new table may be larger or smaller than the old one, so settle the
+  // difference in whichever direction it went.
+  const uint32_t oldSizeInBytes = rawSelf->hashTable_.size() * sizeof(uint32_t);
+  const uint32_t newSizeInBytes = newHashTable.size() * sizeof(uint32_t);
   rawSelf->hashTable_ = std::move(newHashTable);
-  runtime.getHeap().debitExternalMemory(*self, oldSizeInBytes);
+  if (newSizeInBytes >= oldSizeInBytes) {
+    runtime.getHeap().creditExternalMemory(
+        *self, newSizeInBytes - oldSizeInBytes);
+  } else {
+    runtime.getHeap().debitExternalMemory(
+        *self, oldSizeInBytes - newSizeInBytes);
+  }
   rawSelf->dataTable_.setNonNull(runtime, *lv.newDataTable, runtime.getHeap());
   return ExecutionStatus::RETURNED;
 }
@@ -671,11 +679,19 @@ ExecutionStatus OrderedHashMapBase<BucketType, Derived>::clear(
     return ExecutionStatus::RETURNED;
   }
 
-  // Clear the hash table.
+  // Clear the hash table. The external memory credit tracks
+  // hashTable_.size() * sizeof(uint32_t), so debit the difference between the
+  // discarded table and the reinitialized one.
+  const uint32_t oldSizeInBytes = self->hashTable_.size() * sizeof(uint32_t);
   self->hashTable_ = std::vector<uint32_t>();
   // Resize the hash table to the initial size.
   self->hashTable_.resize(kInitialCapacity, kHashTableElementUnused);
   self->capacity_ = kInitialCapacity;
+  constexpr uint32_t newSizeInBytes = kInitialCapacity * sizeof(uint32_t);
+  assert(
+      oldSizeInBytes >= newSizeInBytes &&
+      "Capacity never shrinks below kInitialCapacity");
+  runtime.getHeap().debitExternalMemory(*self, oldSizeInBytes - newSizeInBytes);
 
   // Resize the data table back to 0.
   self->dataTable_.getNonNull(runtime)->clear(runtime);


lib/VM/OrderedHashMap.cpp
```diff
@@ -671,11 +671,19 @@ ExecutionStatus OrderedHashMapBase<BucketType, Derived>::clear(
...
+  // and re-credit the initial size so the credit keeps matching
+  // hashTable_.size() * sizeof(uint32_t): initializeStorage()/rehash()
+  // maintain that invariant and cleanUp() debits exactly the current size.
+  const uint32_t oldHashTableSizeInBytes =
+      self->hashTable_.size() * sizeof(uint32_t);
   self->hashTable_ = std::vector<uint32_t>();
   // Resize the hash table to the initial size.
   self->hashTable_.resize(kInitialCapacity, kHashTableElementUnused);
   self->capacity_ = kInitialCapacity;
+  runtime.getHeap().debitExternalMemory(*self, oldHashTableSizeInBytes);
```
lavenzg: Let's compute the size diff and directly debit back the diff value (old size should be larger or equals to the init size). We probably should do the same for rehash() but there we need to check if it's a debit or credit.

lavenzg (changes requested):

lib/VM/OrderedHashMap.cpp
vernon99: Done, thanks — updated in both places.

`clear()` now issues a single `debitExternalMemory()` of the difference, with an assert for the "old size >= init size" invariant that `shouldShrink()` / `checkedNextCapacity()` guarantee.

`rehash()` does the same, with the direction check you called out. Getting one net op there meant moving it: nothing can be netted at the old credit site because the old table is still live, so the single credit-or-debit now lands where the new table is installed. That has a side effect worth flagging explicitly, since it goes past what you asked for — the old code credited `newCapacity * sizeof(uint32_t)` *before* `StorageType::create()`, so a failure there returned `EXCEPTION` with the credit already applied and the new table thrown away. The net form can't strand anything on that path. It also stops the credit/debit pair from splitting across the young and old generation counters if a GC promotes the cell in between. Happy to drop that part back to a purely local change if you'd rather keep this PR narrow.

Measured both, reading `js_externalBytes` from `HermesInternal.getInstrumentedStats()` on a long-lived `Set` (promoted to the old gen — young-gen external bytes get recomputed per collection, which masks this entirely) refilled to 48 entries and cleared:

| | per `clear()` | over 320k clears |
| --- | --- | --- |
| before | +448 B | +143 MB |
| after | ~0 | flat |

448 is exactly `(128 - kInitialCapacity) * sizeof(uint32_t)`. A grow/shrink workload that never calls `clear()` reports a delta of 0 both before and after, so the `rehash()` rework is accounting-neutral on the normal path.

One suggestion, for a separate PR rather than this one. The invariant "credit == `hashTable_.size() * sizeof(uint32_t)`" is currently open-coded at four sites — `initializeStorage()`, `rehash()`, `clear()`, and `cleanUp()` — and that restatement is what let `clear()` drift in the first place; this fix corrects two of them but adds two more restatements. A single mutator owning it would make the drift unrepresentable:

```c++
  /// Install \p newHashTable, keeping the cell's external memory credit equal
  /// to hashTable_.size() * sizeof(uint32_t).
  void setHashTable(Runtime &runtime, std::vector<uint32_t> &&newHashTable);
```

`rehash()` becomes `setHashTable(runtime, std::move(newHashTable))`, `clear()` and `initializeStorage()` pass a fresh `kInitialCapacity` table, and `cleanUp()` stays as the final debit to zero. Net smaller than what is here now. Glad to send it as a follow-up if you want it — I left it out of this PR to keep the fix reviewable.

Also still happy to add a VMRuntime unit test; a pointer to the preferred file would help.

lavenzg: You could just add a JS test for this regression. Assert that the external bytes is reset back after loops of clear().

PR #2114, opened 2026-07-25
fix: IdentifierHashTable growing indefinitely with deleted entries

tjzel:
## Summary

`IdentifierHashTable` capacity grows without bound when its nonEmpty entries count is nearing the capacity. However, this also happens when most of the entries are already invalid.

This PR makes the table rehash in place when the table is dominated by invalid entries (more than 75% entries are invalid).

## Test plan

~30-line standalone script, no dependencies — run with the `hermes` CLI and watch process RSS:

```js
const sink = {};
let interned = 0;
for (let round = 0; round < 240; round++) {
  for (let j = 0; j < 200000; j++) {
    const s = "unique_key_" + interned++ + "_suffix_padding";
    if (s in sink) throw new Error("unreachable");
    const garbage = { a: j, b: [j, j + 1], c: "g" + (j & 1023) }; // drive GC
    if (garbage.a === -1) print(garbage.b, garbage.c);
  }
  const st = HermesInternal.getInstrumentedStats();
  print(round, st.js_heapSize ?? st.heapSize, st.js_numGCs ?? st.numGCs);
}
```

48M unique keys, macOS host CLI built from the same source, RSS sampled at 0.5 s:

| binary | RSS peak | RSS final | JS heap (self-reported) |
|---|---|---|---|
| before | 223 MB | 159 MB | 12.6 MB flat, 3908 GCs |
| after | 68 MB | **36 MB** | 12.6 MB flat, 3916 GCs |

Real-world instance: a React Native app animating interpolated color strings (a library probes a color-keyword dictionary with each animated `hsl(...)` string ≈ 4k unique interned strings/s) grew from 81 MB to 333 MB over 6 hours through seven table doublings (8→16→…→256 MB tables).

### DIFF
--- a/lib/VM/detail/IdentifierHashTable.cpp
+++ b/lib/VM/detail/IdentifierHashTable.cpp
@@ -125,7 +125,18 @@ void IdentifierHashTable::insert(uint32_t idx, SymbolID id) {
   ++nonEmptyEntryCount_;
 
   if (shouldGrow()) {
-    growAndRehash(capacity() * 2);
+    if (size_ >= (capacity() >> 2)) {
+      growAndRehash(capacity() * 2);
+    } else {
+      // The load factor is dominated by deleted entries, not live ones.
+      // Rehash at the current capacity to reclaim the deleted entries
+      // instead of growing. This keeps the table capacity proportional to
+      // the peak live identifier count; without it, sustained insert/free
+      // churn of unique identifiers doubles the capacity indefinitely,
+      // because deleted entries count towards shouldGrow() and are only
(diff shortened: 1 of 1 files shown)


tjzel:
cc @tmikov

tjzel:
Thanks!

tmikov:
For future reference, check out the changes I made to your commit before landing it in https://github.com/facebook/hermes/commit/47b8dbab74e308e85a31165f51748c2393340ca1:
- Added a test
- FIxed a critical bug with overflow checking

PR #2119, opened 2026-07-26
Guard debugger null deref and spurious StepFinish on async interrupt

### LINKED ISSUE #2118: Debugger SIGSEGV (null codeBlock deref) + spurious pause on DevTools attach
# Debugger SIGSEGV (null `codeBlock` deref) + spurious "paused in debugger" on DevTools attach

## Summary

Connecting React Native DevTools (the CDP / fusebox debugger) to a Hermes app crashes the runtime with a null-pointer `SIGSEGV` inside the debugger's step logic, and — once that is guarded — surfaces a second bug where every attach produces a spurious user-visible "paused in debugger" at an arbitrary location.

Both originate in the same place: the `Ret` branch of `Debugger::runUntilValidPauseLocation` switches to `StepMode::Out` **without ever initializing `preStepState_`**.

## Environment

| | |
|---|---|
| Hermes | v0.16.0 (as vendored by React Native 0.85.3; Maven `hermes-android:250829098.0.10`) |
| React Native | 0.85.3, New Architecture + Bridgeless |
| Platforms | iOS and Android (confirmed on Android arm64-v8a, physical device) |
| Debugger | React Native DevTools (CDP / fusebox) |
| Build | Hermes built from source (required to apply the fix on Android; iOS ships prebuilt hermesvm and cannot work around it) |

## Repro

1. Hermes app with the CDP debugger reachable (RN New Arch + Bridgeless).
2. Have *some* JS executing when the debugger attaches — a small periodic timer/`fetch` loop makes it 100% reproducible. It does **not** require high-volume TurboModule traffic.
3. Attach DevTools (or press "reconnect"). → instant `SIGSEGV` at address `0x0`.

## Symbolicated stack

```
#0 hermes::vm::CodeBlock::getDebugSourceLocationsOffset()        lib/VM/CodeBlock.cpp
#1 hermes::vm::CodeBlock::getSourceLocation(unsigned int)        lib/VM/CodeBlock.cpp
#2 hermes::vm::Debugger::getLocationForState(...)                include/hermes/VM/Debugger/Debugger.h (inlined)
#3 hermes::vm::Debugger::sameStatementDifferentInstruction(...)  include/hermes/VM/Debugger/Debugger.h (inlined)
#4 hermes::vm::Debugger::runUntilValidPauseLocation(...)         lib/VM/Debugger/Debugger.cpp
#5 hermes::vm::Debugger::runDebugger(...)                        lib/VM/Debugger/Debugger.cpp
```

## Root cause

The `Ret` branch in `runUntilValidPauseLocation` sets `curStepMode_ = StepMode::Out` but never initializes `preStepState_`. This branch is reached while servicing the implicit CDP async interrupt that fires on attach, whenever the current IP has no valid pause location. On the **next** pause attempt the loop evaluates `sameStatementDifferentInstruction(state, preStepState_)`, and `getLocationForState(preStepState_)` dereferences `preStepState_.codeBlock == nullptr` → fault `0x0`. `sameStatementDifferentInstruction` additionally dereferences both `OptValue` locations without `hasValue()` checks.

Second symptom (masked by the crash): the same `Ret` branch synthesizes `curStepMode_` while only servicing an implicit interrupt, so `runDebugger` reports `PauseReason::StepFinish` instead of the pending async-trigger reason — surfacing as a spurious "paused in debugger" at wherever JS happened to be.

## Fix

Three defensive guards (PR to follow):

- `getLocationForState`: return `llvh::None` when `state.codeBlock == nullptr`.
- `sameStatementDifferentInstruction`: bail (`return false`) when either location has no value.
- `runDebugger`: prefer a pending `asyncTriggerPauseReason_` over synthesizing `StepFinish`.

A deeper alternative is to initialize `preStepState_ = state` in the `Ret` branch and condition the step synthesis on an actual step in progress; the guards above are the minimal defensive version.

## Credit / prior report

Root-cause analysis and the original fix are **@Elehiggle's**, posted on facebook/react-native#55571 (also tracked as facebook/react-native#56284). That RN issue is closed/stale and the fix lives only in a comment. I independently reproduced and verified the fix on a from-source Android build (breakpoints pause and step correctly, no `SIGSEGV`, no spurious pause) and am filing it here since the code lives in Hermes. Verification discipline note: a synthetic `Runtime.enable`/`Debugger.enable` handshake does **not** reliably reproduce this — only a real breakpoint that resolves a source location does.

Mootzali:
## Summary

Fixes a null-pointer `SIGSEGV` in the Hermes debugger that fires when React Native DevTools (CDP) attaches to a New-Architecture app, plus a second bug that surfaces every attach as a spurious "paused in debugger".

Both stem from the `Ret` branch of `Debugger::runUntilValidPauseLocation` switching to `StepMode::Out` without initializing `preStepState_`. On the next pause, `getLocationForState(preStepState_)` dereferences `preStepState_.codeBlock == nullptr`, and `sameStatementDifferentInstruction` dereferences both `OptValue` locations without `hasValue()` checks. Separately, the branch synthesizes `curStepMode_` while only servicing an implicit interrupt, so `runDebugger` reports `PauseReason::StepFinish` instead of the pending async-trigger reason.

Closes #2118.

## Changes

- `include/hermes/VM/Debugger/Debugger.h`
  - `getLocationForState`: return `llvh::None` when `state.codeBlock == nullptr`.
  - `sameStatementDifferentInstruction`: `return false` when either location has no value.
- `lib/VM/Debugger/Debugger.cpp` (`runDebugger`)
  - Prefer a pending `asyncTriggerPauseReason_` over synthesizing `PauseReason::StepFinish`.

## Credit

Root-cause analysis and the original fix are **@Elehiggle's**, from facebook/react-native#55571 (credited via `Co-authored-by:` on the commit). This PR carries that fix upstream to Hermes, where the code lives; the RN issue is closed/stale with the fix sitting in a comment.

## Testing

Verified manually on a from-source Android build (RN 0.85.3 / Hermes v0.16.0, New Architecture + Bridgeless, physical arm64 device): attached React Native DevTools, set a breakpoint in app JS — it pauses and steps correctly, with no `SIGSEGV` and no spurious "paused in debugger". Without the patch the same sequence crashes reliably.

**No automated test included.** This is a defensive null/`hasValue` guard on a debugger/CDP-attach code path that is awkward to exercise in the existing unit tests. Happy to add a regression test if maintainers can point me at the right harness (e.g. under `unittests/API` or the debugger lit tests).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/include/hermes/VM/Debugger/Debugger.h
+++ b/include/hermes/VM/Debugger/Debugger.h
@@ -620,6 +620,10 @@ class Debugger {
     auto aLoc = getLocationForState(a);
     auto bLoc = getLocationForState(b);
 
+    // Locations may be missing (null codeBlock or no debug info).
+    if (!aLoc.hasValue() || !bLoc.hasValue())
+      return false;
+
     // Same statement in the same codeBlock, but different offsets.
     return a.codeBlock == b.codeBlock && aLoc->statement == bLoc->statement &&
         a.offset != b.offset;
@@ -627,6 +631,8 @@ class Debugger {
 
   OptValue<hbc::DebugSourceLocation> getLocationForState(
       const InterpreterState &state) const {
+    if (state.codeBlock == nullptr)
+      return llvh::None;
     return state.codeBlock->getSourceLocation(state.offset);
   }
 
--- a/lib/VM/Debugger/Debugger.cpp
+++ b/lib/VM/Debugger/Debugger.cpp
@@ -244,8 +244,12 @@ ExecutionStatus Debugger::runDebugger(
         // Continue to run the debugger loop.
 
         // Done stepping.
-        pauseReason =
-            curStepMode_ ? PauseReason::StepFinish : *asyncTriggerPauseReason_;
+        // Prefer a pending async-trigger reason: runUntilValidPauseLocation's
+        // Ret branch can synthesize curStepMode_ while servicing an async
+        // interrupt, which would surface implicit CDP interrupts as
+        // user-visible StepFinish pauses.
+        pauseReason = asyncTriggerPauseReason_ ? *asyncTriggerPauseReason_
+                                               : PauseReason::StepFinish;
         curStepMode_ = llvh::None;
         asyncTriggerPauseReason_ = llvh::None;
       } else {


lorenc-tomasz:
Hi @tmikov,
could you check it, please?

lorenc-tomasz:
Tested it on RN 0.86.2 with Expo 57 and it seems to be working fine :)

tmikov:
We are checking it. We have discovered some problems, we are adding a test, etc.

tmikov:
Check out what actually landed in https://github.com/facebook/hermes/commit/2bfcfa189dd667ba5cb1a28fd4755ddde7a1d2c9 and https://github.com/facebook/hermes/commit/aa4dc53995121184faadceecf2d9d0a535f962c7

joeahn95:
@tmikov Thanks for the update. Not too familiar with the Hermes versioning process. Is there an estimate on when this commit will be tied to a stable React Native version?

tmikov:
@joeahn95 the fix is in the Hermes stable branch. It should be included in the next RN patch release

joeahn95:
@tmikov copy. Expected to release on RN 0.88

PR #2127, opened 2026-08-09
Fix super() and `this` in async arrows inside class constructors

tmikov:
An async arrow becomes three nested IR functions: an outer async
function, an outer generator function, and the inner generator arrow
holding the user code. `genAsyncFunction()` and `genGeneratorFunction()`
pass `capturedState` and `typedClassContext` down to the nested
`FunctionContext`s, but not `legacyClassContext`.

By the time the arrow's parameters and body are emitted,
`hasLegacyClassContext()` is false, so IRGen takes the typed-class path
for `super()` and `this`. This crashed:

```
  class A {}
  new class extends A {
    constructor() {
      (async(a = super(), b) => {})();
    }
  }
```

on the "SemanticResolver must check super() is in a class with a
superclass" assertion, because `superClassNode_` is only set for typed
classes. A `super()` in the arrow body instead of a parameter
initializer crashed the same way.

`this` was worse, because it didn't crash. `genThisExpression()` skipped
the derived-constructor TDZ guard and returned the raw `?CHECKED_this`
slot. The IR verifier rejects that. A release build has no verifier and
leaks the empty sentinel into JS instead of throwing ReferenceError.

Pass `legacyClassContext` down in both places. Non-arrow async and
generator methods are unaffected: only `emitFunctionEpilogue()` and
`initCaptureStateInES5FunctionHelper()` consult the class context in
these outer functions, and both also require a constructor
`constructorKind` or `DefinitionKind`. The outer `ES5Function` wrappers
never have one.

Fixes https://github.com/facebook/hermes/issues/2126
Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

Issue #2126, opened 2026-08-08
Assertion failure in IRGen/ESTreeIRGen-expr.cpp: "SemanticResolver must check super() is in a class with a superclass"


ericliuu:
## Bug Description

Hello, I found this crash in Hermes where valid JavaScript triggers an assertion failure

Hermes git revision (if applicable): https://github.com/facebook/hermes/commit/aa4dc53995121184faadceecf2d9d0a535f962c7, the latest commit on static_h
OS: Linux

## Steps To Reproduce

1. build the debug version of the Hermes CLI
2. run `hermes crash.js`

crash.js
```js
class A {}
new class extends A {
  constructor() {
    (async(a = super(), b) => {
    })();
  }
}
```

output:
```sh
hermes: hermes/lib/IRGen/ESTreeIRGen-expr.cpp:565: hermes::Value* hermes::irgen::ESTreeIRGen::genCallExpr(hermes::ESTree::CallExpressionNode*): Assertion `curFunction()->superClassNode_ && "SemanticResolver must check super() is in a class with a superclass"' failed.
```

backtrace:
```sh
(gdb) bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff704527e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff70288ff in __GI_abort () at ./stdlib/abort.c:79
#5  0x00007ffff702881b in __assert_fail_base (fmt=0x7ffff71d01e8 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n",
    assertion=assertion@entry=0x5555561f49a0 "curFunction()->superClassNode_ && \"SemanticResolver must check super() is in a class with a superclass\"",
    file=file@entry=0x5555561f45b0 "hermes/lib/IRGen/ESTreeIRGen-expr.cpp", line=line@entry=565,
    function=function@entry=0x5555561f4940 "hermes::Value* hermes::irgen::ESTreeIRGen::genCallExpr(hermes::ESTree::CallExpressionNode*)") at ./assert/assert.c:96
[...]

PR #2128, opened 2026-08-09
Report an error for super() in an arrow in a typed class constructor

tmikov:
Typed classes only support `super()` as the first statement of the
derived constructor. `genClassDeclaration()` rejects a constructor that
doesn't start with one, and `genCallExpr()` builds the call out of the
constructor's own `this` parameter, new.target parameter and class type.

An arrow nested in the constructor gets its own `FunctionContext`, and
`genCapturingFunction()` doesn't set `superClassNode_` or
`typedClassContext` on it. A `super()` in that arrow reached
`genCallExpr()` with a null `superClassNode_` and hit an assertion:

```
  class A { x: number; constructor() { this.x = 1; } }
  class B extends A {
    constructor() {
      super();
      (() => { super(); })();
    }
  }
```

Nothing is wrong with that program as far as the compiler is concerned.
The required first statement is there, and no diagnostic was emitted
before the assertion fired. Release builds drop the assertion and call
`genExpression()` on a null node.

Making it work would mean routing the constructor's captured `this` and
new.target into the arrow. That is a feature, and it cuts against the
restriction `genClassDeclaration()` deliberately imposes. Report the
restriction instead. The assertion goes away with it, since it no longer
guards anything reachable.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/lib/IRGen/ESTreeIRGen-expr.cpp
+++ b/lib/IRGen/ESTreeIRGen-expr.cpp
@@ -551,6 +551,19 @@ Value *ESTreeIRGen::genCallExpr(ESTree::CallExpressionNode *call) {
     if (curFunction()->hasLegacyClassContext()) {
       return genLegacyDirectSuper(call);
     }
+    if (!curFunction()->superClassNode_) {
+      // This is a typed class, where super() is only supported as the first
+      // statement of the derived constructor itself; genClassDeclaration
+      // rejects a constructor which doesn't start with one. SemanticResolver
+      // has already verified that we are lexically inside a derived
+      // constructor, so the only way to get here is a super() nested in an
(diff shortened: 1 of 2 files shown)


tmikov:
Depends on #2127 and must land after it.

The new check fires whenever `superClassNode_` is null. That only means "typed class, nested arrow" because #2127 makes every untyped derived constructor and its arrows carry a `legacyClassContext` and return early above it. Without #2127, an untyped `super()` in an async arrow falls through to the same check, so this PR on its own rejects the valid program from #2126 with `error: super() is only supported as the first statement of a derived class constructor` (verified by building this commit alone on top of `static_h`).

PR #2130, opened 2026-08-10
feat: expose "queueMicrotask" on global object

tjzel:
## Summary

I noticed that while Hermes allows to enable microtasks in the engine, it doesn't expose `queueMicrotask` in the global object thus requiring using `HermesInternal` or `jsi::Runtime::drainMicrotasks`.

I don't know if there's any particular reason for it not being exposed on the global object but I figured it could be useful.

## Test Plan

I added a simple test suite.


tmikov:
We don't accept PRs against the `main` branch. Can you please base it against `static_h`.

tjzel:
@tmikov done

Issue #2158, opened 2026-08-27
hermes-parser: babel:true conversion is O(n^2) in sibling count — 88s on a file that parses in 0.1s


elirangoshen:
### Summary

`hermes-parser`'s `parse(src, {babel: true})` takes **88 seconds** on a 502 KB minified JavaScript file. The parse itself takes **0.1 seconds**. All of the time is in the JavaScript code that converts the Hermes AST into a Babel AST.

The cause is that each node replacement copies the whole array of sibling nodes. With one large array in the file, the conversion becomes quadratic.

### Measurements

`hermes-parser@0.36.1`, Node 26.5.0, macOS arm64, on `@lottiefiles/dotlottie-react@0.13.5` (`dist/browser/index.js`, 502 KB):

| call | time |
|---|---|
| `parse(src, {babel: false})` | **0.1s** |
| `parse(src, {babel: true})` | **88.3s** |

CPU profile of the slow call:

| time | share | function |
|---|---|---|
| 36.1s | 41% | `enter` (`transform/SimpleTransform.js:50`) |
| 26.2s | 30% | `getParentKey` (`transform/astNodeMutationHelpers.js:29`) |
| 17.3s | 20% | `replaceInArray` (`transform/astArrayMutationHelpers.js:59`) |
| 4.0s | 5% | garbage collector |

That file has one `ArrayExpression` with **128,834 elements**, out of 153,581 AST nodes in total.

### Minimal repro

```js
import * as hermes from 'hermes-parser';

for (const n of [2000, 4000, 8000, 16000, 32000]) {
  const src = `var a=[${Array.from({length: n}, (_, i) => i).join(',')}];`;
  const t = Date.now();
  hermes.parse(src, {babel: true, sourceType: 'script'});
  console.log(n, Date.now() - t + 'ms');
}
```

```
 2000    46ms
 4000   100ms
 8000   381ms
16000  1350ms
32000  5593ms
```

Time goes up 4x each time the element count doubles, which is quadratic. Extending the curve to 128,834 elements gives roughly 88s, matching the real file.

With `babel: false` the same inputs are fast, so this is only the conversion step.

Long lines and non-ASCII characters are not involved — we checked both. A file with 40,000 statements on a single line converts in 321ms, and the slow file is pure ASCII.

### Cause

`replaceInArray` (`transform/astArrayMutationHelpers.js:59`) rebuilds the entire array for every replacement:

```js
function replaceInArray(array, index, elements) {
[...]

PR #2160, opened 2026-08-28
fix(hermes-parser): make single-node AST replacement O(1) instead of O(siblings)

HuzaifaAbdulRehman:
## Summary

`parse(src, {babel: true})` is quadratic in sibling count. A 502 KB prebuilt file takes 88 seconds where the parse itself takes 0.1 (#2158), which put a 94 second floor on every cold build for a downstream bundler (callstack/repack#1447).

Each single-node replacement runs two independent O(siblings) operations. `getParentKey` (`astNodeMutationHelpers.js:24`) rescans the parent array to rebuild an index the traversal already had, then `replaceInArray` (`astArrayMutationHelpers.js:61`) rebuilds the whole array. `_traverse` holds that index in `j`, but `TraverserCallback` is typed `(node, parent) => void`, so it never reaches the callback.

This threads the position down as two optional trailing parameters and uses it in `replaceNodeOnParent` after an O(1) identity probe, falling back to the existing scan when the hint does not match. Array-valued replacements change sibling count and still go through `replaceInArray`. At 32,000 elements: 21,187ms to 225ms.

Both halves are needed. The issue suggests either change alone removes most of the cost. Measured separately, the index alone leaves 9,664ms and the in-place write alone leaves 12,649ms, against 21,187ms baseline and 225ms with both.

`TraverserCallback` and `replaceNodeOnParent` are exported from `index.js`, so this touches public API. No existing caller changes behaviour, because the fast path only engages when the new parameters are passed. `replaceInArray` is deliberately untouched, since `hermes-transform` calls it directly and relies on its copy-on-write result.

## Test Plan

From `tools/hermes-parser/js`: `yarn test` gives 2,615 passing and 456 snapshots, against 2,610 on a pristine checkout. The difference is the five tests added here. `yarn flow` and `yarn lint` are clean.

The new tests cover replacement at the first, middle and last array position, inside a nested array, and one node replaced by many so the length-changing path stays covered. A guard asserts the sibling array keeps its object identity, and both mechanism tests fail without the fix.

### DIFF
--- a/tools/hermes-parser/js/hermes-parser/__tests__/transform/SimpleTransform-test.js
+++ b/tools/hermes-parser/js/hermes-parser/__tests__/transform/SimpleTransform-test.js
@@ -119,5 +119,91 @@ describe('SimpleTransform', () => {
         },
       });
     });
+    it('Array element at each position', () => {
+      // A replacement must land correctly at the start, middle and end of the
+      // sibling array.
+      for (const target of [1, 2, 3]) {
+        expectTransformToEqual({
+          code: `[1, 2, 3];`,
+          result: `[${[1, 2, 3].map(v => (v === target ? 9 : v)).join(', ')}];`,
+          transform(node) {
+            if (node.type === 'Literal' && node.value === target) {
+              // $FlowFixMe[incompatible-type]
+              return {
+                type: 'Literal',
+                value: 9,
+                raw: '9',
+                literalType: 'numeric',
+              };
+            }
+            return node;
+          },
+        });
+      }
+    });
+    it('Nested array element', () => {
+      expectTransformToEqual({
+        code: `[[1, 2], [3]];`,
+        result: `[[1, 9], [3]];`,
+        transform(node) {
+          if (node.type === 'Literal' && node.value === 2) {
+            // $FlowFixMe[incompatible-type]
+            return {
+              type: 'Literal',
+              value: 9,
+              raw: '9',
+              literalType: 'numeric',
+            };
+          }
+          return node;
+        },
+      });
+    });
+    it('One node with many nodes inside an array', () => {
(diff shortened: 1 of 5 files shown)


HuzaifaAbdulRehman:
@SamChou19815 apologies for the direct ping. You've been the most active in `tools/hermes-parser/js` lately so you seemed like the right person to ask, though happy to be pointed elsewhere.

Short version. `parse(src, {babel: true})` is quadratic in sibling count, which put a 94 second floor on every cold build for anyone whose bundler routes through hermes-parser. A 32,000 element array goes from 21,187ms to 225ms. Output is unchanged, verified by the existing 456 snapshots plus 15 syntax samples hashed before and after. Flow, lint and the full suite are green, with 2,615 passing against 2,610 on a clean checkout, the difference being the five tests added here.

One thing the issue gets wrong is worth flagging. It proposes two changes and says either alone should remove most of the cost. Measured separately they leave 46% and 60% behind, because each removes only one of the two O(siblings) operations per replacement and the other still runs. Both are needed.

The piece I'd most like a second opinion on is the `hermes-transform` interaction. I left `replaceInArray` alone deliberately, since `ReplaceNode.js` and `ReplaceStatementWithMany.js` call it directly and the referential-equality checks in `nodeWith` and `arrayIsEqual` rely on its copy-on-write result. Keeping the in-place write inside `replaceNodeOnParent` is what leaves that contract alone, but you know that code far better than I do, so I'd rather have it checked than assume.

Happy to split this into two commits if that reviews more easily.

HuzaifaAbdulRehman:
@SamChou19815 no rush. Import Status has been pending since the 28th with both Meta checks green.

Happy to split the traverser plumbing out from the `replaceNodeOnParent` change if two smaller diffs would review more easily.

PR #2161, opened 2026-08-28
Add crypto.getRandomValues as a JSI extension

### LINKED ISSUE #1003: Expose important CSPRNG: crypto.getRandomValues
RN does not support `globalThis.crypto`. It's understandable, the module is huge.

There is no need in re-implementing the whole `crypto` module. Just one function: `crypto.getRandomValues`. The function exposes CSPRNG (cryptographically secure pseudorandom number generator). The function is critical to MANY apps which are not even related to deep tech.

It's important for one reason: without exposing the function, many people will go path of least resistance and simply switch to `Math.random()` which is PRNG, but not cryptographically secure. This has happened in the past and apps have been broken with catastrophic results: https://www.deepsource.com/blog/dont-use-math-random, https://owasp.org/www-community/vulnerabilities/Insecure_Randomness, https://blog.ledger.com/Funds-of-every-wallet-created-with-the-Trust-Wallet-browser-extension-could-have-been-stolen/

I write popular cryptographic libraries. Everything else is easily polyfillable. ed25519, sha3, pbkdf2, whatever. The CSPRNG is not easily polyfillable: it exposes low-level system interface. You cannot emulate or simulate this.

alexgleason:
## Summary

Closes #1003.

Implements the Web Crypto API's `crypto.getRandomValues()` as a JSI extension, following the existing extension pattern (TextEncoder):

- `12-Crypto.js` validates the argument via intrinsics captured before any user code runs (integer TypedArray brand check through the genuine `%TypedArray%.prototype` accessors, 65536-byte quota) and installs `globalThis.crypto`.
- `Crypto.cpp` fills the view's bytes directly from the OS CSPRNG:
  - `CCRandomGenerateBytes` on Apple
  - `BCryptGenRandom` (`BCRYPT_USE_SYSTEM_PREFERRED_RNG`) on Windows
  - the `getrandom(2)` syscall with `flags=0` on Linux/Android (works on older bionic/glibc without the libc wrapper, and blocks until the kernel entropy pool is initialized)
  - `getentropy()` elsewhere (Emscripten, the BSDs), chunked at 256 bytes

**There is deliberately no fallback to a weaker source.** If entropy cannot be obtained, `getRandomValues` throws; a platform with no known OS CSPRNG interface fails to compile rather than silently degrading. `std::random_device` was rejected because the C++ standard permits it to be a deterministic PRNG (older MinGW-w64 shipped exactly that), and libstdc++ on x86-64 services it from RDSEED/RDRAND alone, bypassing the kernel pool. This matters because `getRandomValues` is the primitive wallet and cryptography libraries build on — the motivating use case in #1003.

The native fill re-validates `byteOffset`/`byteLength` against the actual ArrayBuffer and roots the buffer via `TypedArrayBufferInfo`, so writes cannot escape the genuine buffer even if user code tampers with properties or prototypes.

### Spec deviations

Hermes has no `DOMException`, so the closest ECMAScript equivalents are thrown:

- non-integer TypedArray argument: `TypeError` (spec: `TypeMismatchError`)
- view larger than 65536 bytes: `RangeError` (spec: `QuotaExceededError`)
- view on a detached ArrayBuffer: `TypeError`

## Test plan

- New lit test `test/hermes/crypto-get-random-values.js` covering: all nine integer TypedArray kinds, zero-length views, the 65536-byte quota boundary (bytes, not elements), subarray windows only being written, rejection of non-integer-TypedArray arguments (floats, DataView, ArrayBuffer, plain objects), detached buffers, and continued operation after user code replaces `TypeError`/`RangeError`/`Uint8Array` globals.
- `check-hermes` passes on Linux x86-64 (ASan+Debug). Verified under `strace` that a 64 KiB fill is serviced by a single `getrandom(..., 65536, 0)` syscall.
- `clang-format` clean.


paulmillr:
Looks good from my end

Although agent mentioned this:

- **[P2] Android fails on supported older kernels.** The [direct `getrandom` syscall](https://github.com/facebook/hermes/blob/1965f761715e60a7cf78259bb834abac39b9bf66/API/hermes/extensions/Crypto.cpp#L67-L82) always throws `ENOSYS` on pre-3.17 kernels. Hermes’ minSdk is 24, and AOSP documents supported Android devices running 3.10 without this syscall. Bionic recommends `arc4random_buf()`, available at every API level. Add an Android-specific secure compatibility path. ([AOSP evidence](https://android.googlesource.com/platform/art/+/d6e2c388d0d91c0e7cce66147cc188caed3208b6), [Bionic API](https://android.googlesource.com/platform/bionic/+/master/libc/include/sys/random.h#55))

alexgleason:
@paulmillr good catch!

I added an android conditional that uses `arc4random_buf` instead of `getrandom`

PR #2163, opened 2026-08-31
proposal: copy native state for structured clone

tjzel:
## Summary

I was experimenting with `jsi::Serialized` and noticed that the structured clone algorithm underneath didn't copy a `jsi::NativeState` tied to cloned object. `jsi::NativeState` is extensively used for React Native libraries and `jsi::Serialized` not supporting it is, in my opinion, a serious hindrance to the `jsi::Serialized` adoption.

The code in this PR is an AI generated PoC to see how extensive would the changes be.

Does support for `jsi::NativeState` in `jsi::Serialized` and structured clone align with your idea of how it should be implemented?

## Test Plan

New unit tests.


tjzel:
cc @tmikov

PR #2172, opened 2026-09-07
Add `Promise.allKeyed` and `Promise.allSettledKeyed` methods

retyui:
## Summary

Implement new methods: `Promise.allKeyed`, `Promise.allSettledKeyed`

TC39 proposal (Stage 3): https://github.com/tc39/proposal-await-dictionary
TS: https://github.com/microsoft/TypeScript/pull/64093

`Promise.allSettledKeyed()`:
 - mdn: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettledKeyed
 - spec: https://tc39.es/proposal-await-dictionary/#sec-promise.allsettledkeyed

`Promise.allKeyed()`:
 - mdn: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allKeyed
 - spec: https://tc39.es/proposal-await-dictionary/#sec-promise.allkeyed

## Test Plan

```bash
# Build
rm -rf ./build
cmake -S hermes -B build -GNinja -DHERMES_ENABLE_INTL=ON -DCMAKE_BUILD_TYPE=Debug -DHERMESVM_ALLOW_JIT=2 
cmake --build ./build

# Run tests
LIT_OPTS="-j1" LIT_FILTER="promise-all-keyed.js" cmake --build build --target check-hermes 
LIT_OPTS="-j1" LIT_FILTER="promise-all-settled-keyed.js" cmake --build build --target check-hermes 

LIT_OPTS="-j1" LIT_FILTER="promise-all-keyed-resolution-order.js" cmake --build build --target check-hermes 
LIT_OPTS="-j1" LIT_FILTER="promise-all-settled-keyed-resolution-order.js" cmake --build build --target check-hermes 
```


retyui:
@lavenzg could you please review this PR ?

lavenzg:
> @lavenzg could you please review this PR ?

Thanks for contributing this PR! I will review it.

lib/InternalJavaScript/01-Promise.js
```diff
@@ -133,6 +133,22 @@
       return capability.promise;
     }
 
+    static allKeyed(promisesObject) {
+      if(typeof promisesObject !== 'object' || promisesObject === null) {
```
lavenzg, line 137: It should accept `function`.

lib/InternalJavaScript/01-Promise.js
```diff
@@ -133,6 +133,22 @@
...
+        var C = this;
+        var capability = newPromiseCapability(C);
+        capability.reject(new TypeError('Argument of Promise.allKeyed is not a non-null object'));
```
lavenzg, line 140: > b. Perform ? [Call](https://tc39.es/ecma262/#sec-call)(promiseCapability.[[Reject]], undefined, « error »).

The rejection callback should use `undefined` as receiver.

lib/InternalJavaScript/01-Promise.js
```diff
@@ -133,6 +133,22 @@
...
+        return capability.promise;
+      }
+      var keys = Reflect.ownKeys(promisesObject);
+      return this.all(
+        keys.map(key => promisesObject[key])
```
lavenzg, line 145: This eager `map` is wrong. The spec evaluates and resolves properties one by one. Also we need to handle exception that could be raised in the input.

lib/InternalJavaScript/01-Promise.js
```diff
@@ -133,6 +133,22 @@
...
     }
 
+    static allKeyed(promisesObject) {
+      if(typeof promisesObject !== 'object' || promisesObject === null) {
+        var C = this;
+        var capability = newPromiseCapability(C);
+        capability.reject(new TypeError('Argument of Promise.allKeyed is not a non-null object'));
+        return capability.promise;
+      }
+      var keys = Reflect.ownKeys(promisesObject);
```
lavenzg, line 143: > b. If propertyDesc is not undefined and propertyDesc.[[Enumerable]] is true, then

We need to do this check.

lib/InternalJavaScript/01-Promise.js
```diff
@@ -133,6 +133,22 @@
...
+      return this.all(
+        keys.map(key => promisesObject[key])
+      ).then(results => keys.reduce((acc, key, index) => {
+        acc[key] = results[index];
+        return acc;
+      }, {}));
```
lavenzg, line 149: > 1. Let obj be [OrdinaryObjectCreate](https://tc39.es/ecma262/#sec-ordinaryobjectcreate)(null).

It should be `Object.create(null)` instead of `{}`.

lib/InternalJavaScript/01-Promise.js
```diff
@@ -133,6 +133,22 @@
...
 
+    static allKeyed(promisesObject) {
+      if(typeof promisesObject !== 'object' || promisesObject === null) {
+        var C = this;
+        var capability = newPromiseCapability(C);
+        capability.reject(new TypeError('Argument of Promise.allKeyed is not a non-null object'));
+        return capability.promise;
+      }
+      var keys = Reflect.ownKeys(promisesObject);
+      return this.all(
```
lavenzg, line 144: Better not depending on `all`. The spec does not rely on that. Also the current way returns the promise created in `then`, but the spec only creates once.

```
class P extends Promise {
    static all() {
      throw new Error("all was called");
    }
  }

  P.allKeyed({});
```

lavenzg (changes requested):
My comments apply to `allSettledKeyed` as well. Please also add tests to cover these cases. Make sure it's compliant to the spec (if possible).

lib/InternalJavaScript/01-Promise.js
retyui: @lavenzg so, I need to call `reject` with `undefined` , right ?
```
capability.reject.call(undefined, new TypeError('...'));
```

lib/InternalJavaScript/01-Promise.js
retyui: @lavenzg Do you know if it possible to re-use this helper by importing it (`import {isObject} from './myfile'`) 
to avoid creating duplicates 

https://github.com/facebook/hermes/blob/9a41742f5ffbd5e7f9e7bc36b51b5e9c4e14111e/lib/InternalJavaScript/05-IteratorHelperDependencies.js#L616-L619

lavenzg: No these JS files are ordered and here `isObject` is a function expression. Just do the full check as other places do in the same file. We can refactor all those to a single helper later.

retyui:
@lavenzg  could you please make another round of review ?

lib/InternalJavaScript/01-Promise.js
retyui: fixed

lib/InternalJavaScript/01-Promise.js
retyui: fixed

lib/InternalJavaScript/01-Promise.js
retyui: fixed

lib/InternalJavaScript/01-Promise.js
retyui: fixed

lib/InternalJavaScript/01-Promise.js
retyui: fixed

retyui:
@lavenzg could you please make another round of review ?

PR #2177, opened 2026-09-08
Use OIDC trusted publishing on 260318099.0.0-stable

cipolleschi:
## Summary

Ports #2171 to `260318099.0.0-stable`.

- Inline the publish job into `rn-build-hermes.yml`, because npm validates the top-level workflow filename in the OIDC `workflow_ref` claim.
- Grant the publish job `id-token: write` and use the `npm-publish` environment.
- Remove the long-lived `GHA_NPM_TOKEN` credential path and the reusable `publish.yml` workflow.
- Upgrade to npm 11 before publishing artifacts; trusted publishing requires npm >= 11.5.1.
- Add an opt-in `update-latest-v1` workflow input. Releases use `latest` by default and `latest-v1` when selected.
- Always pass an explicit npm dist-tag, including for dry runs.
- Upload npm logs even when publishing fails.

## Why

Trusted publishing is enabled for `hermes-compiler` on npmjs.com, but this release branch still uses `GHA_NPM_TOKEN`. The equivalent fix landed on `250829098.0.0-stable` in #2171 and on `static_h` in 86ec984e7.

Before a release, confirm that the npm trusted-publisher entry names workflow `rn-build-hermes.yml` and environment `npm-publish`, or leaves the environment blank.

## Test plan

- Parsed `.github/workflows/rn-build-hermes.yml` successfully as YAML.
- `node --check utils/scripts/hermes/publish-npm.js`
- `git diff --check`

A real OIDC exchange cannot be validated with `npm publish --dry-run`; the first release run provides the end-to-end validation.


cipolleschi:
landed as 4188b63f24026ca62b79dc81d95db122a59c6089

PR #2178, opened 2026-09-08
Bump hermes-compiler version to 260318099.0.3

cipolleschi:
## Summary

We just released 260318099.0.2. The current value in the hermes-compiler/package.json should always point to the next version we need to publish.

## Test Plan

N/A


fabriziocucci (approved):

cortinico (approved):

cipolleschi:
landed as 21cc9fb2a4021d43e3a27c183211ed8f05e41790

PR #2181, opened 2026-09-13
Fix Date for time zones whose standard offset changed over history

### LINKED ISSUE #2159: Date: a DST transition at local midnight makes setHours read back the wrong day
## Summary

When a timezone's DST transition happens at local midnight, `Date` converts a wall-clock time to an instant using the UTC offset in force **before** the jump, then reads that instant back using the offset **after** it. A time in the last hour of the preceding day therefore reads back as the next day.

## Repro

```js
process.env.TZ = 'Europe/Kiev' // or run the device in that timezone

const d = new Date(1984, 2, 31)   // 31 March 1984, local
d.setHours(23, 59, 59, 999)
console.log(d.getDate())
```

| | result |
|---|---|
| Hermes | `1` |
| V8 / JSC / Node | `31` |

Kyiv moved its clocks at `00:00` from 1981 through 1984, so `1984-04-01T00:00` local does not exist. Hermes resolves `1984-03-31T23:59:59.999` to an instant that it then formats as 1 April.

Instrumented on device:

```
start     1984-03-30T22:00:00.000Z   getDate=31
setHours  1984-03-31T21:59:59.999Z   getDate=1
```

The forward conversion applied no DST term; reading the same instant back applied one.

## Why this matters beyond a corner case

`dayjs` implements `daysInMonth()` as `endOf('month').date()` - take the last day of the month, `setHours(23, 59, 59, 999)`, read the day. Under this bug that returns `1` instead of the real length, and every consumer that builds a month grid from it renders a single cell.

We hit it through [`react-native-ui-datepicker`](https://github.com/farhoudshapouran/react-native-ui-datepicker), whose `getDaysInMonth` feeds `dayjs().daysInMonth()` straight into `Array.from({ length })`. A user picking a birth date in March 1984 saw a calendar with one day in it.

Scanning every month from 1930 to 2030 on the device under `Europe/Kiev` gives exactly four wrong answers: **March 1981, 1982, 1983, 1984** - the years the transition was at midnight. From 1985 Kyiv moved it to 02:00 and the results are correct again.

This is not limited to historical dates. Any zone that still transitions at local midnight is exposed for its **current** month: Chile, Cuba, Iran, Paraguay and parts of the Levant among them.

## Possibly related

On the same device, Hermes reports Kyiv as UTC+2/+3 for 1984, while Android's own tz database (`TZ=Europe/Kiev date` in the shell) reports MSK +3 / MSD +4, matching macOS and Node. Hermes appears to carry the zone's current standard offset backward through history. That does not cause the bug above, but it may share a root.

## Environment

- Hermes `v0.17.0`, as shipped with React Native `0.86.2`
- Android 16, Samsung SM-S916B (arm64-v8a)
- Device timezone `Europe/Kiev`

## Workaround

Compute month length with `Date.UTC`, which has no offset to resolve:

```js
const daysInMonth = (year, month) => new Date(Date.UTC(year, month + 1, 0)).getUTCDate()
```

Sanjays2402:
Fixes #2159.

**The bug**

With `TZ=Europe/Kyiv`, this prints `1` (April 1st) instead of `31`:

```js
const d = new Date(1984, 2, 31);
d.setHours(23, 59, 59, 999);
console.log(d.getDate());
```

**What's going on**

`LocalTimeOffsetCache` cached a single standard offset (`ltza_`, taken at startup) and combined it with per-interval DST offsets. That works as long as the *standard* offset never changes — but it does change over history. Kyiv was UTC+3 (standard time) until 1991 and is UTC+2 now, so 1984 dates got the modern UTC+2 standard offset mixed with the 1984 DST value and landed an hour off, pushing the date into the next day.

**The fix**

Cache the *total* local offset (`tm_gmtoff`, i.e. standard offset + DST) per interval instead of just the DST offset. `ltza_` is kept for the initial local→UTC guess and to detect a genuine `TZ` environment change — a historical standard-offset difference no longer resets the cache (only a change in the *current* standard offset does). The Windows path is untouched behaviorally (no `tm_gmtoff` there; Windows applies current rules to all dates anyway). The pre-transition probing that keeps the ECMA-262 ambiguous/skipped-hour rule intact is unchanged.

**Tests**

- Renamed `DaylightSavingTATest` → `LocalOffsetTest` with total-offset expectations.
- Added `HistoricalStandardOffsetChangeTest`: asserts the cached offset for 1984-03-31 in Europe/Kyiv is UTC+3, and runs the issue's repro end-to-end (`getDate() === 31`). I verified the new test **fails on unfixed main** (`getDate()` comes back `1`) and passes with the fix.
- All 15 `DateUtilTest` unit tests pass; the date lit tests (`date-constructor`, `date-default-timezone`, `date-local-to-utc`, `regress-date-cache`) run clean.
- I also validated the algorithm with a standalone harness: the issue repro plus ~25k date round-trips and 2M random offset probes across six time zones (incl. Kyiv, Asuncion, Lord Howe, Kwajalein) — zero mismatches against the system tzdata, and agreement with V8 on edge cases like Kwajalein's skipped day.

Validation was done on Linux with the system IANA tzdata; I haven't tested on Windows/macOS.


lavenzg:
This will not work. We need to have a single standard offset when computing the local time offset (we need to guess a UTC time from the local time so that we can get its DST offset). And it's documented that this conversion will not work for historical timezone changes.

To cache the correct total offset, we would need the timezone database (and ICU) to compute it.

I think this PR will fail other cases:
```
// TZ=Europe/Kyiv
var d = new Date(1984, 3, 1, 0, 30);
print(d.toISOString());
```

Sanjays2402:
You're right — reproduced it. With the PR as it was, `TZ=Europe/Kyiv; new Date(1984, 3, 1, 0, 30)` gave `1984-03-31T20:30:00.000Z`: the guessed UTC lands past the 1984-03-31 transition, so the probed total offset (+4) comes from the wrong side.

Pushed a fix: re-probe at the UTC time implied by the total offset at the guess. If the offsets disagree, a transition lies between the two, so the guess is unreliable and we fall back to the single current standard offset plus the DST offset at the guess (the documented behavior for ambiguous times). Everything else still uses the cached total offset, so the #2159 fix is untouched.

Added `HistoricalDstGapLocalToUtcTest` with your exact case — now `1984-03-31T21:30:00.000Z`, matching V8. All 16 `DateUtilTest` tests pass, the date lit tests are clean, and a ~9k-case differential run against V8 across six time zones shows no mismatches.

One honest caveat: local-to-UTC for dates whose standard offset changed remains best-effort — exact wall-time resolution would need the tz database — but the realistic cases now agree with V8.

Sanjays2402:
Going to close this for now — thanks for the thorough review. Happy to revisit if the approach looks worth pursuing.

PR #2183, opened 2026-09-16
Resolve the calendar in Intl.DateTimeFormat resolvedOptions() on Apple platforms

### LINKED ISSUE #1716: `Intl.DateTimeFormat().resolvedOptions().calendar` is undefined
## Bug Description

`Intl.DateTimeFormat().resolvedOptions().calendar` is not defined on iOS.

I suspect this is a new change as of about a week ago and is possibly related to an iOS update?

My Expo app (SDK 52.0.44) suddenly started crashing on start up about a week ago because this property wasn't defined. In my case, the crash was related to the [Temporal polyfill](https://github.com/js-temporal/temporal-polyfill/) which is expecting the `calendar` property to exist. I had recently updated from version 0.44 to 0.51 of the polyfill, **however** there was no issue for about a month after the update until about a week ago when the polyfill started throwing an error due to `calendar` not being defined. The code in the polyfill also doesn't suggest that it's using this value any differently between the versions.

React Native version: 0.76.9
OS: iOS
Platform (most likely one of arm64-v8a, armeabi-v7a, x86, x86_64): tested on an iPhone 16 Pro Max, whatever that uses

## Steps To Reproduce

Here is an Expo Snack that reproduces the issue:

https://snack.expo.dev/@miramar-zeff/ios-missing-intl-datetimeformat-calendar-bug?platform=ios

Launch the iOS instance and note how the "Calendar:" is blank. It shows "Calendar: gregory" (with my locale, at least) in Android and Web.

## The Expected Behavior

Based on [the MDN article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions#calendar), I believe this value should always be defined, though I guess I'm not certain. It appears in web and Android builds, at least.

kassdinzheng123:
## Summary

`Intl.DateTimeFormat().resolvedOptions().calendar` is missing on Apple platforms when no calendar is requested explicitly (#1716). Resolve it from the formatter's actual default calendar (`gregory` for `en-US`, `buddhist` for `th-TH`).

For a requested calendar, set the `ca` extension on the locale passed to `NSDateFormatter`, rather than constructing and assigning an `NSCalendar`. This lets Foundation use the same calendar for formatting and pattern selection. Following review, the full bidirectional mapping has been removed: `bcp47CalendarIdentifierFor` only translates `NSCalendarIdentifierGregorian` to `gregory` and `NSCalendarIdentifierEthiopicAmeteAlem` to `ethioaa`, passing other identifiers through.

## Root cause

The Apple backend has no locale-data table, and `resolveLocale` only supplies a calendar when it appears in an option or a Unicode extension. Without an explicit request, `calendar_` remained empty, so `resolvedOptions()` omitted it.

The original code also passed UTS-35 identifiers directly to `initWithCalendarIdentifier:`, although some Foundation names differ. The first PR revision mapped those identifiers, but directly assigning a calendar still left pattern selection tied to the original locale. On the local macOS 27.0 build, `{calendar: 'ethioaa', year: 'numeric', timeZone: 'UTC'}` formatted 2020-01-02 as `7512`, omitting the era. Setting `ca` in the formatter's locale produces `7512 AA`.

## Preserved behavior and scope

- The calendar option continues to override a locale's `ca` extension. Other locale extensions are retained.
- Only the formatter's private locale is updated; `resolvedOptions().locale` keeps the existing ResolveLocale result and does not gain extensions solely from options.
- Calendar option validation and the backend's existing unsupported-calendar handling in `resolveLocale` are unchanged. This does not add supported-calendar negotiation or fill in `numberingSystem`.
- The change is limited to the Apple backend and its lit regression test; ICU and Android implementations are untouched.

## Test Plan

`test/hermes/intl/date-time-format-calendar.js` runs through both Hermes bytecode and Static Hermes. It covers locale-default calendars, Japanese and ISO-8601 requests, actual formatted dates, the review's `en-US-u-ca-ethioaa` reproduction, explicit `ethioaa`, option-over-extension precedence, preservation of the resolved locale, and the `dateStyle` path.

Before this revision (`260fe9af1`), the expanded regression fails in both execution paths: expected `en-US ethioaa 7512 AA`, received `en-US ethioaa 7512`. The reviewer's Gregorian fallback was not reproduced on this macOS 27.0 machine; the exact locale-extension reproduction is retained in the regression test.

Validation on macOS 27.0, Debug, `HERMES_ENABLE_INTL=ON`:

```sh
LIT_FILTER='intl' cmake --build cmake-build-intl --target check-hermes -j 8
# Expected Passes: 21; Unsupported Tests: 1

cmake --build cmake-build-intl --target check-hermes -j 8
# Expected Passes: 4138; Expected Failures: 6; Unsupported Tests: 175
# Node.js NAPI tests: 41 pass; 0 fail; 23 skipped
```

Both commands exit successfully, with no unexpected failures. The expanded calendar regression passes in both Hermes and Static Hermes. All four newly added output checks also agree with Node 26.4.0. `clang-format` reports no changes, and `git diff --check` passes.

Fixes #1716


lib/Platform/Intl/PlatformIntlApple.mm
```diff
@@ -899,6 +899,45 @@ uint8_t getCurrencyDigits(std::u16string_view code) {
...
+/// Mapping between the calendar identifiers of
+/// https://unicode.org/reports/tr35/#UnicodeCalendarIdentifier and the
+/// corresponding NSCalendar identifiers. Identifiers with no counterpart are
+/// omitted; the fallback path returns "gregory".
+static const std::pair<std::u16string_view, NSString *const>
+    kCalendarIdentifiers[] = {
+        {u"buddhist", NSCalendarIdentifierBuddhist},
+        {u"chinese", NSCalendarIdentifierChinese},
+        {u"coptic", NSCalendarIdentifierCoptic},
+        {u"ethioaa", NSCalendarIdentifierEthiopicAmeteAlem},
```
lavenzg: On my mac `NSCalendarIdentifierEthiopicAmeteAlem` produces Gregorian (maybe something wrong in the underlying Apple SDK). Need to fix it.

```
var f = new Intl.DateTimeFormat("en-US-u-ca-ethioaa", {
  timeZone: "UTC",
  year: "numeric"
});
print(f.resolvedOptions().calendar, f.format(Date.UTC(2020, 0, 2)));
```

lavenzg (changes requested):
I wonder if the mapping here is really necessary. Ideally we should just change `bcp47CalendarIdentifierFor` to specially handle NSCalendarIdentifierGregorian and NSCalendarIdentifierEthiopicAmeteAlem, and update the locale to use correct `ca` extension when calendar is given.

lib/Platform/Intl/PlatformIntlApple.mm
kassdinzheng123: Thanks, addressed in b6fe16f9b. I removed the bidirectional mapping: `bcp47CalendarIdentifierFor` now only special-cases Gregorian and Ethiopic Amete Alem. Requested calendars are set through the formatter locale's `ca` extension, with no explicit `NSCalendar` assignment. The public resolved locale and its other extensions are preserved.

Your exact reproduction now has a regression test and prints `ethioaa 7512 AA` here. I could not reproduce the Gregorian fallback on macOS 27.0, but did reproduce a related failure with an explicit `{calendar: 'ethioaa', year: 'numeric'}` option: the previous revision printed `7512` without the era. Using the calendar in the formatter locale fixes that to `7512 AA`. The expanded test fails before this revision and passes after it in both Hermes and Static Hermes; it also covers option precedence and the `dateStyle` path.

The Intl suite passes (21 expected passes, 1 unsupported), as does full `check-hermes` (4138 expected passes, 6 expected failures, 175 unsupported; no unexpected failures). The four new output checks agree with Node 26.4.0. Could you also check the original reproduction on your SDK with this change?

lavenzg: Yeah it's fixed.

PR #2184, opened 2026-09-16
Keep the requested unit in Intl.NumberFormat unit style on Apple platforms

### LINKED ISSUE #2011: Intl.NumberFormat unit style formats day/hour/minute as seconds on Hermes
## Bug Description

`Intl.NumberFormat` with `style: 'unit'` returns the wrong unit for duration-style units in Hermes. In particular, `hour` and `minute` are formatted as seconds, and `day` is formatted with the singular unit in `en-US` even for plural values.

This caused a React Native app countdown that should render as `23 days 22 hr` to render as `23 day 79,200s` on Hermes.

I searched existing `Intl.NumberFormat` issues and found related NumberFormat correctness issues such as #1035 and #1318, but I did not find an exact duplicate for `style: 'unit'` with `day` / `hour` / `minute`.

- [ ] I have run `gradle clean` and confirmed this bug does not occur with JSC. N/A for this report: the repro below is iOS/macOS Hermes CLI. I did compare against JavaScriptCore on macOS and V8/Node; both return the expected unit strings.
- [x] The issue is reproducible with the latest version of React Native. I reproduced it with the Hermes `0.16.0` iOS prebuilt tarball referenced by `react-native@0.85.3` when Hermes V1 is disabled. The same issue also reproduces in my app on React Native `0.83.4` / Hermes `0.14.1`.

Hermes git revision (if applicable): `hermes-v0.16.0` release tarball; also observed with `hermes-v0.14.1`
React Native version: `0.85.3` for the latest-version check; originally observed in `0.83.4`
OS: macOS 26.4.1 (`arm64`), iOS React Native app runtime
Platform (most likely one of arm64-v8a, armeabi-v7a, x86, x86_64): `arm64` / iOS + macOS Hermes CLI from the iOS prebuilt tarball

## Steps To Reproduce

1. Install or download Hermes `0.16.0` from the React Native `0.85.3` Hermes iOS prebuilt artifact.
2. Run the code below with the Hermes CLI from that artifact.
3. Compare the output with JavaScriptCore or Node/V8.

Commands used for the latest React Native / Hermes check:

```sh
mkdir -p /tmp/hermes-rn-latest-check
cd /tmp/hermes-rn-latest-check
bun init -y
bun add react-native@0.85.3

curl -L \
  https://repo1.maven.org/maven2/com/facebook/hermes/hermes-ios/0.16.0/hermes-ios-0.16.0-hermes-ios-debug.tar.gz \
  -o hermes-ios-debug.tar.gz
mkdir hermes-016
tar -xzf hermes-ios-debug.tar.gz -C hermes-016

./hermes-016/destroot/bin/hermes repro.js
```

Code example (`repro.js`):

```js
const log = typeof print === 'function' ? print : console.log;

function show(unit, value) {
  const formatted = new Intl.NumberFormat('en-US', {
    style: 'unit',
    unit,
    unitDisplay: 'short',
    maximumFractionDigits: 1,
  }).format(value);

  log(`${unit}: ${formatted}`);
}

show('day', 2);
show('hour', 2);
show('minute', 2);
show('second', 2);
```

Hermes `0.16.0` output:

```text
day: 2 day
hour: 7,200s
minute: 120s
second: 2s
```

Hermes version output:

```text
Hermes JavaScript compiler and Virtual Machine.
  Hermes release version: 0.16.0
  HBC bytecode version: 96
```

JavaScriptCore on the same machine, using `osascript -l JavaScript`, returns:

```text
day: 2 days
hour: 2 hr
minute: 2 min
second: 2 sec
```

Node/V8 on the same machine returns:

```text
day: 2 days
hour: 2 hr
minute: 2 min
second: 2 sec
```

The same Hermes behavior also reproduces with my app's installed Hermes `0.14.1` from React Native `0.83.4`:

```text
Hermes release version: 0.14.1
HBC bytecode version: 96

day: 2 day
hour: 7,200s
minute: 120s
second: 2s
```

## The Expected Behavior

`Intl.NumberFormat` should preserve the requested unit when `style: 'unit'` is used.

Expected `en-US` short-unit output for the example above:

```text
day: 2 days
hour: 2 hr
minute: 2 min
second: 2 sec
```

At minimum, `unit: 'hour'` should not format the value as seconds (`2 hr` should not become `7,200s`), and `unit: 'minute'` should not become seconds (`2 min` should not become `120s`).

kassdinzheng123:
Fixes #2011

## Summary

ECMA-402 formats the number as-is in the requested unit, but the Apple implementation left `NSMeasurementFormatter`'s `unitOptions` at the default (`Automatic`), which converts the measurement to the locale's preferred unit of the dimension. In `en-US` this renders:

| expression | before | after | V8 / JSC |
|---|---|---|---|
| `unit: 'hour'` | `7,200s` | `2 hr` | `2 hr` |
| `unit: 'minute'` | `2 min`→`120s` | `2 min` | `2 min` |
| `unit: 'kilometer'` | `1.243mi` | `2 km` | `2 km` |
| `unit: 'liter'` | `0.528gal` | `2 L` | `2 L` |
| `unit: 'kilometer-per-hour'` | `1.243mph` | `2 km/h` | `2 km/h` |

Two changes:

1. Pass `NSMeasurementFormatterUnitOptionsProvidedUnit` so the requested unit is kept instead of being converted to the locale's preferred unit of the dimension (this is what turned `2 hr` into `7,200s`, `2 km` into `1.243mi`, etc.).
2. Swap the `short`/`narrow` → `NSFormattingUnitStyle` mapping, which was inverted: `NSFormattingUnitStyleShort` is the symbol-only form CLDR calls *narrow* (`2h`) and `NSFormattingUnitStyleMedium` is CLDR *short* (`2 hr`).

## Known limitation (left as-is)

Units with no `NSUnit` counterpart (`day`, `month`, `week`, `year`, `percent`) still format through a custom `NSUnit`, and `NSMeasurementFormatter` only pluralizes built-in units, so they get no localized plural forms (`2 day` instead of `2 days`). `NSDateComponentsFormatter` has the CLDR data, but it cannot honor NumberFormat's digit options, so it cannot replace the current path; fixing this properly needs a real CLDR unit data source on Apple platforms.

## Test plan

Rebased onto `static_h` (`494787151`); the Apple `NumberFormat` unit path is identical on both branches, so the patch applies unchanged.

New `test/hermes/intl/number-format-unit.js` (`REQUIRES: intl && apple`, run through both `%hermes` and `%shermes`) covering hour/minute/second/kilometer/liter/kilometer-per-hour across short/narrow/long plus `maximumFractionDigits`, verified against the V8/JSC outputs quoted in the issue.

On a local macOS Debug build with `HERMES_ENABLE_INTL=ON`:

- reverting only the `PlatformIntlApple.mm` change fails the new test at its first check (`7,200s` instead of `2 hr`);
- with the change, `test/hermes/intl` is 14 expected passes and 1 unsupported;
- `test/hermes` is 612 expected passes, 3 expected failures and 37 unsupported, with no unexpected failures.


lavenzg:
Please make the PR against `static_h` branch instead of `main`.

kassdinzheng123:
Rebased onto `static_h` and retargeted the PR there (head is now 942e91fad). The Apple `NumberFormat` unit path is byte-identical on `main` and `static_h`, so the patch applied unchanged; the diff is still the same two files.

Re-verified on `static_h`: reverting only the `PlatformIntlApple.mm` hunk makes the new test fail at its first check (`7,200s` instead of `2 hr`), and with the fix `test/hermes/intl` is 14 expected passes plus 1 unsupported, `test/hermes` is 612 expected passes, 3 expected failures and 37 unsupported, with no unexpected failures. Both the `%hermes` and `%shermes` RUN lines are covered.

PR #2188, opened 2026-09-17
TS2Flow: convert function expression return types

thejustinwalsh:
## Summary

Convert TypeScript return annotations on `FunctionExpression` and
`ArrowFunctionExpression` nodes during TS-to-Flow AST conversion.

`FunctionDeclaration` already performed this conversion. The other two
function node kinds did not, so the post-conversion validator rejected their
remaining `TSTypeAnnotation` nodes. TypeScript class methods are represented by
`FunctionExpression`, so they were affected as well.

## Reproduction

Given:

```ts
const expression = function (value: number): string {
  return String(value);
};

const arrow = (value: number): string => String(value);

class Example {
  method(value: number): string {
    return String(value);
  }
}
```

Run:

```sh
shermes -fno-std-globals -typed -parse-ts \
  -dump-transpiled-ast -pretty -script repro.ts
```

Before this change, TS2Flow reports three errors:

```text
ts2flow: remaining TS node TSTypeAnnotation in FunctionExpression
ts2flow: remaining TS node TSTypeAnnotation in ArrowFunctionExpression
ts2flow: remaining TS node TSTypeAnnotation in FunctionExpression
```

## Root cause

`TS2FlowConverter` had a specialized visitor for `FunctionDeclaration` that
converted `_returnType`, but no equivalent visitors for `FunctionExpression`
or `ArrowFunctionExpression`. Generic child traversal cannot replace a node
stored in a scalar field, so those annotations reached `TS2FlowValidator`
unchanged.

This change adds the two missing visitors and follows the existing
`FunctionDeclaration` conversion pattern.

## Tests

Added an AST regression covering:

- a function expression;
- an arrow function; and
- a class method, whose value is a function expression.

Verified with the repository's lit runner:

```sh
cmake --build build --target shermes FileCheck
python3 build/bin/hermes-lit -sv \
  --param shermes=build/bin/shermes \
  --param FileCheck=build/bin/FileCheck \
  --param FileCheckOrRegen=build/bin/FileCheck \
  test/AST/ts2flow
```

Result: all 5 TS2Flow AST tests pass.

## Scope

This only converts return annotations for the two missing runtime function
node kinds. It does not add support for TypeScript generic type parameters or
otherwise broaden the set of TS type nodes accepted by TS2Flow.


tmikov:
lgtm

